PR #4, opened 2018-12-20
Add contentful

lbiedinger:
Using contentful and adding a .env file using dotenv.

For development purposes the space and token can be found in contentful.

In contentful a page with sections which has cards with images should be present. Once this is the case the page can be viewed at localhost:3000/PAGE_ID


nuxt.config.js
```diff
@@ -57,6 +58,7 @@ module.exports = {
     ** You can extend webpack config here
     */
     extend(config, ctx) {
+      config.node = { fs: 'empty' };
```
lbiedinger: this is needed to get dotenv working

.env.example
```diff
@@ -0,0 +1,2 @@
+CTF_SPACE_ID= "ABC123",
```
rwd: Remove the space and comma

.env.example
```diff
@@ -0,0 +1,2 @@
...
+CTF_CDA_ACCESS_TOKEN= "ABCDEF123456"
```
rwd: Remove the space

package.json
```diff
@@ -27,10 +27,14 @@
...
   "dependencies": {
+    "@nuxtjs/axios": "^5.0.0",
+    "@nuxtjs/dotenv": "^1.3.0",
+    "contentful": "^7.0.5",
     "cross-env": "^5.2.0",
-    "nuxt": "^2.0.0",
+    "dotenv": "^6.2.0",
     "express": "^4.16.3",
-    "@nuxtjs/axios": "^5.0.0"
+    "fs": "0.0.1-security",
```
rwd: This can be omitted. `fs` is part of node core.

components/contentCard.vue
```diff
@@ -0,0 +1,51 @@
+<template>
```
rwd: The file name should be ContentCard.vue. See https://vuejs.org/v2/style-guide/#Single-file-component-filename-casing-strongly-recommended

components/contentCard.vue
```diff
@@ -0,0 +1,51 @@
...
+        <img :src="imageUrl">
+      </div>
+    </a>
+  </li>
+</template>
+
+
+<script>
+export default {
+  props: {
```
rwd: With blank defaults throughout, could this be made more concise with:
```js
props: {
  cardTitle: String,
  contentSource: String,
  // ...
}
```

package.json
```diff
@@ -27,10 +27,14 @@
     "precommit": "npm run lint"
   },
   "dependencies": {
+    "@nuxtjs/axios": "^5.0.0",
```
rwd: @lbiedinger please add yourself to `contributors` in this file. :)

pages/_slug/index.vue
```diff
@@ -0,0 +1,95 @@
...
+        'sys.id': params.slug,
+        'include': 2 // this retrieves nested/linked resources like cards and their images
+      })
+    ]).then(([pages]) => {
+      // return data that should be available
+      // in the template
+      return {
+        page: pages.items[0]
+      };
+    }).catch(console.error);
```
rwd: We don't want to use console, and I don't think this use of it would do anything anyway.

plugins/contentful.js
```diff
@@ -0,0 +1,19 @@
+// ./plugins/contentful.js
+
+require('dotenv').config();
```
rwd: Let's stick this in server/index.js

pages/_slug/index.vue
```diff
@@ -0,0 +1,95 @@
...
+import ContentCard from '~/components/contentCard.vue';
+import {createClient} from '~/plugins/contentful.js';
+
+const contentfulClient = createClient();
+
+export default {
+  asyncData ({params}) {
+    return Promise.all([
+      // fetch the browsePage data
+      contentfulClient.getEntries({
```
rwd: Use `getEntry` instead.

components/ContentCard.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <li class="card">
+    <a :href="url">
```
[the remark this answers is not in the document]
lbiedinger: fair point, although currently none of the links are to anything else 'internal' to the js-portal. I wonder if there'll be a good way of identifying such links down the line.

plugins/contentful.js
lbiedinger: It doesn't seem to be required at all. However removing it resulted in some console errors, so I've simplified the export funcion of this file.

components/ContentCard.vue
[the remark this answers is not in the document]
mirjamverloop: There might be a way to intercept all links and identify them as external (and open them in a new tab/window) or internal (and use the router). Not sure if that is slowing things done.

PR #15, opened 2019-02-04
ava + nyc + package update

lbiedinger:


lbiedinger:
close as the relevant changes have been included in #20

PR #16, opened 2019-02-05
Retrieve and output record from Europeana API

rwd:


rwd:
Misnamed branch. Replaced by #17

PR #17, opened 2019-02-05
Retrieve and output record from Europeana API

rwd:
Records are available at e.g. `/record/123/abc`

Requires setting the environment variable `EUROPEANA_API_KEY` e.g. in `.env`


pages/record/_.vue
```diff
@@ -0,0 +1,112 @@
...
+        edmLanguage: edmLanguage,
+        edmRights: providerAggregation.edmRights
+      },
+      media: webResources
+    };
+  }
+
+  export default {
+    asyncData ({ params }) {
+      return axios.get(`https://api.europeana.eu/api/v2/record/${params.pathMatch}.json?wskey=${process.env.EUROPEANA_API_KEY}`)
```
mirjamverloop: Maybe, this can be moved to a configuration file. So there is one file to update the url in case there is a change.

pages/record/_.vue
```diff
@@ -0,0 +1,112 @@
...
+  function dataFromApiResponse(response) {
+    const edm = response.data.object;
+    const providerAggregation = edm.aggregations[0];
+    const europeanaAggregation = edm.europeanaAggregation;
+    const providerProxy = edm.proxies.find((proxy) => {
+      return proxy.europeanaProxy === false;
+    });
+    const edmLanguage = europeanaAggregation.edmLanguage['def'][0];
+
+    const webResources = providerAggregation.webResources.map(webResource =>{
```
mirjamverloop: Add a space between => and {

pages/record/_.vue
rwd: Agreed, but when a second use of it is introduced instead of now.

PR #22, opened 2019-02-07
Record page layout

mirjamverloop:


pages/record/_.vue
```diff
@@ -1,52 +1,64 @@
...
-    <h1 class="title">
-      Record
-    </h1>
-
-    <p v-if="error">
-      <strong>Error:</strong> {{ error }}
-    </p>
-    <template v-else>
-      <div>
+  <p v-if="error">
```
rwd: The error message is now outside of the container, all the way over to the left. Is that intentional?

mirjamverloop: No, so I made some changes.

PR #31, opened 2019-02-13
Search result metadata display logic

rwd:
* Display eng, en or def values if present; else display raw value
* Remove duplicates
* Remove URIs if other values are present
* Output multiple values as list


pages/search/index.vue
```diff
@@ -103,11 +114,51 @@
 <script>
   import LoadingSpinner from '~/components/LoadingSpinner.vue';
   import axios from 'axios';
+  // TODO: cherry-pick Lodash methods? or load from CDN?
+  import _ from 'lodash';
```
mirjamverloop: Maybe it's worth to investigate if we need lodash at all. [This](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore) is a handy overview.

rwd:
@lbiedinger re:
> Down the line something that's quite apparent is missing labels.

Do you mean field labels like "creator"? If so, we may very well not have them down the line on a search results page.

pages/search/index.vue
rwd: Where native equivalents exist, they have been switched to. Where not, Lodash's have been cherry-picked.

lbiedinger:
> Do you mean field labels like "creator"? If so, we may very well not have them down the line on a search results page.

Yes, creator & providing institution, or perhaps "by" and "from". But yes, until it's clear if and how these should be displayed let's not add anything.

PR #34, opened 2019-02-15
EC-3333 test quality

lbiedinger:
PR to see tavis results and CodeclimateCoverage report.


lbiedinger:
Won't use this as we've restructured tests anyway.

PR #39, opened 2019-02-18
Component unit tests

lbiedinger:
EC-3333

### DIFF
--- a/tests/unit/PageHeader.spec.js
+++ b/tests/unit/PageHeader.spec.js
@@ -11,9 +11,27 @@ const factory = () => shallowMount(PageHeader, {
 });
 
 describe('PageHeader', () => {
-  it('contains a search form', () => {
+  it('contains a search form with a query field', () => {
     const wrapper = factory();
 
-    expect(wrapper.html()).to.contain('form');
+    const queryField =  wrapper.find('[data-qa="search query input"]');
+
+    expect(queryField.attributes().name).to.equal('query');
+  });
+
+  it('contains a search form submit button', () => {
+    const wrapper = factory();
+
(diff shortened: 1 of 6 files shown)


tests/unit/PageHeader.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    const submitButton =  wrapper.find('[data-qa="search submit button"]');
+
+    expect(submitButton.attributes().type).to.equal('submit');
+  });
+
+  it('contains the logo', () => {
+    const wrapper = factory();
+
+    const logo = wrapper.find('img');
```
rwd: Add a data-qa attribute to the logo element?

tests/unit/PageHeader.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    expect(logo.attributes().src).to.match(/.?\.svg/);
```
rwd: A better RegExp may be `/\.svg$/`

tests/unit/PageHeader.spec.js
```diff
@@ -0,0 +1,37 @@
...
+describe('PageHeader', () => {
+  it('contains a search form with a query field', () => {
+    const wrapper = factory();
+
+    const queryField =  wrapper.find('[data-qa="search query input"]');
+
+    expect(queryField.attributes().name).to.equal('query');
+  });
+
+  it('contains a search form submit button', () => {
```
rwd: Let's also test that clicking the button results in the function being called.

https://vue-test-utils.vuejs.org/guides/dom-events.html

PR #40, opened 2019-02-21
Split pages into components

mirjamverloop:


components/browse/index.js
```diff
@@ -0,0 +1,5 @@
+import ContentCardSection from './ContentCardSection';
```
rwd: This file seems redundant when it only exports one .vue component.

pages/_.vue
```diff
@@ -4,35 +4,16 @@
...
+      <ContentCardSection
+        :page="page"
+      />
     </div>
   </section>
 </template>
 
 <script>
-  import ContentCard from '~/components/ContentCard.vue';
+  import { ContentCardSection } from '~/components/browse';
```
rwd: Change this to reference the Vue component directly: `import ContentCardSection from '../../components/ContentCardSection';`

(Also using `../../` instead of `~` as we have had issues writing unit tests with the latter.)

pages/search/index.vue
```diff
@@ -252,11 +190,9 @@
       });
     },
     methods: {
-      submitSearchForm () {
-        if (this.$route.query.query !== this.query) {
-          this.isLoading = true;
-        }
-        this.$router.push({ name: 'search', query: { query: this.query ? this.query : '' } });
+      updateIsLoading (status) {
+        console.log('updateIsLoading', status);
```
lbiedinger: can be removed

components/browse/index.js
mirjamverloop: I have removed all index.js and updated the relevant import statements in components and pages.

pages/_.vue
mirjamverloop: Updated this in all relevant files.

components/browse/ContentCardSection.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <div>
+    <div
+      v-for="section in page.hasPart"
```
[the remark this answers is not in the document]
mirjamverloop: Yes, that makes sense. I have updated pages/_.vue and the component itself.

components/browse/ContentCardSection.vue
```diff
@@ -0,0 +1,40 @@
...
+
+<script>
+  import { ContentCard } from '~/components/generic';
+
+  export default {
+    components: {
+      ContentCard
+    },
+    props: {
+      page: {
```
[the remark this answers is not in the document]
mirjamverloop: You are right, I have updated the prop.

pages/search/index.vue
mirjamverloop: Removed it.

components/record/MediaImage.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <a
+    v-if="image"
```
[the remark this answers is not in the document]
mirjamverloop: I have changed the conditionals in this component.

rwd:
@mirjamverloop on this branch, after searching, the query term is no longer displayed in the search input field on the results page.

components/search/SearchForm.vue
```diff
@@ -0,0 +1,57 @@
...
+        default: () => {}
+      },
+      isLoading: {
+        type: Boolean,
+        default: false
+      }
+    },
+    data () {
+      return {
+        query: this.$route.query.query ? this.$route.query.query : null
```
rwd: Let's just move this into `props` instead:
```
$ git diff
diff --git a/components/search/SearchForm.vue b/components/search/SearchForm.vue
index c448c66..4de28da 100644
--- a/components/search/SearchForm.vue
+++ b/components/search/SearchForm.vue
@@ -31,6 +31,10 @@
       LoadingSpinner
     },
     props: {
+      query: {
+        type: String,
+        default: ''
+      },
       fields: {
         type: Object,
         default: () => {}
@@ -40,11 +44,6 @@
         default: false
       }
     },
-    data () {
-      return {
-        query: this.$route.query.query ? this.$route.query.query : null
-      };
-    },
     methods: {
       submitSearchForm () {
         if (this.$route.query.query !== this.query) {
diff --git a/pages/search/index.vue b/pages/search/index.vue
index 116d416..b0b9341 100644
--- a/pages/search/index.vue
+++ b/pages/search/index.vue
@@ -9,6 +9,7 @@
       <b-col>
         <SearchForm
           :is-loading="isLoading"
+          :query="query"
           @update="updateIsLoading"
         />
       </b-col>
```

PR #42, opened 2019-02-27
Page components (with v-model and watch)

rwd:


components/search/SearchForm.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <b-form
+    inline
+    @submit.prevent="$emit('submit:searchForm')"
+  >
+    <b-form-input
+      ref="query"
```
mirjamverloop: This ref can be deleted.

rwd: Indeed, and it has been in the following commit.

PR #48, opened 2019-03-04
Record page plugin and test

mirjamverloop:


pages/record/_.vue
```diff
@@ -37,78 +37,38 @@
...
+      return {
+        error: null,
+        record: null
+      };
+    },
+    asyncData ({ env, params }) {
+      return getRecord({
+        path: params.pathMatch,
+        key: env.EUROPEANA_API_KEY
+      }).then((result) => {
```
rwd: We need to `catch` any rejected promises here too.

pages/record/_.vue
```diff
@@ -37,78 +37,38 @@
...
     },
+    data () {
+      return {
+        error: null,
+        record: null
+      };
+    },
+    asyncData ({ env, params }) {
+      return getRecord({
+        path: params.pathMatch,
```
rwd: I think it's clearer if `getRecord` receives a parameter named `europeanaId` which is already prefixed with a `/` character.

plugins/europeana/record.js
```diff
@@ -0,0 +1,76 @@
+import axios from 'axios';
+import omitBy from 'lodash/omitBy';
+
+/**
+ * Parse the record data based on the data from the API response
+ * @param {Object} response data from API response
+ * @return {Object} v parsed data
```
rwd: Remove "v "

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+describe('plugins/europeana/record', () => {
+  describe('getRecord()', () => {
+
+    describe('when trying to request an unknown record', () => {
+      const errorMessage = 'Invalid record identifier';
+      beforeEach('stub API response', () => {
+        nock(apiUrl)
+          .get(apiEndpoint)
+          .query(true)
+          .reply(400, {
```
rwd: Let's make this `404` which is what the API returns in this scenario.

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+            record: null,
```
rwd: Omit `record` as the API does not return it, but add `"success": false` which it does return.

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+        return response.should.eventually.have.property('error', errorMessage);
+      });
+    });
+
+    describe('when trying to request a record', () => {
+      const apiResponse = {
+        success: true,
+        object: {
+          aggregations: [{
+            edmIsShownAt: 'https://thisisalink.eu',
```
rwd: Let's use example.org FQDNs instead of these .eu ones.

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+            webResources: [{
+              about: 'https://thisistheabout.eu',
+              dcDescription: 'This is a description',
```
rwd: A web resource's `dcDescription` is always a language map. Let's make it so here too.

e.g.
```json
"dcDescription": {
  "de": [
    "Umfang: F1 - f3",
    "Latcham, Michael: The pianos of Johann Andreas Stein. In: Zur Geschichte des Hammerklaviers. 14. Musikinstrumentenbau-Symposium in Michaelstein am 12. und 13. November 1993. Michaelstein 1996, S. 15-49."
  ]
}
```

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+              webResourceEdmRights: 'These are the rights'
```
rwd: Bizarrely, this is also a language map.

e.g.
```json
"webResourceEdmRights": {
  "def": [
    "http://creativecommons.org/licenses/by-nc-sa/3.0/"
  ]
}
```

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -0,0 +1,75 @@
...
+            }]
+          }],
+          europeanaAggregation: {
+            edmLanguage: { def: [ 'en' ] },
+            edmRights: { def: [ 'http://rightsstatements.org/vocab/InC/1.0/' ] },
+            edmPreview: 'https://thisisapreview.eu'
+          },
+          proxies: [{
```
rwd: As the provider proxy is where we take much of the displayed record data from, let's add one with a few fields.

PR #49, opened 2019-03-04
Search component tests

rwd:


components/search/SearchResult.vue
```diff
@@ -1,18 +1,22 @@
...
+    v-if="result"
+    no-body
+  >
     <b-media-aside class="w-25 mr-3">
       <b-img
         v-if="result.edmPreview"
         slot="aside"
         :src="result.edmPreview"
         class="mw-100"
+        data-qa="edmPreview"
```
lbiedinger: let's make the qa attribute a little more human readable.
 'result image', 'result thumbnail', or even 'edm preview'
For cucumber we have  nested selectors so even just 'thubmnail' would be sufficient, since we can then access it via for example ``` `search result` `thumbnail` ``` as long as the result has the `search result` attribute.

rwd: No, let's keep it as `edmPreview` as that is consistent with the other EDM fields' `data-qa` attribute.

rwd: We could prefix them all with something, but I want to keep the RDF predicate in there.

lbiedinger: sure let's use both then so it'll be `data-qa="thumbnail edmPreview"`

You'll need to update the SearchResult spec to look for: `const image =  wrapper.find('img[data-qa~="edmPreview"]');` with `~=`.

And in order to get cucumber to match on these elements then we update the nested selector logic with:
```
  ...
  return prefixSelectors
    .concat(selectors)
    .map(x => x.split(' ').map(y => `[data-qa~="${y.replace(/`/g, '')}"]`).join(''))
    .join(' ');
```

taken from https://css-tricks.com/multiple-attribute-values/ and tried out on my machine, seems to work.

PR #51, opened 2019-03-05
Cucumber tests for search interactions

lbiedinger:


tests/features/search.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Search
+
+  Scenario: Search existing Europeana content
+
+    Given I visit a `search page`
```
rwd: These scenarios are missing a `When` step.

Perhaps the `Given` steps should be `When` steps? Or the `Given` reworded as ``Given I am on a `search page` `` and the first following step made a `When`.

tests/features/support/steps.js
```diff
@@ -20,9 +20,9 @@ defineStep(/^I (?:find|identify|see|spot).*? (`.*`).*?$/, selectorChain =>
   client.expect.element(nestedSelector(selectorChain)).to.be.visible);
 
 defineStep(/^I (?:can|don)'t (?:find|identify|see|spot).*? (`.*`).*?$/, selectorChain =>
-  client.expect.element(nestedSelector(selectorChain)).to.not.be.visible);
+  client.expect.element(nestedSelector(selectorChain)).to.not.be.present);
```
rwd: Are we sure about this change? The step text is about finding/seeing/etc rather than presence/absence in the HTML.

lbiedinger: I think we may want both present and visible checks.

However for this situation '.not.be.visible' causes an error because it can't find the element to check visibility for.

PR #56, opened 2019-03-05
Record component unit tests

rwd:


tests/unit/components/record/MediaImage.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      image.attributes().src.should.eq(props.src);
+    });
+  });
+
+  describe('when neither src nor link are present', () => {
+    it('does not show an image', () => {
+      const wrapper = factory();
+      const props = { };
+
+      wrapper.setProps(props);
```
mirjamverloop: This seems a bit redundant, it might be consistent, but is it necessary to set empty props?

PR #58, opened 2019-03-05
Updated and new unit tests for generic components

mirjamverloop:


tests/unit/components/generic/ContentCard.spec.js
```diff
@@ -9,11 +9,36 @@ const factory = () => shallowMount(ContentCard, {
...
+    const link =  wrapper.find('[data-qa="content card"] .card-link');
+    link.attributes().href.should.eq('https://example.org');
+  });
+
+  it('has an image', () => {
+    const wrapper = factory();
+    wrapper.setProps({ imageUrl: 'https://example.org' });
+
+    const card =  wrapper.find('[data-qa="content card"]');
+    card.attributes().imgsrc.should.eq('https://example.org');
```
rwd: I would be tempted to do a full mount, not a shallow mount, and test that there is an actual `img` displayed.

PR #59, opened 2019-03-06
Content card section tests

lbiedinger:
EC-3377

### DIFF
--- a/components/browse/ContentCardSection.vue
+++ b/components/browse/ContentCardSection.vue
@@ -1,23 +1,17 @@
 <template>
   <div
-    v-if="section"
+    v-if="section && section.fields"
     class="banner"
     data-qa="browse section"
   >
-    <h2
-      v-if="section.fields"
-      data-qa="section headline"
-    >
+    <h2 data-qa="section headline">
       {{ section.fields.headline }}
     </h2>
-    <p
-      v-if="section.fields"
-    >
+    <p>
       {{ section.fields.text }}
     </p>
 
     <b-card-group
-      v-if="section.fields"
       deck
       data-qa="section group"
     >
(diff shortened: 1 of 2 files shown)


components/browse/ContentCardSection.vue
```diff
@@ -1,12 +1,26 @@
 <template>
   <div
+    v-if="section"
     class="banner"
     data-qa="browse section"
   >
-    <h2>{{ section.fields.headline }}</h2>
-    <p>{{ section.fields.text }}</p>
+    <h2
+      v-if="section.fields"
```
mirjamverloop: There is an if for every element right now, that looks as an unnecessary repetition. I think that you don't have to check if the section exists on line 3, as that happened already in the page itself. Check on line 3 if the section.fields exists.

tests/unit/components/browse/ContentCardSection.spec.js
```diff
@@ -0,0 +1,64 @@
...
+            },
+            {
+              sys: { id: '456' },
+              fields: { name: 'Card two title', description: 'the second card', url: 'http://europeana.eu', image: {
+                fields: { file: { url: 'img/portrait.jpg' } } }
+              }
+            }
+          ]
+        }
+      };
```
mirjamverloop: I think you can create a more global variable "section" outside of the individual tests and us them in both test. As you can test if there is a headline using this more detailed "section" as well.

tests/unit/components/browse/ContentCardSection.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+      wrapper.setProps({ section: section });
+
+      const cardGroup = wrapper.find('[data-qa="section group"]');
+
+      console.log(cardGroup.html());
+      console.log(cardGroup.findAll('[data-qa="content card"]'));
```
mirjamverloop: You can remove these two console.logs.

PR #60, opened 2019-03-06
Record page cukes

lbiedinger:
EC-3344


components/record/WebResources.vue
```diff
@@ -3,7 +3,7 @@
     <b-list-group-item
       v-for="webResource in media"
       :key="webResource.rdfAbout"
-      :data-qa="'webResource ' + webResource.rdfAbout"
+      :data-qa="'web resource webResource ' + webResource.rdfAbout"
```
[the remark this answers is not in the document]
lbiedinger: A readme certainly makes sense. It wouldn't be component specific though, we can have QA elements at page level. Perhaps something more generic in our Wiki as opposed to in the repository?

I wouldn't be too concerned about the nestedSelector logic since I don't think we should specify anything as being "non xyz", as that would lead to describing everything as not anything but itself. As long as we only use "affirmatively descriptive" QA attributes we can use negative steps   like the ones with "I (can't|don't|shouldn't|won't) see..." to check for things not being/containing certain content.

tests/features/record-page.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Record page
+
+  Scenario: View any existing record page
+
+    When I open a `record page`
+    Then I see the `record page`
+    And I see a `record page` `field`
+    And I see a `record page` `web resource`
+    And I see a `record page` `web resource` `field`
```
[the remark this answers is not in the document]
lbiedinger: I've tweaked them a bit and added more 'flavour' text to make them more understandable.

components/record/WebResources.vue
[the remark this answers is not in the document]
mirjamverloop: I agree with documenting the specifications for the data attribute. The wiki makes sense, the readme on the other hand is a bit closer to the actual code.

tests/features/header.feature
```diff
@@ -11,3 +11,8 @@ Feature: Page Header
     Then I see the `header`
     And I see the `header` `logo`
     And I see the `header` `search box`
+
+    When I open a `record page`
+    Then I see the `header`
+    And I see the `header` `logo`
+    And I see the `header` `search box`
```
rwd: Let's rewrite these to use the "`x` in the `y`" syntax used in `record-page.feature`.

tests/README.md
```diff
@@ -0,0 +1,43 @@
+## Tests
+
+The test directory contains unit and e2e tests.
+
+
+## Guidlines on testing end to end(e2e)
+
+End to end tests are written as Gherkin scripts for cucumber.js inside the tests/features directory.
+End to end scripts should be written in as human readable way as possible.
```
rwd: Add "a" before "way".

tests/README.md
```diff
@@ -0,0 +1,43 @@
...
+The step definitions which are available are based of [this](https://markus.oberlehner.net/blog/acceptance-testing-with-nightwatch-and-cucumber-setup/) and [this guide](https://markus.oberlehner.net/blog/acceptance-testing-with-nightwatch-and-cucumber-smart-step-definitions/).
```
rwd: "based on"

tests/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Usage of _data-qa_
+
+In order to isolate testing from the implementation we make use of data-qa attributes.
+The data-qa attributes should be used primarily in e2e tests and can be added on any element.
+Matching of data-qa elements will look for an exact match when using the standard deffinitions.
```
rwd: "definitions"

tests/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+When adding or refactoring data-qa attributies, keep in mind that they:
```
rwd: "attributes"

tests/features/support/steps.js
```diff
@@ -27,3 +32,13 @@ defineStep(/^I (?:enter|fill|input|supply|type).*? "(.*?)" in.*? (`.*`)$/, (valu
...
   client.click(nestedSelector(selectorChain)));
+
+defineStep(/^I should be on.*? `(.*)`$/, pageName => {
+  let expectedUrl;
+  if (pageName.startsWith('/')) {
+    expectedUrl = `${url}${pageName}`;
+  } else {
+    expectedUrl = pages[pageName];
+  }
+  client.assert.urlEquals(expectedUrl);
```
rwd: Use `expect` as elsewhere instead of `assert`.

tests/features/support/europeana-identifiers.js
```diff
@@ -0,0 +1,15 @@
+// Helpers for finding europeana IDs for records matching specific criteria
```
rwd: Use JSDoc `@file`

tests/features/support/steps.js
lbiedinger: I couldn't find the expect equivalent https://github.com/nightwatchjs/nightwatch/issues/861, I'm not sure expect suppports urlEquals

rwd: The last comment there shows an alternative, i.e.
```js
  client.url(currentUrl => {
    client.expect(currentUrl.value).to.eql(expectedUrl);
  });
```

PR #63, opened 2019-03-07
Updated header and footer tests

mirjamverloop:


components/PageHeader.vue
```diff
@@ -15,39 +15,29 @@
...
-            src="../assets/img/magnifier.svg"
-            alt="Search"
-          >
-        </b-button>
-      </b-nav-form>
+      <SearchForm
+        v-model="query"
+        :is-loading="isLoading"
+        :in-header="inHeader"
+        data-qa="search form header"
```
lbiedinger: 'header' is redundant, we can infer this in both unit and e2e tests by selecting the b-navbar first element which has a 'data-qa="header"' already.

lbiedinger: in fact 'data-qa="search form"' is applied to the form itself so we don't need any qa attribute here at all.

components/PageHeader.vue
```diff
@@ -15,39 +15,29 @@
...
-          <img
-            src="../assets/img/magnifier.svg"
-            alt="Search"
-          >
-        </b-button>
-      </b-nav-form>
+      <SearchForm
+        v-model="query"
+        :is-loading="isLoading"
+        :in-header="inHeader"
```
lbiedinger: the in header 'argument' is a bit too specific perhaps. Since when it gets detected all it does is set the input size to small,  let's call the SearchForm component with that information here already.

tests/unit/components/PageFooter.spec.js
```diff
@@ -3,12 +3,11 @@ import PageFooter from '../../../components/PageFooter.vue';
...
 
-describe('PageFooter', () => {
+describe('components/search/PageFooter', () => {
   it('it exists', () => {
     const wrapper = factory();
 
     const footer = wrapper.find('footer');
-
-    footer.should.exist;
+    footer.exists().should.eq(true);
```
lbiedinger: if it's equivalent let's stick with `footer.should.exist;`

tests/unit/components/PageHeader.spec.js
```diff
@@ -9,28 +9,18 @@ const factory = () => shallowMount(PageHeader, {
...
-    queryField.attributes().name.should.equal('query');
-  });
-
-  it('contains a search form submit button', () => {
-    const wrapper = factory();
-
-    const submitButton =  wrapper.find('[data-qa="search button"]');
-
-    submitButton.attributes().type.should.equal('submit');
+    form.exists().should.eq(true);
```
lbiedinger: `form.should.exist;`

components/PageHeader.vue
mirjamverloop: I have mounted the component, so I am able to check if the form exists. I left the data-qa="search form" on the form in the actual component.

components/PageHeader.vue
mirjamverloop: I have removed the inHeader completely, so there is no longer a prop for styling purposes. This way it is simple and functional for now and we can make design modifications later.

PR #65, opened 2019-03-08
Labels for record page

lbiedinger:


components/record/Metadata.vue
```diff
@@ -6,19 +6,24 @@
       :data-qa="'field ' + key"
       class="border-bottom mb-3"
     >
-      <div><strong>{{ key }}</strong></div>
+      <div><strong>{{ fieldLabel(key) }}</strong></div>
```
rwd: Use a [Vue computed property](https://vuejs.org/v2/guide/computed.html) instead of calling the function here.

plugins/europeana/record.js
```diff
@@ -1,6 +1,27 @@
...
+  dcType: 'Type of object',
+  dctermsCreated: 'Creation date',
+  edmCountry: 'Providing country',
+  edmDataProvider: 'Providing institution',
+  edmRights: 'License of the media in this record (unless otherwise specified)'
+};
+
+/**
+ * Lookup the label for the provided metadata field key
+ * @key {String} key for metadata field
```
rwd: `@param` not `@key`.

Strings are documented lowercase, i.e. `{string}` in JSDoc.

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -86,4 +86,65 @@ describe('plugins/europeana/record', () => {
...
+          it('looks up the label from the fieldLabels object', () => {
+            return fieldLabel(availableFields[key]).should.not.eq(availableFields[key]);
+          });
+        });
+      }
+    });
+
+    describe('looking up a non available key', () => {
+      describe('for rdfAbout', () => {
+        it('retruns the key as is', () => {
```
rwd: "returns"

tests/unit/plugins/europeana/record.spec.js
```diff
@@ -86,4 +86,65 @@ describe('plugins/europeana/record', () => {
...
+      'edmDataProvider',
+      'edmRights',
+      'dcContributor',
+      'dcCreator',
+      'dcDescription'
+    ];
+
+    describe('looking up an available key', () => {
+      let key;
+      for (key in availableFields) {
```
rwd: Use `for...of` here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Iterating_over_an_Array

components/record/Metadata.vue
rwd: That likely however implies moving the template for one metadata field into another component with a computed property for its key.

rwd: Or perhaps more simply just moving the `v-for` up to the parent component and making this component responsible for rendering a single metadata field, renaming it to `MetadataField.vue`.

mirjamverloop: You can create a custom (global) filter as well for this renaming. 

https://vuejs.org/v2/guide/filters.html
https://scotch.io/tutorials/how-to-create-filters-in-vuejs-with-examples

components/record/MetadataField.vue
```diff
@@ -3,16 +3,18 @@
...
     </div>
-    <pre data-qa="metadata field value">
+    <pre data-qa="value">
       <code>{{ value }}</code>
     </pre>
   </div>
 </template>
 
 <script>
+  import fieldLabel from '../../plugins/europeana/labeling';
```
rwd: "labelling" not "labeling"

components/record/MetadataField.vue
```diff
@@ -22,6 +24,15 @@
       value: {
         type: [String, Object],
         default: ''
+      },
+      context: {
+        type: String,
+        default: null
```
rwd: Make this 'default'.

components/record/MetadataField.vue
```diff
@@ -22,6 +24,15 @@
       value: {
         type: [String, Object],
         default: ''
+      },
+      context: {
```
rwd: Should we give this a [validator](https://vuejs.org/v2/guide/components-props.html#Prop-Validation) to ensure it's a known context?

plugins/europeana/labeling.js
```diff
@@ -0,0 +1,40 @@
...
+    edmRights: 'License of the media in this record (unless otherwise specified)'
+  },
+  webResource: {
+    edmRights: 'License for this media resource'
+  }
+};
+
+function lookupFromContext(key, context = 'default') {
+  context = fieldLabels.hasOwnProperty(context) ? fieldLabels[context] : fieldLabels['default'];
+  return context.hasOwnProperty(key) ? context[key] : null;
```
rwd: Shorten to:
```js 
return context[key] || null;
```

plugins/europeana/labeling.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+function fieldLabel(key, options = {}) {
+  let label;
+  if (options.hasOwnProperty('context')) {
+    label = lookupFromContext(key, options['context']);
+  }
+  if (!label) {
+    label = lookupFromContext(key);
+  }
+  return label ? label : key;
```
rwd: Shorten to:
```js 
return label || key;
```

plugins/europeana/labeling.js
```diff
@@ -0,0 +1,40 @@
...
+    edmDataProvider: 'Providing institution',
+    edmRights: 'License of the media in this record (unless otherwise specified)'
+  },
+  webResource: {
+    edmRights: 'License for this media resource'
+  }
+};
+
+function lookupFromContext(key, context = 'default') {
+  context = fieldLabels.hasOwnProperty(context) ? fieldLabels[context] : fieldLabels['default'];
```
rwd: Perhaps define a new `const` here named `contextLabels` instead of overwriting the parameter.

rwd: If `context` parameter has been passed and does not exist as a property of `fieldLabels`, this method should throw an error instead of falling back to default as the caller has explicitly requested an unavailable context.

plugins/europeana/labeling.js
```diff
@@ -0,0 +1,40 @@
...
+  return context.hasOwnProperty(key) ? context[key] : null;
+}
+
+/**
+ * Lookup the label for the provided metadata field key
+ * @param {string} key for metadata field
+ * @param {Object} options
```
rwd: Document the available options like:
```
* @param {string} options.context context in which label is used
```

components/record/MetadataField.vue
lbiedinger: Could do, but then I'd be tempted to get the available contexts from the labelling plugin via say `fieldLabelContexts` and while that would work, I think it unnecessarily ties the context to be specific to labelling. Perhaps we want contexts for the metadata fields that apply [case filters](https://vuejs.org/v2/guide/filters.html) in future.
The labelling will fall back to the default context as is, so if an unknown labeling context would be applied now I don't see it as worth raising console errors, just let it show the default labels from the record or fall back to the key.

plugins/europeana/labeling.js
lbiedinger: see my comment above about not explicitly wanting to tie contexts to labelling.

components/record/MetadataField.vue
```diff
@@ -22,6 +24,18 @@
...
+        type: String,
+        default: 'default',
+        validator: (value) => {
+          return fieldLabellingContexts().indexOf(value) !== -1;
```
rwd:
```js
return fieldLabellingContexts().includes(value);
```

components/record/MetadataField.vue
```diff
@@ -22,6 +24,18 @@
...
+        }
+      }
+    },
+    computed: {
+      fieldNameLabel: function() {
+        return fieldLabel(this.name, { context: this.context });
```
mirjamverloop: It is not a reserved word, but you'd better avoid "name".
See also: https://www.w3schools.com/js/js_reserved.asp

lbiedinger: As discussed let's leave it for now, if it's causing us issues down the line we may refacotr with something like 'fieldName'.

PR #66, opened 2019-03-08
Request `type` facet from API in searches

rwd:


rwd:
@lbiedinger this would be a good one to squash and merge.

PR #69, opened 2019-03-11
Facet and selected facets component

mirjamverloop:
+ Storybook
+ Unit tests


pages/search/index.vue
```diff
@@ -102,6 +121,9 @@
           this.isLoading = true;
           this.$router.push({ name: 'search', query: { query: this.query || '' } });
         }
+      },
+      selectTypeFacet (selected) {
```
lbiedinger: Let's make this generic already. `selectFacet(facet, value)`

components/search/SearchFacets.stories.js
```diff
@@ -0,0 +1,20 @@
+import { storiesOf } from '@storybook/vue';
+
+import SearchFacets from './SearchFacets.vue';
+
+storiesOf('Search page', module)
+  .add('Search facets', () => ({
+    components: { SearchFacets },
+    data() {
+      return { optionsType: { 'TYPE': { 'TEXT': 123456, 'VIDEO': 789, '3D': 10 } } };
```
lbiedinger: optionsType containing 'TYPE' seems redundant. If the facet keys are present anyway let's just use the same options everywhere.

components/search/SearchFacets.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <b-card
+    v-if="optionsType && optionsType['TYPE']"
```
lbiedinger: Perhaps just iterate over each of the values inside options and then have a separate component to render the facet options.

lbiedinger: Every facet _should_ be displayable as checkboxes, even if some down the line may be better displayed using another component.

pages/search/index.vue
```diff
@@ -12,6 +12,9 @@
           :is-loading="isLoading"
           @submit:searchForm="submitSearchForm"
         />
+        <SearchSelectedFacets
+          :selected="typeFacet"
```
lbiedinger: pass the whole facets in and get the kind of facet from the key, then maybe create a sub component.

PR #70, opened 2019-03-12
Multi-lingual record metadata

rwd:


components/record/LangMap.vue
```diff
@@ -0,0 +1,56 @@
...
+      mapContent: function() {
+        const langMap = this.value;
+
+        let listOfValues = [];
+        for (let key in langMap) {
+          // "und" is the ISO-639-2 code for "undetermined"
+          const contentLang = key == 'def' ? 'und' : key;
+          for (let singleValue of langMap[key]) {
+            if (this.prioritisedLanguages.includes(key)) {
+              listOfValues.unshift({ content: singleValue, lang: contentLang });
```
lbiedinger: While the ticket doesn't have it as an explicit requirement, I thought we weren't going to show the language on the 'default'/'english' values.

In which case this could just not set 'lang' and check if it's set in the template.

components/record/MetadataField.vue
```diff
@@ -6,14 +6,27 @@
...
+    <LangMap
+      v-if="isLangMap"
+      :value="value"
+      data-qa="metadata field value"
+    />
+    <pre
+      v-else
+      data-qa="metadata field value"
+    >
       <code>{{ value }}</code>
```
lbiedinger: if we'll render mapped values inside `<span :lang="mapValue.lang">xyz</span>`, then let's render literal values as `<span :lang="und">xyz</span>` too.

lbiedinger: or perhaps just inside a span, language could set on the record as a whole.

components/record/LangMap.vue
```diff
@@ -0,0 +1,56 @@
+/*
+ * @file Presentation of "LangMap" data type in metadata
+ * @see {@link https://pro.europeana.eu/page/europeana-rest-api#getting-started|Datatypes}
+ */
+
+<template>
+  <ul data-qa="LangMap">
```
lbiedinger: LangMap isn't a great data-qa attribute if we're going to be using these for feature tests as outlined in the readme on the record e2e tests. Perhaps 'translated values'?

components/record/MetadataField.vue
rwd: Switched to a `div` without `lang` attribute.

components/record/LangMap.vue
rwd: Went with "multi-lingual metadata".

components/record/LangMap.vue
```diff
@@ -0,0 +1,56 @@
...
+      },
+      prioritisedLanguages: {
+        type: Array,
+        default: function() {
+          return ['eng', 'en'];
+        }
+      }
+    },
+    computed: {
+      // TODO: move to a plugin? or a filter?
```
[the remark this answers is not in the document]
rwd: Not sure if a filter is best for this given that it returns an Array. In any case, we will refactor to a plugin at a later iteration when the need arises.

PR #71, opened 2019-03-12
Pagination component

mirjamverloop:
+ Storybook update
+ Unit tests

### DIFF
--- a/components/generic/PaginationNav.vue
+++ b/components/generic/PaginationNav.vue
@@ -39,11 +39,7 @@
     },
     computed: {
       totalPages: function () {
-        if (this.totalResults > maxResults) {
-          return Math.ceil(maxResults / this.perPage);
-        } else {
-          return Math.ceil(this.totalResults / this.perPage);
-        }
+        return Math.ceil(Math.min(this.totalResults, maxResults) / this.perPage);
       }
     }
   };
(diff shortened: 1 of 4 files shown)


components/generic/PaginationNav.vue
```diff
@@ -0,0 +1,50 @@
...
+      }
+    },
+    data () {
+      return {
+        currentPage: null
+      };
+    },
+    computed: {
+      totalPages: function () {
+        if (this.totalResults > maxResults) {
```
rwd: The code in this function can be shortened to:
```js
return Math.ceil(Math.min(this.totalResults, maxResults) / this.perPage);
```

components/generic/PaginationNav.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <b-pagination-nav
```
rwd: The links for next/previous page seem to change the page number in the URL incorrectly.

components/generic/PaginationNav.vue
```diff
@@ -0,0 +1,50 @@
...
+    v-model="currentPage"
+    :limit="limit"
+    :hide-ellipsis="hideEllipsis"
+    :number-of-pages="totalPages"
+    base-url="#page="
```
rwd: I think we want to be using [router links](https://bootstrap-vue.js.org/docs/components/pagination-nav/#router-links) here with parameters not after the `#`.

PR #77, opened 2019-03-14
Pagination error handling

rwd:
* Detect API pagination limit errors and rephrase without reference to cursor pagination
* Display "no more results" message if user requests page beyond number available
* Redirect to page 1 if invalid value for page requested, e.g. "0", "last"


pages/search/index.vue
```diff
@@ -136,8 +144,13 @@
         page: 1
       };
     },
-    asyncData ({ env, query, res }) {
-      const currentPage = query.page ? Number(query.page) : 1;
+    asyncData ({ env, query, res, redirect }) {
+      const currentPage = pageFromQuery(query.page);
+      if (currentPage === null) {
+        // Redirect non-positive integer values for `page` to `page=1`
+        return redirect({ name: 'search', query: { query: query.query || '', page: '1' } });
```
lbiedinger: This and the pagination links below might get messy if we continue to create them like this by plucking each value. Already thinking about adding multiple facets.
Something like:
```
        query['page'] = '1';
        return redirect({ name: 'search', query: query });
```
Will be easier to maintain for when we start adding more params.

plugins/europeana/search.js
```diff
@@ -142,6 +142,27 @@ function facetsFromApiResponse(response) {
...
+/**
+ * Page to request from API based on URL query parameter
+ * If parameter is not present, returns default of page 1.
+ * If parameter is present, and represents a positive integer, return it
+ * typecast to Number.
+ * Otherwise, parameter is invalid for page number, and return `null`.
+ * @param {string} queryPage `page` query parameter from URL
+ * @return {?number}
+ */
+export function pageFromQuery(queryPage) {
```
lbiedinger: Let's add some feature tests, something like:
```
Feature: Pagination

  Scenario: Being redirected

    When I open `/search?query=&page=-1`
    Then I should be on `/search?query=&page=1`

    When I open `/search?query=&page=0`
    Then I should be on `/search?query=&page=1`

    When I open `/search?query=&page=one`
    Then I should be on `/search?query=&page=1`

    When I open `/search?query=&page=last`
    Then I should be on `/search?query=&page=1`

    When I open `/search?query=&page=2.5`
    Then I should be on `/search?query=&page=1`
```

When I ran that locally initially it caused errors with the 'I should be on xyz' step again. Same as before the step which checks the url seems to run outside of the feature and interferes with the next step.
I was able to fix it by making that step await it's completion.
```
defineStep(/^I should be on.*? `(.*)`$/, async(pageName) => {
  let expectedUrl;
  if (pageName.startsWith('/')) {
    expectedUrl = `${url}${pageName}`;
  } else {
    expectedUrl = pages[pageName];
  }
  // TODO: update if a less verbose syntax becomes available.
  // See https://github.com/nightwatchjs/nightwatch/issues/861
  await client.url(currentUrl => {
    client.expect(currentUrl.value).to.eq(expectedUrl);
  });
});

```

PR #82, opened 2019-03-20
Windows compatibility

lbiedinger:
Windows converts all the linebreaks from 'lf' to 'crlf'.
Which causes linting errors due to this line: https://github.com/europeana/incubator-portal-vue-nuxt/blob/master/.eslintrc.js#L30 When `npm start` runs, it sets the isDev mode. When that is set to true, that initializes the eslinting here: https://github.com/europeana/incubator-portal-vue-nuxt/blob/master/nuxt.config.js#L83
The warning overlay with the linting errors on localhost:1337/ then prevents any interaction or detection of elements all the e2e tests fail.
This PR changes the test server to run in non 'isDev' mode.

The concurrently command in package.json needs to have nested strings escaped with \" instead of using single quotes.


server/index.js
```diff
@@ -30,7 +30,7 @@ app.set('port', port);
 
 // Import and Set Nuxt.js options
 let config = require('../nuxt.config.js');
-config.dev = !(process.env.NODE_ENV === 'production');
+config.dev = !(process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'test');
```
rwd: What does this aim to achieve?

lbiedinger: Run the test server(localhost:1337) without development mode so that it doesn't display linting errors.

lbiedinger:
Won't fix

PR #83, opened 2019-03-20
Update e2e test syntax

lbiedinger:
* Await element presence before 'filling in' and click interactions, so that we don't need to explicitly test for these in the scenarios. (+removed all checks like this from scenarios)
* Visiting pages is a 'When' not a 'Given'
* Update language for the check that confirms presence of text in an element. (added 'the')
* Added a human readable getter for 'first page of results'

### DIFF
--- a/tests/features/search/pagination.feature
+++ b/tests/features/search/pagination.feature
@@ -13,16 +13,16 @@ Feature: Search pagination
     Then I should be on the `first page of results`
 
     When I open `/search?query=&page=0`
-    Then I should be on `first page of results`
+    Then I should be on the `first page of results`
 
     When I open `/search?query=&page=one`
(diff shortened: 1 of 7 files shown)


tests/features/search/pagination.feature
```diff
@@ -2,35 +2,35 @@ Feature: Search pagination
...
 
   Scenario: Invalid `page` param redirects to page 1
 
     When I open `/search?query=&page=-1`
-    Then I should be on `/search?query=&page=1`
+    Then I should be on the `first page of results`
 
     When I open `/search?query=&page=0`
-    Then I should be on `/search?query=&page=1`
+    Then I should be on `first page of results`
```
rwd: Include "the" in all of these `Then` steps.

PR #85, opened 2019-03-22
Stringify Cucumber step definitions

rwd:
Because they are easier to read and comprehend than regular expressions.

Also:
* Add a `{target}` Cucumber parameter type which matches Gherkin in backticks, e.g. `` `search box` ``
* Move step definitions to test/features/support/step-definitions.js with actual runners in step-runners.js


tests/features/support/step-runners.js
```diff
@@ -0,0 +1,65 @@
...
+const { url } = require('../config/nightwatch.conf.js').test_settings.default.globals;
+
+const pages = {
+  'home page': `${url}/`,
+  'search page': `${url}/search`,
+  'record page': `${url}/record${europeanaId()}`,
+  'first page of results': `${url}/search?query=&page=1`
+};
+
+function qaSelector(qas) {
```
lbiedinger: Let's give this a more understandable param name.
qaElementNames.
Also have the methods below that currently take a 'qas' param rename their param.

Maybe add some documentation to this function explaining it's responsible for returning the css selector.

tests/features/support/step-definitions.js
```diff
@@ -0,0 +1,38 @@
...
+
+defineStep('I browse/open/visit (a/an/the)( ){target}', (pageName) =>
+  i.openAPage(pageName));
+defineStep('I find/identify/see/spot (a/an/the)( ){target}', (qa) =>
+  i.seeATarget(qa));
+defineStep('I find/identify/see/spot (a/an/the)( ){target} in/on the {target}', (qa, parentQa) =>
+  i.seeATarget([qa, parentQa]));
+defineStep('I find/identify/see/spot (a/an/the)( ){target} with the text {string}', (qa, text) =>
+  i.seeATargetWithText(qa, text));
+defineStep('I can\'t/don\'t find/identify/see/spot (a/an/the)( ){target}', (qa) =>
```
lbiedinger: we might want a negative step like this for the more specific selector. "I can't see the 'logo' in the 'search result'"

lbiedinger:
```
defineStep('I can\'t/don\'t  find/identify/see/spot (a/an/the)( ){target} in/on the {target}', (qa, parentQa) =>
  i.doNotSeeATarget([qa, parentQa]));
```

tests/features/support/step-runners.js
```diff
@@ -0,0 +1,65 @@
...
+module.exports = {
+  openAPage: async function (pageName) {
+    await client.url(pageUrl(pageName));
+  },
+  clickOnTheTarget: async function (qas) {
+    const selector = qaSelector(qas);
+    await client.expect.element(selector).to.be.visible;
+    await client.click(selector);
+  },
+  seeATarget: async function (qas) {
```
lbiedinger: Are we sure these need to be async? If we want to check for a bunch of elements on the page at the same time, it would be faster to do so synchronously. The risk being that a check happens far too late, but I've never seen them run into the next scenario.

rwd: Removed from a few of them. All others need it or errors occur.

Total run time is roughly the same as on master. +/- a second depending on the run.

PR #93, opened 2019-04-05
Accessibility as feature

lbiedinger:


lbiedinger:
favour https://github.com/europeana/incubator-portal-vue-nuxt/pull/99

PR #99, opened 2019-04-16
Implementation axe

mirjamverloop:


assets/scss/variables.scss
```diff
@@ -6,18 +6,18 @@ $grey: #777;
 $extralightgrey: #f9f7f7;
 $lightgrey: #f1f1ee;
 $bodygrey: #ededed;
-$midgrey: #9b9b9b;
 $darkgrey: #4d4d4d;
 $black: #1a1a1a;
 
 $blue: #095f90;
 $darkblue: #293745;
+$darkblue: #095f90;
```
mirjamverloop: There are two darkblue's now.

components/generic/InfoMessage.vue
```diff
@@ -23,4 +23,8 @@
   .alert.alert-light {
     font-style: italic;
   }
+
+  .alert-light {
+    color: black;
```
mirjamverloop: You can add the color: black to the declaration above. 
And, let's use $black. That is slightly off black, but it is the "Europeana black".

components/generic/PaginationNav.vue
```diff
@@ -64,3 +64,17 @@
...
   };
 </script>
+
+
+<style lang="scss">
+  @import "./assets/scss/variables.scss";
+
+  .page-link {
+    color: black;
+  }
```
mirjamverloop: Use the $black here as well. 
Or is it then complaining about the contrast?

lbiedinger: works

components/record/MediaImage.vue
```diff
@@ -6,11 +6,13 @@
     <img
       :src="src"
       class="mw-100 mb-3"
+      :alt="link"
```
rwd: This sets the alt text to the edm:isShownAt (via the redirection API method)... definitely not good alternate text for an image. Either leave it blank or set it to something like "View this object at the provider's website"

layouts/default.vue
```diff
@@ -1,6 +1,12 @@
 <template>
   <div>
     <PageHeader />
+    <a
```
rwd: If we have such a link, then it should porbably be the very first tabbable element on the page, before even anything in the header. Also, there is no element with ID "main".

For now, please remove this and we can introduce it, refined, in a later iteration.

PR #102, opened 2019-04-23
View PDF link

mirjamverloop:


plugins/europeana/record.js
```diff
@@ -70,4 +72,23 @@ function getRecord(europeanaId, params) {
...
 }
 
+/**
+ * Find the content type of a file and if type is a pdf return the url
+ * @param {string} url of a file
+ * @return {Object} the url of the file
+ */
+function findPDFContent(file, webResources) {
+  let mimeType;
+  for (let resource in webResources) {
```
rwd: As `webResources` is an array, we can just use `for...of` here.

pages/record/_.vue
```diff
@@ -65,14 +74,15 @@
         error: null,
         image: null,
         fields: null,
-        media: null
+        media: null,
+        pdfLink: null
```
rwd: Should this be `pdf`?

plugins/europeana/record.js
```diff
@@ -70,4 +72,23 @@ function getRecord(europeanaId, params) {
...
+    if (file === webResources[resource].rdfAbout) {
+      mimeType = webResources[resource].ebucoreHasMimeType;
+    }
+  }
+
+  if (mimeType && mimeType.includes('pdf')) {
```
rwd: Let's do a strict match on `=== 'application/pdf'` as that is the only formal MIME type for PDFs.

PR #112, opened 2019-04-30
Related entities

mirjamverloop:


plugins/europeana/entity.js
```diff
@@ -50,4 +67,102 @@ function getEntityId(url) {
...
+ * Format the the entity data
+ * @param {Object} the lookuped data for entities
+ * @return {Object} entity links and titles
+ */
+function getRelatedEntityTitleLink(entities) {
+  let entityDetails = [];
+
+  for (let entity of entities) {
+    if (entity.prefLabel.en) {
+      const entityLink = '/entity/' + getEntityTypeHumanReadable(entity.type) + '/' + getEntityPath(entity.id.toString().split('/').pop(), entity.prefLabel.en);
```
rwd: Instead of constructing the full link here (and in the page view), return the entity type and path as distinct properties, then in views link/redirect to the [named route](https://router.vuejs.org/guide/essentials/named-routes.html) "entity-type-all".

PR #113, opened 2019-05-07
I18n for UI strings

lbiedinger:
Use the Nuxt-i18n package for loading strings.

Deprecated Europeana specific string lookups for consistency and less complex test setup. 

Added English, German(partial) and Spanish(empty).

Moved all strings


plugins/europeana/search.js
```diff
@@ -52,7 +54,7 @@ function display(field) {
 function fieldsForSearchResult(item) {
   let fields = {
     // TODO: fallback to description when API returns dcDescriptionLangAware
-    dcTitle: item.dcTitleLangAware ? display(item.dcTitleLangAware) : `No title provided for record ID ${item.id}`,
+    dcTitle: item.dcTitleLangAware ? display(item.dcTitleLangAware) : `${$t('messages.noTitle')}: ${item.id}`,
```
rwd: Use [named formatting](http://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting) for the item ID.

components/search/SearchSelectedFacets.vue
```diff
@@ -7,7 +7,7 @@
       class="mr-2"
       data-qa="filter badge"
     >
-      {{ selectedFacet.facetName | searchFacetHeader }}: {{ selectedFacet.fieldValue }}
+      {{ $t(`facets.${selectedFacet.facetName}`) }}: {{ selectedFacet.fieldValue }}
```
rwd: Use [named formatting](http://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting) with the colon and a placeholder for the facet field value included.

lbiedinger: that would make the translation unusable in the facet section headers, so would require unnecessary duplication of the same translation.

tests/unit/components/PageHeader.spec.js
```diff
@@ -23,4 +43,12 @@ describe('components/search/PageHeader', () => {
     const logo = wrapper.find('[data-qa="logo"]');
     logo.attributes().src.should.match(/\/logo\..+\.svg$/);
   });
+
+  describe('when ENV enables the language selector', () => {
+    describe('it contains a language selector', () => {
```
rwd: Let's also test the inverse, like:
```js
      const wrapper = factory();
      const selector = wrapper.findAll('[data-qa="language selector"]');
      selector.length.should.equal(0);
```

PR #116, opened 2019-05-14
Update of contentcard: creator + institution

mirjamverloop:


components/generic/ContentCard.vue
```diff
@@ -50,13 +79,40 @@
...
+    },
+    computed: {
+      linkToRecord () {
+        return this.localePath({ name: 'record-all', params: { pathMatch: this.recordId.replace(/^\/+/g, '') } });
+      },
+      backgroundImage () {
+        if (this.recordThumbnailUrl) {
+          return this.recordThumbnailUrl;
+        } else {
+          return (this.image && this.image.fields) ? this.image.fields.file.url : '';
```
rwd: I think that this component ought only need be aware of an image URL being passed to it, in an `imageUrl` prop. Provision of image URLs from different sources should be handled higher up the rendering hierarchy, perhaps even at plugin or page level.

components/browse/ContentCardSection.vue
```diff
@@ -19,13 +19,16 @@
...
+        v-for="card in cards"
         :key="card.sys.id"
         :name="card.fields.name"
         :description="card.fields.description"
         :url="card.fields.url"
-        :image-url="card.fields.image.fields.file.url"
-        :image-title="card.fields.image.fields.title"
+        :image="card.fields.image"
+        :creator="card.fields.creator"
+        :institution="card.fields.provider"
```
rwd: Unless there is a good reason not to, reuse the name of the property from `card.fields`, i.e. `:provider`. Same for the following two attributes.

components/generic/ContentCard.vue
mirjamverloop: I have moved it one level up to the content card section.

components/browse/ContentCardSection.vue
mirjamverloop: I have renamed these.

PR #120, opened 2019-06-04
Entity related Records

lbiedinger:


pages/entity/_type/_.vue
```diff
@@ -39,52 +71,126 @@
...
       };
     },
-    asyncData ({ env, params, res, redirect, app }) {
+    computed: {
+      hasResults: function () {
+        return this.searchResults.results !== null && this.searchResults.totalResults > 0;
+      },
+      entityQuery: function (params) {
+        return `http://data.europeana.eu/agent/base/${params.pathMatch}`;
+      }
```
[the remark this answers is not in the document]
lbiedinger: this is leftover from before, will remove.

components/search/SearchResultsList.vue
```diff
@@ -5,7 +5,7 @@
     <b-list-group-item
       v-for="result in results"
       :key="result.europeanaId"
-      :to="result.linkTo"
+      :to="localePath({ name: 'record-all', params: { pathMatch: result.europeanaId.replace(/^\/+/g, '') } })"
```
rwd: The `g` modifier seems redundant here given the `^` anchor.

To save using a regex, we could just do: 
```js
result.europeanaId.slice(1)
```

pages/entity/_type/_.vue
```diff
@@ -24,11 +24,43 @@
...
+        >
+          {{ $t('noMoreResults') }}
+        </p>
+        <SearchResultsList
+          v-else
+          :results="searchResults.results"
+        />
+        <InfoMessage
+          v-if="searchResults.lastAvailablePage"
+          message="Additional results are not shown as only the first 1000 most relevant results are shown. If you haven't found what you're looking for, please consider refining your search."
```
rwd: This should be in the locale file.

pages/entity/_type/_.vue
```diff
@@ -39,52 +71,123 @@
...
     data () {
       return {
         error: null,
         title: null,
-        relatedEntities: null
+        entity: null,
+        relatedEntities: null,
+        searchResults: {
+          error: null,
+          errorNoResults: 'No results',
```
rwd: In the locale file.

lbiedinger: actually it's not used here, I've removed it and moved the one in the search page(where this was copied from) into the locale file.

PR #121, opened 2019-06-06
Entity depiction and description

mirjamverloop:


plugins/europeana/entity.js
```diff
@@ -182,3 +182,39 @@ function getRelatedEntityTitleLink(entities) {
...
+ * If type is person, use biographicalInformation
+ * @return {String} a description when available in English
+ */
+export function getEntityDescription(type, entity) {
+  let description;
+
+  if (type === 'topic') {
+    description = entity.note && entity.note.en ? entity.note.en[0] : '';
+  } else if (type === 'person') {
+    if (entity.biographicalInformation.length !== undefined) {
```
rwd: Can cause "Cannot read property 'length' of undefined" error, e.g. at /entity/person/18037-todd-woods

plugins/europeana/entity.js
```diff
@@ -182,3 +182,39 @@ function getRelatedEntityTitleLink(entities) {
   }
   return entityDetails;
 }
+
+/** Get the description for the entity
```
rwd: Put the description of the function on the following line, i.e. `/**` should be on its own line.

plugins/europeana/entity.js
```diff
@@ -182,3 +182,39 @@ function getRelatedEntityTitleLink(entities) {
...
+ * If type is topic, use note
+ * If type is person, use biographicalInformation
+ * @return {String} a description when available in English
```
rwd: Document the params.

tests/unit/plugins/europeana/entity.spec.js
```diff
@@ -59,6 +72,13 @@ describe('plugins/europeana/entity', () => {
         const apiResponse = {
```
lbiedinger: just use ```const apiResponse = entitiesResponse.items[0];```

plugins/europeana/entity.js
rwd: Should be with @param, not @type. I have fixed.

PR #127, opened 2019-06-19
footer links from contentful

lbiedinger:


components/PageFooter.vue
```diff
@@ -1,9 +1,68 @@
 <template>
-  <footer>{{ $t('footerText') }}</footer>
```
renekok: remove footerText from locale files

components/PageFooter.vue
```diff
@@ -1,9 +1,68 @@
...
+          :key="footerLink.url"
+        >
+          <b-link
+            v-if="footerLink.url"
+            :href="footerLink.url"
+            class="footer-link"
+          >
+            {{ footerLink.text }}
+          </b-link>
+          <p v-else>
```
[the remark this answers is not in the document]
lbiedinger: it serves 2 functions as of now:
* a hacky way to output any errors, while no other error handling is in place
* future proofing for a potential case where footer elements are to be added that aren't links. (currently the content model wouldn't allow this anyway)

store/footer.js
```diff
@@ -0,0 +1,33 @@
+import contentfulClient from '../plugins/contentful.js';
+
+export const state = () => ({
+  links: null
```
rwd: Initialise this to `[]` instead of `null`

store/footer.js
```diff
@@ -0,0 +1,33 @@
...
+export const actions = {
+  async init ({ commit }) {
+    let data = await contentfulClient.getEntries({
+      'content_type': 'linkGroup',
+      'fields.identifier': 'footer',
+      'limit': 1
+    })
+      .then((response) => {
+        if (response.total == 0) {
+          return [];
```
rwd: This can be omitted if the default state value is made to be `[]`

store/index.js
```diff
@@ -0,0 +1,8 @@
+export const state = () => ({
```
rwd: This seems redundant and can be removed.

tests/unit/components/PageFooter.spec.js
```diff
@@ -1,17 +1,39 @@
...
-  it('it exists', () => {
+  it('it contains elements for each link', async () => {
     const wrapper = factory();
-
-    const footer = wrapper.find('footer');
-    footer.isVisible().should.equal(true);
+    await sleep(20);
+    const renderedList = wrapper.findAll('footer ul li');
+    renderedList.length.should.eq(2);
+    //dummyLinkGroup.items[0].links[0].fields.url
```
rwd: Remove this

tests/unit/components/PageFooter.spec.js
```diff
@@ -1,17 +1,39 @@
...
+};
+
 describe('components/search/PageFooter', () => {
-  it('it exists', () => {
+  it('it contains elements for each link', async () => {
     const wrapper = factory();
-
-    const footer = wrapper.find('footer');
-    footer.isVisible().should.equal(true);
+    await sleep(20);
```
rwd: Why is this needed? If I remove it, the test still passes.

store/footer.js
lbiedinger: it still needs to check for response.total == 0, otherwise it'll throw an error when someone deletes/unpublishes the 'footer' linkGroup.

tests/unit/components/PageFooter.spec.js
lbiedinger: Not needed any more, it was needed when the links were only retrieved on the client as that happened only after the component was rendered.

Have removed it.

nuxt.config.js
```diff
@@ -43,7 +43,7 @@ module.exports = {
   /*
   ** Plugins to load before mounting the App
   */
-  plugins: ['~/plugins/vue-filters'],
+  plugins: ['~/plugins/vue-filters', { src: './plugins/global.js' }],
```
rwd: Let's keep this consistent and concise, with:
```js
plugins: ['~/plugins/vue-filters', '~/plugins/global']
```

components/generic/Hyperlink.vue
```diff
@@ -0,0 +1,31 @@
...
+    :class="linkClass"
+  >
+    <slot />
+  </b-link>
+</template>
+
+<script>
+  export default {
+    props: {
+      link: {
```
rwd: Could we name this something more descriptive, perhaps "destination"?

plugins/global.js
```diff
@@ -0,0 +1,4 @@
+import Vue from 'vue';
+import Hyperlink from '../components/generic/Hyperlink.vue';
```
rwd: > Component names should always be multi-word

https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential

PR #133, opened 2019-06-28
Watch for search query changes

rwd:
When leaving the search results page, remove the user's search query from the form input in the header, but ensure it is restored if they return using their browser's back button.

* Moves reading of the query up to the layout level which passes it down to the header component
* Uses a Vue "navigation guard" to observe when the search page is left for another, per https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards
* Uses an "event hub" to emit global events, per https://github.com/nuxt/nuxt.js/issues/849#issuecomment-307016399

### DIFF
--- a/components/PageHeader.vue
+++ b/components/PageHeader.vue
@@ -69,7 +69,7 @@
     },
     data () {
       return {
-        query: this.searchQuery,
+        query: this.searchQuery ? this.searchQuery : '',
         isLoading: false
       };
     },
@@ -82,7 +82,7 @@
       searchQuery: {
         immediate: true,
         handler(val) {
-          this.query = val;
+          this.query = val ? val : '';
         }
       }
     },
--- a/components/PageHeader.vue
+++ b/components/PageHeader.vue
@@ -69,7 +69,7 @@
     },
     data () {
       return {
-        query: this.searchQuery ? this.searchQuery : '',
+        query: this.searchQuery || '',
         isLoading: false
       };
     },
@@ -82,7 +82,7 @@
       searchQuery: {
         immediate: true,
         handler(val) {
-          this.query = val ? val : '';
+          this.query = val || '';
         }
       }
     },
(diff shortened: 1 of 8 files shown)


tests/features/search/querying.feature
```diff
@@ -5,7 +5,8 @@ Feature: Search querying
     When I visit a `search page`
     And I enter "paris" in the `search box`
     And I click the `search button`
-    Then I see a `search result`
+    Then I see "paris" in the `search box`
```
lbiedinger: This is always failing for me when using the GeckoDriver:
```
Then I see "paris" in the `search box` # tests/features/support/step-definitions.js:44
    Error while running "getValue" command
    + expected - actual
 
    +paris

    AssertionError: Error while running "getValue" command: expected '' to equal 'paris'
        at NightwatchAPI.client.getValue (/incubator-portal-vue-nuxt/tests/features/support/step-runners.js:124:44)
        at transport.locateElement.then.then.catch.then.result (/incubator-portal-vue-nuxt/node_modules/nightwatch/lib/api-loader/element-command.js:144:30)
        at processTicksAndRejections (internal/process/next_tick.js:81:5)
```

Does it pass for you?
First I thought it might be a timing issue, but even a 4 second timeout doesn't help. Although I can see the text in the field while the test runs, so maybe something with the web driver not inspecting it correctly?

I'm on FF-67

components/PageHeader.vue
```diff
@@ -74,12 +78,17 @@
...
     methods: {
       submitSearchForm () {
-        this.$router.push(this.localePath({ name: 'search', query: { query: this.query ? this.query : '' } }));
-      },
-      getQueryFromParam () {
-        return this.$route.query ? this.$route.query.query : null;
+        this.$router.push(this.localePath({ name: 'search', query: { query: this.query } }));
       }
     },
     watchQuery: ['query']
```
lbiedinger: This can be removed as you noticed it's not doing anyting.

rwd: Thought I had done; will do.

tests/features/search/querying.feature
rwd: As discussed elsewhere, given the various issues we are experiencing with geckodriver, we will instead stop using it in `test:e2e` but leave it set up in the project should the problems be resolved upstream.

PR #135, opened 2019-07-02
Grid view with generic cards

lbiedinger:
Use refactored generic cards in for the grid view.
Grid view made default result listing.
Some minor layout tweaks.


components/generic/ContentCard.vue
```diff
@@ -157,10 +110,15 @@
...
+      min-height: 100%;
       color: $black;
       &:hover {
         color: $black;
       }
     }
+
+    .view-more {
+      float: right;
+    }
```
mirjamverloop: Let's keep it to the left, I think that is more consistent with the proposed design.

components/search/SearchResultsGrid.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <b-card-group
+    id="searchResults"
```
mirjamverloop: See first comment as well, use a additional CSS class instead of the id.

components/search/SearchResultsGrid.vue
```diff
@@ -0,0 +1,52 @@
...
+      results: {
+        type: Array,
+        default: () => []
+      }
+    },
+    methods: {
+      cardTexts: function (result) {
+        let texts = [];
+        if (Array.isArray(result.fields.dcCreator)) {
+          let creator =  result.fields.dcCreator.slice(0, 3).join(this.$t('formatting.listSeperator') +' ');
```
mirjamverloop: For readability (and linting?) add a space between the + and the '  '.

components/search/SearchResultsGrid.vue
```diff
@@ -0,0 +1,52 @@
...
+            creator = creator + this.$t('formatting.listSeperator') + this.$t('formatting.elipsis');
+          }
+          texts.push(creator);
+        }
+        if (Array.isArray(result.fields.edmDataProvider)) {
+          let provider = result.fields.edmDataProvider.slice(0, 3).join(this.$t('formatting.listSeperator') +' ');
+          if (result.fields.edmDataProvider.length > 3) {
+            provider = provider + this.$t('formatting.listSeperator') + this.$t('formatting.elipsis');
+          }
+          texts.push(provider);
```
mirjamverloop: The code for getting the creator and provider is in fact the same, is it possible to create and use a function?

lang/en.js
```diff
@@ -1,6 +1,8 @@
 export default {
   formatting: {
-    labelledValue: '{label}: {value}'
+    labelledValue: '{label}: {value}',
+    elipsis: '…',
```
mirjamverloop: Elipsis is with double ll, ellipsis.

tests/unit/components/generic/ContentCard.spec.js
```diff
@@ -15,15 +17,15 @@ const factory = () => mount(ContentCard, {
...
+    wrapper.setProps({ texts: ['The Milkmaid by Vermeer'] });
 
     const description =  wrapper.find('[data-qa="content card"] .card-body');
     description.text().should.eq('The Milkmaid by Vermeer');
   });
 
   it('has a creator and institution', () => {
     const wrapper = factory();
-    wrapper.setProps({ creator: 'Edvard Munch', provider: 'Munchmuseet (The Munch Museum)' });
+    wrapper.setProps({ texts: ['Edvard Munch','Munchmuseet (The Munch Museum)'] });
```
mirjamverloop: Add a space after the ,

components/browse/ContentCardSection.vue
```diff
@@ -57,6 +55,31 @@
...
+    methods: {
+      moreLabelKey: function (card) {
+        return card.fields.identifier ? 'goToRecord' : 'readMore';
+      },
+      cardDestination: function (card) {
+        if (card.fields.url) {
+          return card.fields.url;
+        } else if (card.fields.identifier) {
+          return this.localePath({ name: 'record-all', params: { pathMatch: card.fields.identifier.slice(1) } });
+        }
```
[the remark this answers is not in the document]
lbiedinger: It'll pass null to the card and there the default URL is the europena home page.

components/browse/ContentCardSection.vue
```diff
@@ -57,6 +55,31 @@
           return card;
         });
       }
+    },
+    methods: {
+      moreLabelKey: function (card) {
+        return card.fields.identifier ? 'goToRecord' : 'readMore';
```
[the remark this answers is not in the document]
lbiedinger: perhaps at that point we'll want to allow for providing a literal value which can be translated in contentful, as keys would require separate translation updates.

components/search/SearchResultsGrid.vue
rwd: Add a space after the `+`

lang/en.js
rwd: "ellipsis"

components/search/SearchResultsGrid.vue
rwd: These should probably be marked up as lists in the HTML, with styling achieved via CSS.

components/search/SearchResult.vue
```diff
@@ -29,11 +29,16 @@
         </template>
         <ul v-else>
           <li
-            v-for="(element, index) in value"
+            v-for="(element, index) in value.slice(0, 3)"
```
rwd: Let's have this be a computed method instead of doing it here in the template.

components/search/SearchResult.vue
```diff
@@ -29,11 +29,16 @@
...
             :key="index"
           >
             {{ element }}
           </li>
+          <li
+            v-if="value.length > 3"
```
rwd: Let's have another computed method returning boolean for this.

components/search/SearchResult.vue
```diff
@@ -29,11 +29,16 @@
         </template>
```
rwd: The name of this component is a bit misleading now as it's not always used for search results. Perhaps rename to something like `ListSearchResult.vue`

components/browse/ContentCardSection.vue
```diff
@@ -57,6 +55,31 @@
           return card;
         });
       }
+    },
+    methods: {
+      moreLabelKey: function (card) {
```
rwd: When we end up having many methods all taking the same card parameter, it points to the need for a new component scoped to one of those cards.

components/generic/ContentCard.vue
```diff
@@ -83,25 +51,9 @@
...
-        default: ''
-      },
-      provider: {
-        type: String,
-        default: ''
-      },
-      identifier: {
+      viewMoreLabelKey: {
         type: String,
-        default: ''
```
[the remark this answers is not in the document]
lbiedinger: this is set to null to allow a blank label.

components/search/SearchResultsGrid.vue
lbiedinger: Let's do that later.

components/search/SearchResult.vue
lbiedinger: I think that's more confusing with us also having a component SearchResultsList.

components/browse/ContentCardSection.vue
lbiedinger: yes, certainly something we could refactor by adding these values to contentful. Let's do so later when we have more of the cards ready, including entity cards and potential changes coming out of the Open Graph work.

components/search/SearchResult.vue
rwd: Fair point. Let's leave it for now and revisit if and when there is a case for reuse of this component, e.g. as a general record metadata + thumbnail preview component.

components/search/SearchResultsGrid.vue
rwd: Agreed.

PR #141, opened 2019-07-03
Persist search results `view` with params and storage

lbiedinger:
Use params and local/session storage to track views.

This PR also integrates all components and adds:
* Feature tests
* Styling for distinguishing active/inactive view toggles
* Stop emitting events when toggle doesn't change the value

### DIFF
--- a/components/search/ViewToggles.vue
+++ b/components/search/ViewToggles.vue
@@ -9,13 +9,12 @@
       :active="activeView == view"
       :data-qa="`search ${view} view toggle`"
       class="pl-3"
-      @click="selectView"
+      @click="selectView(view)"
     >
       <img
         :src="iconSrc(view)"
         :alt="$t(`searchViews.${view}`)"
         :title="$t(`searchViews.${view}`)"
-        :data-view="view"
       >
     </b-nav-item>
   </b-nav>
@@ -45,9 +44,9 @@
           name: 'search', query: { ...this.$route.query, ...{ view: view } }
         });
       },
-      selectView: function (event) {
-        if (event.target.getAttribute('data-view') !== this.activeView) {
-          this.activeView = event.target.getAttribute('data-view');
+      selectView: function (view) {
+        if (view !== this.activeView) {
+          this.activeView = view;
           this.$emit('changed', this.activeView);
         }
       }
--- a/components/search/ViewToggles.vue
+++ b/components/search/ViewToggles.vue
@@ -46,8 +46,10 @@
         });
       },
       selectView: function (event) {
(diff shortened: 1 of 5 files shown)


pages/search/index.vue
```diff
@@ -291,7 +297,21 @@
         this.rerouteSearch({ qf: this.qfForSelectedFacets, reusability: this.reusability, page: '1' });
       },
       selectView (view) {
+        if (process.browser) {
+          sessionStorage.view = view;
```
rwd: As discussed in person, let's preempt the likely need to make the local/session storage key more specific, by creating a top-level `settings` object like:
```js
// in an init function somewhere:
sessionStorage.settings = sessionStorage.settings || {};
sessionStorage.settings.results = sessionStorage.settings.results || {};
sessionStorage.settings.results.view = sessionStorage.settings.results.view || 'grid';
```

pages/search/index.vue
```diff
@@ -291,7 +297,21 @@
...
         this.view = view;
+      },
+      selectedView: function () {
+        if (process.browser) {
+          if (this.$route.query.view) {
+            sessionStorage.view = this.$route.query.view;
+          }
+          return sessionStorage.view || localStorage.view || 'grid';
+        }
+        return 'grid';
```
rwd: Should this return `this.view`?

pages/search/index.vue
lbiedinger: this won't work unless we also add a mechanism for parsing sessionStorage values as they only take 'simple' values. See https://vuejs.org/v2/cookbook/client-side-storage.html#Working-with-Complex-Values
I think adding that at this point is out of scope especially as we have only one thing to store there for now.

I have however renamed 'view' to 'searchResultsView' so as to make the name more specific and thereby less likely to interfere with any future use we may have for "sessionStorage.view"

pages/search/index.vue
lbiedinger: I think that could lead to an infinite loop, but I've made it aware of any setting in the URL params so as to prevent the observed issue of always rendering 'grid' server side.

components/search/ViewToggles.vue
```diff
@@ -46,8 +46,10 @@
         });
       },
       selectView: function (event) {
-        this.activeView = event.target.getAttribute('data-view');
-        this.$emit('changed', this.activeView);
+        if (event.target.getAttribute('data-view') !== this.activeView) {
```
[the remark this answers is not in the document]
lbiedinger: As discussed passing the value into the method seems a more standard way of keeping track of which button does what.

PR #143, opened 2019-07-09
View param from all pages

lbiedinger:


pages/search/index.vue
```diff
@@ -264,6 +264,10 @@
           view: this.view
         };
 
+        if (current.reusability === null) {
```
rwd: Let's combine this with the following, which should future-proof for other parameters:
```js
      updateCurrentSearchQuery(updates) {
        const current = {
          page: this.page || '1',
          qf: this.qfForSelectedFacets,
          query: this.query || '',
          reusability: this.reusability,
          view: this.view
        };

        const updated = { ...current, ...updates };

        // If any values are now `null`, remove them from the query
        for (const key in updated) {
          if (updated[key] === null) {
            delete updated[key];
          }
        }

        return updated;
      }
```

components/PageHeader.vue
```diff
@@ -88,7 +88,9 @@
     },
     methods: {
       submitSearchForm () {
-        const newSearchQuery = { ...this.searchQuery, ...{ query: this.query } };
+        const view = this.$route.query.view || sessionStorage.searchResultsView || localStorage.searchResultsView || 'grid';
```
rwd: I think we probably don't want this component to need be concerned with the source of the various parameters other than `query`. How about instead we add to layouts/default.vue:
```js
    mounted () {
      if (!this.searchQuery.hasOwnProperty('view')) {
        this.searchQuery.view = sessionStorage.searchResultsView || localStorage.searchResultsView || 'grid';
      }
    }
```

PR #145, opened 2019-07-09
Thematic collections facet

mirjamverloop:
- The facet component can handle radio buttons as well.
- The actual themes are hard-coded for now. 
- The theme facet is the first one in the column.

### DIFF
--- a/components/search/SearchSelectedFacets.vue
+++ b/components/search/SearchSelectedFacets.vue
@@ -13,6 +13,9 @@
 </template>
 
 <script>
+
+  import { thematicCollections } from '../../plugins/europeana/search';
+
   export default {
     props: {
       facets: {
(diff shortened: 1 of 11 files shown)


pages/search/index.vue
```diff
@@ -141,6 +142,21 @@
     };
   }
 
+  const thematicCollections = [ { label: 'All items', value: '' },
```
rwd: The labels should go into lang/en.js

components/search/SearchSelectedFacets.vue
```diff
@@ -24,8 +24,16 @@
       facetList: function() {
         let listOfFacets = [];
         for (let facetName in this.facets) {
+
```
rwd: In this component, the filter badge text should use the _label_ of the theme, not its query value, e.g. "1914-1918" not "ww1"

components/search/SearchSelectedFacets.vue
```diff
@@ -7,7 +7,12 @@
       class="mr-2"
       data-qa="filter badge"
     >
-      {{ $t('formatting.labelledValue', { label: $t(`facets.${selectedFacet.facetName}`), value: selectedFacet.fieldValue}) }}
+      <template v-if="selectedFacet.facetName === 'THEME'">
```
rwd: Let's make this more reusable by switching to:
```js
<template v-if="$te(`facets.${selectedFacet.facetName}.options`)">
```

mirjamverloop: Good suggestion, thanks. I have updated the code. 

I might add a test with options as well, and update the mocked $te.

PR #153, opened 2019-07-31
Static hero banner

mirjamverloop:


components/generic/HeroBanner.vue
```diff
@@ -0,0 +1,18 @@
+<template>
```
rwd: Let's add a <script> block already with the hard-coded values as defaults for props.

assets/scss/hero.scss
```diff
@@ -0,0 +1,87 @@
+.jumbotron {
+  background-image: url('https://proxy.europeana.eu/90402/SK_A_1709?api_url=https%3A%2F%2Fapi.europeana.eu%2Fapi&view=https%3A%2F%2Flh3.googleusercontent.com%2FItR9rLYOjTXu2zdHVxeY2sgB3INSKXDzGVLQHjEe59IBknUFKsPX-AdxAZHLOLXnWw42ADN8n63BJIEqAdQo8kT0dMk%3Ds0'); // temp image
```
monkemedia: New to the project, so please correct me if I am wrong. Wonder if 'background-image' should be inline instead, that way we can easily change the image from the backend?

mirjamverloop: I was planning to do that as part of one of the other subtasks (EC-3646), but it makes sense to move it inline already as part of this work.

monkemedia: My bad :/ Will stay quiet next time 😄

PR #160, opened 2019-08-07
Lokalise POC

mirjamverloop:


.eslintrc.js
```diff
@@ -48,6 +48,12 @@ module.exports = {
       'rules': {
         'indent': 'off'
       }
+    },
+    {
+      "files": ["lang/*.js"],
```
rwd: Let's use single quotes here and below.

PR #173, opened 2019-08-19
Hero banner and rights statement

mirjamverloop:
- The hero banner only shows on browse page where a banner is selected
- The banner itself displays content from Contentful
- There is logic to show a readable name and icon(s) for the rights statement
- Tests


pages/_.vue
```diff
@@ -1,6 +1,14 @@
 <template>
   <div>
-    <HeroBanner />
+    <HeroBanner
+      v-if="page.primaryImageOfPage"
+      :hero-image="page.primaryImageOfPage.fields.image.fields.file.url"
+      :headline="page.primaryImageOfPage.fields.headline"
+      :description="page.primaryImageOfPage.fields.description"
+      :identifier="page.primaryImageOfPage.fields.identifier"
```
rwd: The record identifier coming from Contentful will start with a slash, resulting in two slashes in the record URL link from the hero banner. This needs to be taken into account somewhere.

assets/scss/licenses.scss
```diff
@@ -0,0 +1,47 @@
+.icon-license-pd {
+  background-image: url('../../assets/img/cc-icons/pd.svg');
```
monkemedia: Do the paths still work if we use `~assets` instead of `../../assets`? If so, it will be cleaner plus we could drop the file into any directory without breaking the paths

components/generic/RightsStatement.vue
```diff
@@ -0,0 +1,117 @@
...
+<script>
+  export default {
+    props: {
+      rightsStatementUrl: {
+        type: String,
+        default: ''
+      }
+    },
+    computed: {
+      rights: function () {
```
monkemedia: Wonder if this could be a little more descriptive?

also, we can write the function as `rights () {}`. Will keep it more consistent throughout.

components/generic/RightsStatement.vue
```diff
@@ -0,0 +1,117 @@
...
+        case new RegExp('https?://www.europeana.eu/rights/out-of-copyright-non-commercial').test(val):
+          rights.name = 'No Copyright - Non-Commercial Use Only';
+          rights.iconClass = ['icon-license-rs-no'];
+          break;
+        case new RegExp('https?://rightsstatements.org/vocab/InC-EDU/*').test(val):
+          rights.name = 'In Copyright - Educational Use Permitted';
+          rights.iconClass = ['icon-license-rs-yes'];
+          break;
+        case new RegExp('https?://rightsstatements.org/vocab/NoC-NC/*').test(val):
+          rights.name = 'No Copyright - Non-Commercial Use Only';
```
monkemedia: Will these strings be localized (maybe another ticket for that I guess)?

pages/_.vue
mirjamverloop: It's in HeroBanner.vue, similar to how this is done on the cards.

assets/scss/licenses.scss
mirjamverloop: I have updated this. It used to be ../../ to get it working in Storybook, but since we are not using that actively, I have updated it in this file. I will create a technical debt task to update this in other CSS files.

components/generic/RightsStatement.vue
mirjamverloop: I have updated the name of the function and I agree to keep it consistent, but we are using nameOfFunction: function () {} for almost every computed property functions.

components/generic/RightsStatement.vue
mirjamverloop: Yes, one for later.

PR #175, opened 2019-08-19
Previews from contentful preview API

lbiedinger:
* requires setting env variable `CTF_CPA_ACCESS_TOKEN` which shouldn't be in the .env file but rather set on the machine itself as to avoid the client from being aware of it.
* only for browse pages
* doesn't include footer content


.env.example
```diff
@@ -1,5 +1,6 @@
 CTF_SPACE_ID="ABC123"
 CTF_CDA_ACCESS_TOKEN="ABCDEF123456"
+CTF_PREVIEW_CDA_ACCESS_TOKEN="GHIJKLM7891011"
```
rwd: Contentful abbreviate the Preview API as CPA. Let's therefore name this `CTF_CPA_ACCESS_TOKEN`.

pages/_.vue
```diff
@@ -16,14 +16,20 @@
...
 
   export default {
     components: {
       ContentCardSection,
       HeroBanner
     },
-    asyncData ({ params, error, app }) {
+    asyncData ({ params, query, error, app }) {
+      let contentfulClient;
+      if (query.mode == 'preview' && process.env['CTF_CPA_ACCESS_TOKEN']) {
```
monkemedia: Should we be using strict operators in this project?

plugins/contentful.js
```diff
@@ -2,13 +2,17 @@ const contentful = require('contentful');
 
 // These values will be set via env vars.
 // If this file is imported these values may be available to the client.
-const config = {
+let config = {
```
monkemedia: I think we can still use a CONST here as we are not re-assigning or re-declaring the variable, however, I'm sure your eslint is probably more accurate

plugins/contentful.js
```diff
@@ -2,13 +2,17 @@ const contentful = require('contentful');
...
   space: process.env.CTF_SPACE_ID,
   accessToken: process.env.CTF_CDA_ACCESS_TOKEN
 };
 
-function createClient () {
+export function createClient (mode) {
   let client;
+  if (mode == 'preview') {
```
monkemedia: Strict?

PR #176, opened 2019-08-20
HTML-supported video detection

rwd:
* Creates a new component responsible for media presentation with logic to detect how to handle various media
* Moves PDF link detection out of Europeana Record API plugin and into the new component
* Adds HTML-supported video detection
* Renders the video player component for such videos


components/record/MediaPresentation.vue
```diff
@@ -0,0 +1,69 @@
...
+        type: String,
+        default: ''
+      }
+    },
+    computed: {
+      displayImage: function() {
+        return (this.imageSrc !== '') && !this.isHTMLVideo;
+      },
+      isPDF: function() {
+        return this.mimeType == 'application/pdf';
```
monkemedia: Should we be using strict operators here?

PR #179, opened 2019-08-21
Lokalise: Translations update

EuropeanaDevelopment:
This pull request was initiated by Lokalise (user Lutz Biedinger) at 2019-08-21 12:52:54


lbiedinger:
Using wrong placeholder variable names, these shouldn't be translated.

PR #181, opened 2019-08-21
Feat/ec 3685 create video player

monkemedia:


components/generic/LanguageSelector.vue
```diff
@@ -71,7 +71,7 @@
 
   .language-icon {
     top: 7px;
-    width: 18px;
+    width: 31px;
     height: 28px;
```
lbiedinger: The height needs a few extra pixels too. There's an arrow under the icon that's currently just cut off.

monkemedia:
Wrong brnach :/

PR #184, opened 2019-08-22
Default to exclude content tier 0

monkemedia:


plugins/europeana/search.js
```diff
@@ -171,6 +171,18 @@ function search(params) {
   const start = ((page - 1) * perPage) + 1;
   const rows = Math.max(0, Math.min(maxResults + 1 - start, perPage));
 
+  const qfhandler = (qf) => {
```
lbiedinger: rather than making this a const in here, add it as it's own function like this:

```
function qfhandler(qf) {
  let newQf = qf ? [].concat(qf) : [];
  newQf.push('contentTier:(1 OR 2 OR 3 OR 4)');
  return newQf;
}
```
note also, that'll handle 3 cases of qf being undefined, a string or an array.

lbiedinger: that should also reduce the method complexity that codeclimate raised.

plugins/europeana/search.js
```diff
@@ -171,6 +171,12 @@ function search(params) {
   const start = ((page - 1) * perPage) + 1;
   const rows = Math.max(0, Math.min(maxResults + 1 - start, perPage));
 
+  function qfhandler(qf) {
+    const newQf = qf ? [].concat(qf) : [];
+    newQf.push('contentTier:(1 OR 2 OR 3 OR 4)');
```
rwd: Let's add a comment above this line to document what we are doing here, i.e.
```js
    // Tier 0 content is excluded by default as they are considered not to meet
    // Europeana's publishing criteria.
```

PR #186, opened 2019-08-22
Set width and height on video element

rwd:
Also stop sending specific fields from web resources in API response, and just send the full objects, but whitelist on the web resources component the fields to display.

### DIFF
--- a/components/record/WebResources.vue
+++ b/components/record/WebResources.vue
@@ -2,8 +2,8 @@
   <b-list-group>
     <b-list-group-item
       v-for="webResource in media"
-      :id="webResource.rdfAbout"
-      :key="webResource.rdfAbout"
+      :id="webResource.about"
+      :key="webResource.about"
       data-qa="web resource"
       class="mb-3"
     >
--- a/components/record/WebResources.vue
+++ b/components/record/WebResources.vue
@@ -31,7 +31,7 @@
         default: () => []
       }
     },
-    data: function () {
+    data() {
       return {
         fields: ['about', 'dcDescription', 'edmRights', 'ebucoreHasMimeType']
       };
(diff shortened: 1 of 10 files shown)


components/record/WebResources.vue
```diff
@@ -30,6 +30,11 @@
         type: Array,
         default: () => []
       }
+    },
+    data: function () {
```
monkemedia: data () instead of data: function () ?

tests/unit/components/record/WebResources.spec.js
```diff
@@ -15,14 +15,14 @@ describe('components/record/WebResources', () => {
...
-        { rdfAbout: 'http://www.example.org/video.mp4', edmRights: { 'def': [ 'http://creativecommons.org/licenses/by-nc-sa/3.0/' ] } }
+        { about: 'http://www.example.org/image.jpg', dcDescription: ['Art'] },
+        { about: 'http://www.example.org/video.mp4', edmRights: { 'def': [ 'http://creativecommons.org/licenses/by-nc-sa/3.0/' ] } }
       ]
     };
 
     wrapper.setProps(props);
     for (let webResource of props.media) {
-      const webResourceBlock = wrapper.find('#' + cssesc(webResource.rdfAbout, { 'isIdentifier': true }));
+      const webResourceBlock = wrapper.find('#' + cssesc(webResource.about, { 'isIdentifier': true }));
```
monkemedia: Template literals here instead?

rwd: Do you mean `isIdentifier: true`?

rwd: That may need excessive escaping due to the presence of { and } in the code.

PR #189, opened 2019-08-23
Middleware redirect for homepage preview; catch empty images on cards

lbiedinger:


middleware/home-preview.js
```diff
@@ -0,0 +1,7 @@
+export default function ({ route, redirect }) {
+  // Contentful uses '/' as a slug for the home page, and iterpolates
```
rwd: "interpolates"

nuxt.config.js
```diff
@@ -241,6 +241,7 @@ module.exports = {
     // See https://github.com/nuxt-community/axios-module#options
   },
   router: {
+    middleware: ['home-preview'],
```
rwd: Let's move this middleware to `middleware/router/root-double-slash-redirect.js` to:
1. Organise the middleware into subdirectories from the outset
2. Name the file after its function rather than the specific circumstances motivating its introduction

middleware/home-preview.js
rwd: Also, reword slightly as it's not Contentful that uses it, but our content model in Contentful.

PR #190, opened 2019-08-26
Display message advising tier 0 content displayed

monkemedia:


components/search/TierToggler.vue
```diff
@@ -0,0 +1,90 @@
...
+      </button>.
+    </p>
+  </div>
+</template>
+
+<script>
+  export default {
+    name: 'TierToggler',
+
+    props: {
```
lbiedinger: Since the toggle is always for the content tier, it shouldn't be necessary to pass in all these props, they're always going to be the same.
The only thing we should need to pass in is the state/active status, in case we want to initialize the component where tier 0 is already being shown. For the case where someone loads the URL with it active.
The rest of the attributes can be computed properties based on the state.
Perhaps the component needs to know the query applied on the current page in order to generate the `to` link?

components/search/TierToggler.vue
```diff
@@ -0,0 +1,90 @@
...
+      toggled: {
```
lbiedinger: I think toggled is a little confusing since semantically either state could be toggled.
Have it be `activeState`

components/search/TierToggler.vue
```diff
@@ -0,0 +1,90 @@
...
+      },
+      queryKey: {
+        type: String,
+        default: 'tier_zero'
+      }
+    },
+
+    data () {
+      return {
+        toggle: false
```
lbiedinger: This should be `active` instead of `toggle`

pages/search/index.vue
```diff
@@ -214,6 +237,10 @@
     watch: watchList,
     asyncData ({ env, query, res, redirect, app }) {
       const currentPage = pageFromQuery(query.page);
+      const qf = query.qf;
```
lbiedinger: I don't think any of this belongs here. The tier toggle is already in qfForSelectedFacets if it's present, which is also then wired up with updateCurrentSearchQuery to update the search results when the qf changes and keep the values of the qfs in selectedFacets.

pages/search/index.vue
```diff
@@ -104,6 +104,16 @@
               />
             </b-col>
           </b-row>
+          <b-row>
+            <b-col>
+              <TierToggler
+                :text="tier.text"
+                :button="tier.button"
+                :toggled="tierToggle"
```
lbiedinger: don't use `tierToggle` here, rather get the value from the selectedFacets. Perhaps in a computed method if it's too complex to just pull out of the selectedFacets. It'll also only want to be true if the selected value is actually `*`.

pages/search/index.vue
```diff
@@ -263,6 +291,13 @@
           return { results: null, error: errorMessage, query: query.query };
         });
     },
+    mounted () {
```
lbiedinger: This shouldn't be needed if the toggle is wired up with the selectedFacets instead of relying on the matches.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    mounted () {
+      this.toggle = this.toggled;
+    },
+
+    methods: {
+      toggleHandler () {
+        if (!this.toggle) {
+          this.$emit('click', 'contentTier', ['(*)']);
```
lbiedinger: leave this to be just `*` without the parentheses. It's not actually helping with the stripping out, it just delays it one time. The problem is that the value for the contentTier facet is not in selectedFacets properly. You can see this if you use the Vue inspect tools on the page components.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <div>
+    <p class="font-italic">
+      {{ toggleText }}
+      <button
```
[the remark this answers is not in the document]
monkemedia: Where should it direct the user to?

plugins/europeana/search.js
```diff
@@ -132,9 +132,10 @@ export function selectedFacetsFromQuery(query) {
   let selectedFacets = {};
   if (query.qf) {
     for (const qf of [query.qf].flat()) {
+      const reqex = /^".*"$/; // Find double quotes at beginning and end of string
       const qfParts = qf.split(':');
       const facetName = qfParts[0];
-      const facetValue = qfParts[1].slice(1, -1);
+      const facetValue = qfParts[1].match(reqex) ? qfParts[1].slice(1, -1) : qfParts[1]; // Slice only if double quotes exist
```
lbiedinger: just put  `/^".*"$/` in here directly, no need for an additional assignment

components/search/TierToggler.vue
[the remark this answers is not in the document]
lbiedinger: to the current search page with all the query params +/- the `qf=contentTier:*` param depending on the current state.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,89 @@
...
+    mounted () {
+      if (this.activeState) {
+        this.active = this.activeState;
+      }
+    },
+
+    methods: {
+      toggleHandler () {
+        if (!this.active) {
+          this.$emit('click', 'contentTier', ['*']);
```
lbiedinger: click is quite generic. I suggest either, making this consistent with how facets are applied and emitting 'changed', or emitting 'toggle'.
The function being picked up on in pages/search/index.vue will need to be updated accordingly.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div>
+    <p class="font-italic">
+      {{ toggleText }}
+      <nuxt-link
+        :to="$route.fullPath"
```
lbiedinger: the link will need to lead to the toggled version not the current path.

components/search/TierToggler.vue
monkemedia: Made this 'toggle'

components/search/TierToggler.vue
monkemedia: this should now be set up correctly

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+  <div>
+    <p class="font-italic">
+      {{ toggleText }}
+      <nuxt-link
```
mirjamverloop: We have used `<b-link>` on several other components, which will use a `<nuxt-link>` if you specific the to.  
https://bootstrap-vue.js.org/docs/components/link/

I am inclined to say: keep it consistent and use the `<b-link>` here as well.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
...
+        {{ !active ? button.show : button.hide }}.
+      </nuxt-link>
+    </p>
+  </div>
+</template>
+
+<script>
+  export default {
+    name: 'TierToggler',
+
```
mirjamverloop: This is something we might want to discuss as a team as well when talking about consistency: having an extra line between options or not.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
...
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import "./assets/scss/variables.scss";
+
+  .toggle-button {
+    padding: 0;
+    font-style: italic;
```
mirjamverloop: This one is not necessary, as its parent is already set to italic.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import "./assets/scss/variables.scss";
+
+  .toggle-button {
+    padding: 0;
```
mirjamverloop: This can be removed.

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
...
+    font-style: italic;
+    height: auto;
```
mirjamverloop: In general, this CSS seems to be a leftover. Except for the "text-decoration" it all seems unnecessary. Can you double check that and remove?

components/search/TierToggler.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+  <div>
+    <p class="font-italic">
```
mirjamverloop: Let's add a "mb-5" class here, to make the paragraph more distinct from the pagination and having it more in line with the design.

components/search/TierToggler.vue
monkemedia: I prefer the extra line as I believe it makes the code look cleaner and easier to read. But I am happy to go with the majority.

PR #192, opened 2019-08-28
Use "home" as the homepage slug

rwd:


tests/features/pages/home-page.feature
```diff
@@ -4,9 +4,9 @@ Feature: Home page
   a curated page is available at the "http(s)://[domain]/" URL.
   This page should allow preview functionality.
 
-  Scenario: Previewing the homepage
-    Contentful will link to "//?mode=preview" as the slug for the homepage there is simply '/'
+  Scenario: Viewing the homepage
+    The homepage is a browse page from Contentful
 
-    When I open `//?mode=preview`
-    Then I should be on `/?mode=preview`
```
lbiedinger: we should still verify that ?mode=preview can be loaded.
This test as it is, is just a duplicate of our browse-page test.

rwd: What this _is_ testing is that the root path results in a browse page being loaded, despite our not having a browse page content entry with an empty slug in Contentful. This is different behaviour to the other browse pages which do have entries with their requested URL slug.

Testing `?mode=preview` in this context is no longer anything specific to the homepage. Also, we do not have a CPA env var set on Travis.

lbiedinger: good point about not having previews available in tests.
We should keep the preview in mind for when we stub API calls, until then it doesn't make sense.

PR #199, opened 2019-09-03
oEmbed

lbiedinger:


components/media/OEmbedMedia.vue
```diff
@@ -1,19 +1,29 @@
 <template>
   <!-- eslint-disable vue/no-v-html -->
   <div
+    v-if="oEmbedData && oEmbedData.html"
     data-qa="oembed media container"
-    v-html="html"
+    v-html="oEmbedData.html"
   />
+  <div v-else>
+    <AlertMessage
```
mirjamverloop: The `v-else` can probably go on the `<AlertMessage>` directly.

components/media/OEmbedMedia.vue
```diff
@@ -1,19 +1,29 @@
...
     name: 'OEmbedMedia',
-
+    components: {
+      AlertMessage
+    },
     props: {
-      html: {
-        type: String,
+      oEmbedData: {
+        type: Object,
```
mirjamverloop: Accept Null as well to avoid Vue warnings?

tests/unit/components/record/MediaPresentation.spec.js
```diff
@@ -95,7 +95,7 @@ describe('components/record/MediaPresentation', () => {
 
         wrapper.setProps(props);
 
-        wrapper.vm.isOEmbed.should.be.true;
+        wrapper.vm.isOEmbed.should.be.false;
```
lbiedinger: instead of testing that youtube is false, let's test for an actual oEmbed provider we do support.

components/record/MediaPresentation.vue
```diff
@@ -24,18 +24,17 @@
...
       :height="height"
     />
-    <div
+    <OEmbedMedia
       v-else-if="isOEmbed"
-    >
-      <strong>oEmbed</strong>
-      <div><code>{{ oEmbedData }}</code></div>
-    </div>
+      :o-embed-data="oEmbedData"
```
rwd: This component should accept two typed props:
1. `html`
2. `error`

PR #203, opened 2019-09-05
BrowseChip API request is now in a the BrowseChip component

monkemedia:


rwd:
I have concerns about this approach:
1. Two API requests are still made during pagination whereas the aim is to reduce it to one.
2. Moving the API request into the component makes it client-side only.
3. The BrowseChip component is now very specific to related entities and as such is misnamed and of no reusability. (Also, it should not have the entity route name hard-coded into it but accept a full router link as a prop.)

rwd:
Superceded by #204

PR #206, opened 2019-09-06
Moved BrowseChip badges into right hand column.

monkemedia:


plugins/europeana/entity.js
```diff
@@ -161,7 +161,7 @@ function getDataForEntities(entities, entityKey) {
   })
     .then((response) => {
       let items = response.data.items ? response.data.items : [];
-      return getRelatedEntityTitleLink(items.slice(0,10));
+      return getRelatedEntityTitleLink(items.slice(0,4));
```
rwd: Could we slice earlier to prevent retrieving more data from the entity search than we need, i.e.
```js
const entityLabels = entities.slice(0,4).map(entity => {
```
(Note also that `entityLabels` should use `const` instead of `let`.)

components/browse/BrowseChip.vue
```diff
@@ -1,14 +1,16 @@
...
-    variant="nocolor"
-    class="mb-3 mr-3 font-weight-normal bg-transparent"
-    data-qa="browse chip"
-  >
-    {{ title }}
-  </b-badge>
+  <li>
+    <b-badge
+      v-if="title"
+      :to="localePath({ name: 'entity-type-all', params: { type: type, pathMatch: path }})"
```
rwd: While I appreciate this was pre-existing code, the entity route does not belong in this component which should be reusable in different contexts. Let's fix that on this branch.

Instead of accepting `type` and `path` props, accept a `linkTo` prop which could be either an object or string, and have the parent view supply the full entity route object.

components/browse/BrowseChip.vue
```diff
@@ -1,30 +1,31 @@
...
-    variant="nocolor"
-    class="mb-3 mr-3 font-weight-normal bg-transparent"
-    data-qa="browse chip"
-  >
-    {{ title }}
-  </b-badge>
+  <li>
+    <b-badge
+      v-if="title"
+      :to="localePath({
```
rwd: This is still coupled to the entity-type-all route.

The full path needs to be passed into the component so that here we need only:
```html
:to="linkTo"
```

PR #209, opened 2019-09-09
Preloader styling

lbiedinger:


nuxt.config.js
```diff
@@ -31,7 +31,11 @@ module.exports = {
   /*
   ** Customize the progress-bar color
   */
-  loading: { color: '#fff' },
+  loading: {
+    css: false,
+    duration: 750,
```
rwd: As discussed in person, let's increase this to 2500 ms so that on slow-loading pages it doesn't seem so frantic and distracting.

PR #210, opened 2019-09-09
Create facet dropdown

monkemedia:


components/search/FacetDropdowns.vue
```diff
@@ -0,0 +1,26 @@
...
+      },
+
+      selectedFacets: {
+        type: [Array, Object],
+        required: false,
+        default: () => []
+      }
+    }
+  };
+</script>
```
mirjamverloop: Is this file/component in use?

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,253 @@
...
+          {{ $t(`facets.${facet.name}.options.${option}`) }}
+        </b-form-radio>
+
+        <b-form-checkbox
+          v-else
+          :id="`${option.label}_checkbox`"
+          v-model="preSelected"
+          :value="option.label"
+          :data-qa="`${option.label} checkbox`"
+          :class="{ 'is-selected' : selectedFacet.some(s => s === option.label) }"
```
mirjamverloop: I think you can use the Bootstrap class "font-weight-bold" directly and get rid of the is-selected class here and in the style below.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,253 @@
...
+          :value="option"
+          :data-qa="`${option} ${RADIO}`"
+          @input="applySelection"
+        >
+          {{ $t(`facets.${facet.name}.options.${option}`) }}
+        </b-form-radio>
+
+        <b-form-checkbox
+          v-else
+          :id="`${option.label}_checkbox`"
```
mirjamverloop: This can result in ids that don't match the requirements of a valid id. A valid id can't contain any whitespace. And, for example, the id of the United Kingdom is "United Kingdom_checkbox".

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,253 @@
...
+      }
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import "./assets/scss/variables.scss";
+
+  .dropdown {
+    width: 100%;
```
mirjamverloop: Alternative: you can use a Bootstrap utility class, "w-100".

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,253 @@
...
+      min-width: 280px;
+    }
+
+    .custom-control  {
+      margin-bottom: 4px;
+    }
+  }
+
+  .options-container {
+    overflow: auto;
```
mirjamverloop: Alternative: you can use a Bootstrap utility class, "overflow-auto".

components/search/FacetDropdown.vue
monkemedia: I can do, but since I have already added styling to the dropdown, I think it's better to not clog up the HTML too much with helper classes. If there was no `.dropdown` class then I would agree. Thoughts?

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,254 @@
...
+    :data-type="facetType"
+    data-qa="search facet"
+  >
+    <template v-slot:button-content>
+      <span :data-qa="`${facetName} dropdown button`">
+        {{ facetName }}
+      </span>
+    </template>
+
+    <b-dropdown-form class="options-container">
```
rwd: Let's change the logic within this element to:
```html
      <template
        v-if="isRadio"
      >
        <b-form-radio
          v-for="(option, index) in sortOptions"
          :key="index"
          ...
        >
          ...
        </b-form-radio>
      </template>
      <template
        v-else
      >
        <b-form-checkbox
          v-for="(option, index) in sortOptions"
          :key="index"
          ...
        >
          ...
        </b-form-checkbox>
      </template>
```
The advantages of this are:
1. The superfluous `<div>` element is avoided
2. The if/else condition is only tested once for each use of the component instead of once for each option

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,202 @@
...
+      <div
+        v-for="(option, index) in sortOptions"
+        :key="index"
+      >
+        <b-form-radio
+          v-if="facetType === 'radio'"
+          :id="`${option}_radio`"
```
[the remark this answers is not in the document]
rwd: Let's avoid modifying code to suit tests, and instead modify the tests thus:
```diff
diff --git a/tests/features/search/faceting.feature b/tests/features/search/faceting.feature
index d933477..ffe9247 100644
--- a/tests/features/search/faceting.feature
+++ b/tests/features/search/faceting.feature
@@ -21,7 +21,7 @@ Feature: Search faceting
 
     When I visit the `search page`
     And I click the `Type of media dropdown button`
-    And I check the "IMAGE" checkbox
+    And I check the "IMAGE" "TYPE" checkbox
     And I click the `Type of media apply button`
     And I wait 2 seconds
     Then I should be on `/search?page=1&qf=TYPE%3A%22IMAGE%22&query=&view=grid`
diff --git a/tests/features/support/step-definitions.js b/tests/features/support/step-definitions.js
index 6393e3c..af5c493 100644
--- a/tests/features/support/step-definitions.js
+++ b/tests/features/support/step-definitions.js
@@ -53,8 +53,8 @@ defineStep('I activate/click (the/a/an)( ){target}', (qa) =>
 defineStep('I activate/click (on )(the/a/an)( ){target} in/on a/an/the {target}', (qa, parentQa) =>
   i.clickOnTheTarget([qa, parentQa]));
 
-defineStep('I check/click the {string} checkbox', (inputValue) =>
-  i.checkTheCheckbox(inputValue));
+defineStep('I check/click the {string} {string} checkbox', (inputValue, inputName) =>
+  i.checkTheCheckbox(inputName, inputValue));
 
 defineStep('I check/click the {string} radio', (inputValue) =>
   i.checkTheRadio(inputValue));
diff --git a/tests/features/support/step-runners.js b/tests/features/support/step-runners.js
index ee72a6c..bb6ca63 100644
--- a/tests/features/support/step-runners.js
+++ b/tests/features/support/step-runners.js
@@ -51,9 +51,13 @@ module.exports = {
 
     await client.initAccessibility().assert.accessibility('html', axeOptions);
   },
-  async checkTheCheckbox(inputValue) {
-    const selector = `label[for="${inputValue}_checkbox"]`;
-    await client.click(selector);
+  async checkTheCheckbox(inputName, inputValue) {
+    const checkboxSelector = `input[type="checkbox"][name="${inputName}"][value="${inputValue}"]`;
+    await client.getAttribute(checkboxSelector, 'id', (result) => {
+      const checkboxId = result.value;
+      const labelSelector = `label[for="${checkboxId}"]`;
+      client.click(labelSelector);
+    });
   },
   async checkTheRadio(inputValue) {
     await client.click(`label[for="${inputValue}_radio"]`);
```
Do the same for radios, and you can then remove the custom id attributes from the inputs.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,248 @@
...
+      },
+
+      idGenerator(label, type) {
+        return label.replace(/\s+/g, '_').toLowerCase() + '_' + type;
+      },
+
+      applySelection() {
+        if (this.isRadio) {
+          this.$emit('updated', this.THEME, this.radioSelected);
+          this.$refs.dropdown.hide(true);
```
mirjamverloop: This `this.$refs.dropdown.hide(true);` is the same in the if as well as in the else. I guess, it can go outside of the if/else statement and have it just once.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+        this.preSelected = [];
+      },
+
+      resetRadioSelection() {
+        this.radioSelected = '';
+      },
+
+      applySelection() {
+        if (this.isRadio) {
+          this.$emit('updated', this.THEME, this.radioSelected);
```
rwd: Can't this just use `this.facet.name`?

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+    </li>
+  </b-dropdown>
+</template>
+
+<script>
+  import isEqual from 'lodash/isEqual';
+
+  export default {
+    props: {
+      facet: {
```
rwd: Let's replace this prop with two, one for `name` (a string) and one for `fields` (an array).

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+        type: Object,
+        required: true
+      },
+
+      selectedFacet: {
```
rwd: Rename this to `selected`.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+        type: [Array, String],
+        required: false,
+        default: () => []
+      },
+
+      facetType: {
```
rwd: Rename this to `type`.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+        CHECKBOX: 'checkbox',
+        THEME: 'THEME',
+        preSelected: [],
+        radioSelected: null
+      };
+    },
+
+    computed: {
+      sortOptions() {
+        let newArray = [].concat(this.facet.fields);
```
rwd: This seems redundant as it's never modified.

Just use `this.facet.fields` below where `newArray` is used.

components/search/FacetDropdown.vue
monkemedia: Already have a computed property called selected

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+
+      facetType: {
+        type: String,
+        required: true
+      }
+    },
+
+    data() {
+      return {
+        RADIO: 'radio',
```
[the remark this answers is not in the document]
monkemedia: I like to have one place where you can easily change the string.. ie quickly change it to a 'Radio' instead

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,242 @@
...
+      activateCheckboxResetButton() {
+        return this.preSelected.length > 0;
+      },
+
+      disableRadioResetButton() {
+        return !this.radioSelected;
+      },
+
+      disableApplyButton() {
+        return isEqual(this.preSelected, this.selectedFacet);
```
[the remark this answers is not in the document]
rwd: Stick to using lodash.

components/search/FacetDropdown.vue
rwd: Where?

components/search/FacetDropdown.vue
rwd: This still needs to be renamed.

components/search/FacetDropdown.vue
```diff
@@ -0,0 +1,245 @@
...
+
+      activateCheckboxResetButton() {
+        return this.preSelected.length > 0;
+      },
+
+      disableRadioResetButton() {
+        return !this.radioSelected;
+      },
+
+      disableApplyButton() {
```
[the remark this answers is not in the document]
monkemedia: This is what I had before, but there is a case where this logic fails, and we have to do a more thorough comparison. I think you if click 'netherlands' and select apply. Go back to the facet, untick the Netherlands, and select a different country, the Apply button becomes disabled again.

PR #211, opened 2019-09-09
Entities as search

rwd:
* Creates a new component to gather together all search related components, to be used in both the core search page and the entity pages, and anywhere else record search is wanted
  * When search results need to be updated, e.g. due to pagination or faceting, the component is responsible for querying the Record API, not the page
  * Content tier toggle is optional and disabled on the entity pages
* Entity pages do not show their search query (i.e. the entity URI) in the search box in the header, nor include it in pagination or faceting route updates
* Sets entity page HTML title from the English prefLabel
* Removes isLoading properties which never come into effect
* Moves API error handling into the plugins, using http-errors package, and preserving the status code returned by the APIs
* Makes the search plugin the SSOT for thematic collections and default facets
* Uses Vuex store for search state storage
* Combines grid and list search results components into one


components/PageHeader.vue
```diff
@@ -67,6 +65,7 @@
       submitSearchForm() {
         const newSearchQuery = { ...this.searchQuery, ...{ query: this.query, page: 1 } };
         this.$router.push(this.localePath({ name: 'search', query: newSearchQuery }));
+        this.$root.$emit('submit:searchForm', this.query);
```
monkemedia: I wonder if we should be making the most of using Stores for this kind of functionality?

tests/unit/components/search/ViewToggles.spec.js
```diff
@@ -28,37 +28,41 @@ const factory = () => mount(ViewToggles, {
...
-  });
-
-  it('links to search with view parameter set', () => {
-    const wrapper = factory();
+        const viewToggle = wrapper.find(`[data-qa="search ${view} view toggle"]`);
+        viewToggle.exists().should.eq(true);
+      });
 
-    const viewToggleLink = wrapper.find('[data-qa="search list view toggle"] a');
+      // TODO: why does this fail? href in tests is just e.g. "?view=list" without
```
monkemedia: Took me all morning to resolve. However, I have found the solution. You just need to pass the `path: '/search'` in the `linkGenRoute` property.  So you would end up with something like this:

```
linkGenRoute: {
    name: 'search',
    path: '/search'
}
```

components/search/SearchResultsGrid.vue
```diff
@@ -24,11 +24,24 @@
       ContentCard
     },
     props: {
```
lbiedinger: perhaps we'll want to pass as a prop the maximum number of columns.
Then we could pass this in from the search/entity page and set 'card-deck-3-cols/card-deck-4-cols' accordingly.

components/search/SearchResults.vue
```diff
@@ -0,0 +1,417 @@
...
+        >
+          <b-row>
+            <b-col>
+              <PaginationNav
+                v-if="showPagination"
+                v-model="currentPage"
+                :total-results="currentTotalResults"
+                :per-page="perPage"
+                :link-gen="paginationLink"
+                @changed="changePage"
```
lbiedinger: for consistency consider renaming the method to 'selectPage'

components/search/SearchResults.vue
```diff
@@ -0,0 +1,417 @@
...
+                :message="$t('resultsLimitWarning')"
+              />
+            </b-col>
+          </b-row>
+          <b-row>
+            <b-col>
+              <TierToggler
+                v-if="showContentTierToggle"
+                :active-state="contentTierActiveState"
+                @changed="changeContentTierToggle"
```
lbiedinger: for consistency consider renaming the method to 'selectContentTier'

pages/entity/_type/_.vue
```diff
@@ -135,15 +134,20 @@
...
         entities.relatedEntities(params.type, params.pathMatch, {
           wskey: env.EUROPEANA_API_KEY,
           entityKey: env.EUROPEANA_ENTITY_API_KEY
         }),
+        // TODO: DRY up (shared with search/index)
         search({
           page: currentPage,
-          query: `"${entities.getEntityUri(params.type, params.pathMatch)}"`,
+          qf: query.qf,
+          query: entityQuery,
```
lbiedinger: if a query is present in the url params we'll want to add that to the entityQuery.
Should they be 'AND'ed or do we need to use a fq like we do in the current portal for multiple search terms?

components/PageHeader.vue
rwd: Very good point, and I have done so in b5eb3144398802367de4e8f9dc82610bd9bae4f1.

tests/unit/components/search/ViewToggles.spec.js
rwd: I'm reluctant to do this as it's not needed by the application functionality itself.

rwd:
@lbiedinger re:
> the 'components/search/SearchResults.vue' component does much more than show search results. In it's current implementation it's responsible for the entire 'search' component including various ways to interact with the search results. Could we rename it to just ` components/search/Search.vue`

I agree it's not a good name for the component, but Search.vue is [not acceptable](https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential).

How about SearchInterface.vue?

pages/entity/_type/_.vue
```diff
@@ -17,32 +17,23 @@
         md="9"
       >
         <EntityDetails
-          :title="title"
-          :depiction="depiction"
           :attribution="attribution"
+          :depiction="depiction"
           :description="description"
+          :title="title"
```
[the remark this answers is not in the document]
rwd: Yes, alphabetically on purpose. Those on SearchResults have been ordered so too. I don't think eslint has a relevant rule. We are already using [vue/attributes-order
](https://eslint.vuejs.org/rules/attributes-order.html).

plugins/europeana/entity.js
```diff
@@ -21,8 +22,15 @@ export function getEntity(type, id, params) {
...
-      throw new Error(message);
+      let statusCode = 500;
+      let message = error.message;
+
+      if (error.response) {
+        statusCode = error.response.status;
+        message = error.response.data.error;
+      }
+
+      throw httpError(statusCode, message);
```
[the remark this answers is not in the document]
rwd: Fair point. Created a utility plugin at plugins/europeana/utils.js

pages/entity/_type/_.vue
rwd: That's outside of the scope of this piece of work. It is to follow for https://europeana.atlassian.net/browse/EC-3710.

components/search/SearchResultsGrid.vue
rwd: @mirjamverloop what do you think about this suggestion?

components/search/SearchResults.vue
rwd: I have instead renamed selectFacet to changeFacet.

components/search/SearchResults.vue
rwd: I have instead renamed selectFacet to changeFacet.

components/search/SearchResultsList.vue
```diff
@@ -22,10 +22,23 @@
...
+        results: this.value
+      };
+    },
+    watch: {
+      value: {
+        immediate: true,
+        handler(val) {
+          this.results = val;
+        }
+      }
```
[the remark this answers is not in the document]
rwd: Agreed. These have been combined into one component in bfcedb16eab7856c6d4de639f4d4cae70efcf75c.

components/search/SearchResultsGrid.vue
rwd: After discussion, we will not do this here, but come back to it after this and #210 are into master.

lang/en.js
```diff
@@ -68,9 +69,10 @@ export default {
   "showMore": "Show more",
   "record": "Record",
   "messages": {
+    "externalContentError": "The external content failed to load",
     "notFound": "Not Found",
     "noTitle": "No title provided for \"{record}\"",
```
lbiedinger: this "no title provided for..." string is no longer used.
If we don't want to show it anymore we can get rid of the string.
Otherwise it'll need to be added back to the search result.

nuxt.config.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
   /*
   ** Plugins to load before mounting the App
   */
-  plugins: ['~/plugins/vue-filters', '~/plugins/global.js'],
+  plugins: ['~/plugins/vue-filters'],
```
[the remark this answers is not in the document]
rwd: Just that I would prefer us to always explicitly import the components needed as we generally do.

lang/en.js
rwd: Removed from the lang files in 5263e4a.

PR #212, opened 2019-09-10
Remove Vue warnings on badges, update stylelint command and packages

mirjamverloop:
+ Updated tests


components/browse/BrowseChip.vue
```diff
@@ -17,7 +17,7 @@
   export default {
     props: {
       linkTo: {
-        type: Object,
+        type: String,
```
rwd: As router links can accept either an object or a string, let's permit either type here.

PR #215, opened 2019-09-17
Header menu

monkemedia:


nuxt.config.js
```diff
@@ -235,6 +235,12 @@ module.exports = {
         alwaysRedirect: false,
         // If no locale for the browsers locale is a match, use this one as a fallback
         fallbackLocale: 'en'
+      },
+      vuex: {
+        // Module namespace
+        moduleName: 'i18n',
+        syncLocale: true,
+        syncRouteParams: true
```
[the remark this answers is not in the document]
monkemedia: This should be done in a new ticket

components/PageNavigation.vue
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  .navbar-light {
+    .navbar-nav {
+      .nav-link {
+        color: $darkgrey;
+      }
+    }
+  }
+</style>
```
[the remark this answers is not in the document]
monkemedia: I have followed the same name convention as PageFooter and PageHeader

store/navigation.js
```diff
@@ -0,0 +1,38 @@
...
+    function isoLookUp(code) {
+      const locales = i18n.locales;
+      return locales.find(locale => locale.code === code)['iso'];
+    }
+
+    const setLocale = i18n.locale;
+    const data = await contentfulClient.getEntries({
+      'locale': isoLookUp(setLocale),
+      'content_type': 'linkGroup',
+      'fields.identifier': 'pageMainNavigation'
```
lbiedinger: what do you think about using`mainNavigiation` and updating the identifier in contentful?

I think page is redundant and adds more confusion with the issue Mirjam Raised above re pagination & navigation components

components/PageNavigation.vue
```diff
@@ -0,0 +1,82 @@
...
+      }
+    },
+
+    mounted() {
+      this.getNavigationData();
+    },
+
+    methods: {
+      async getNavigationData() {
+        return this.$store.dispatch('navigation/init');
```
lbiedinger: add this, or rather `    await dispatch('navigation/init');` to store/index.js the same as the footer, that way it avoids the loading in only happening after the page already rendered. Because now it jumps the whole page down on a hard referesh. We'll still need this getNavigationData as otherwise it won't pick up on language switchting.

components/PageNavigation.vue
```diff
@@ -0,0 +1,82 @@
...
+  >
+    <b-container class="p-0">
+      <b-navbar-nav class="px-2">
+        <li
+          v-for="nav in navigation"
+          :key="nav.text"
+          class="nav-item"
+        >
+          <SmartLink
+            :destination="localePath({name: 'slug', params: { pathMatch: nav.url.replace(/^\//, '') }})"
```
lbiedinger: this should just take the url. The localepath call and the '/' replacement should only happen in the smart link component for the case of the destination not being a url.
Now this will break where a top nav link goes to an external site (see the updated 'blogs' one). Also we'll be duplicating this logic for other smart links if we don't move it into the component.

components/PageNavigation.vue
monkemedia: The reason for putting this in a method was due to language selector. If we use `nuxtServerInit` method, the language doesn't update when the user changes the language. Hope that makes sense.

components/PageNavigation.vue
monkemedia: How do we approach this logic? Do we pass a property to let it know we need to add a locale path?

lbiedinger: just always use a locale path. I can't think of a scenario where we'd want to link to something not in the selected locale.

components/PageNavigation.vue
lbiedinger: yes, as I was saying we'll want it in both places.

lbiedinger: we could attempt to skip the mounted call when the local hasn't changed, but I think this is probably overkill at this point.

monkemedia: Apologies.. misread

monkemedia: Dont think mounted is actually need if we are going to use NuxtServerInit

nuxt.config.js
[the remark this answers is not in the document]
rwd: Agreed.

nuxt.config.js
```diff
@@ -235,6 +235,12 @@ module.exports = {
         alwaysRedirect: false,
         // If no locale for the browsers locale is a match, use this one as a fallback
         fallbackLocale: 'en'
+      },
+      vuex: {
```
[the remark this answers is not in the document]
monkemedia: This is so VUEX is able to access any changes in the locale. If you comment it out and change the language to dutch (language selector), 'Collections' link in the menu will not change.

store/link-group.js
```diff
@@ -0,0 +1,46 @@
+import contentfulClient from '../plugins/contentful.js';
+
+export const state = () => ({
+  links: {}
+});
+
+export const mutations = {
+  setLinks(state, data) {
+    for (const d in data) {
```
rwd: As the action triggers the commit for each group, change this to:
```js
  setLinks(state, data) {
    state.links[data.identifier] = data.links;
  }
```
(And see the accompanying change in the action, below.)

components/PageNavigation.vue
```diff
@@ -0,0 +1,73 @@
...
+</script>
+
+<style lang="scss" scoped>
+  @import '../assets/scss/variables.scss';
+
+  .navigation-container {
+    background: $bodygrey;
+    padding-top: 0;
+    padding-bottom: 0;
+  }
```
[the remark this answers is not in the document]
mirjamverloop: Can you take a look at this suggestion, Richard?

components/PageNavigation.vue
```diff
@@ -0,0 +1,73 @@
...
+
+  .nav-item:not(:last-child) {
+    margin-right: 15px;
```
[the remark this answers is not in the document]
mirjamverloop: Can you take a look at this suggestion, Richard?

components/PageNavigation.vue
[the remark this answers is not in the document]
mirjamverloop: And this one is still open as well.

components/PageNavigation.vue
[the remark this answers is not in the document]
monkemedia: Done

components/generic/SmartLink.vue
```diff
@@ -1,14 +1,15 @@
 <template>
   <b-link
     v-if="useRouterLink"
-    :to="destination"
+    :to="path"
+    :exact-active-class="exactActiveClass"
```
mirjamverloop: Where or when is this used? Isn't it always null?

pages/entity/_type/_.vue
```diff
@@ -11,7 +11,7 @@
     v-else
     data-qa="entity page"
   >
-    <b-row class="flex-md-row">
+    <b-row class="flex-md-row pt-3">
```
mirjamverloop: This is not sufficient, having a padding-top on entity or search pages. Eg:
/great-women-in-eu-history

There should be a margin between the header menu and the underlying content on every page where there is no hero banner.

pages/search/index.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <b-container>
-    <b-row>
+    <b-row class="pt-3">
```
mirjamverloop: See previous comment.

components/PageNavigation.vue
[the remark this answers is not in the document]
mirjamverloop: Thanks!

components/generic/SmartLink.vue
monkemedia: This is used for active state in the header navigation. If you look at Pagenavigation.vue, you can see we are passing `font-weight-bold` to it

pages/entity/_type/_.vue
monkemedia: I was going to target the main navigation, but as you mentioned we would have to stop any margins on the homepage. Sadly there is no `homepage` class or ID I can target to achieve this. Can you please let me know your thoughts?

components/PageNavigation.vue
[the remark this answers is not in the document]
mirjamverloop: No, hang on, this is not right. Because you are using <b-navbar> there is a navbar class already. You can use it directly, no need to add it as a class. It is there now twice.

components/generic/SmartLink.vue
mirjamverloop: I am wondering if it is necessary, as nuxt links get an active/exact active class already. You can use that for styling. Eg, if you go to /mirjams-test, and inspect that link in the navigation, you can see the class "nuxt-link-active".

pages/entity/_type/_.vue
mirjamverloop: And a hero banner can be added to any browse page. What I would like to suggest for now is give the main navigation a mb-3 as a default. And add these lines to the hero.scss:

````.navbar + div .jumbotron {
      margin-top: -1rem;
}
`````
I think that should to the trick.

PR #219, opened 2019-09-19
9 results on entity pages

rwd:


tests/features/support/step-definitions.js
```diff
@@ -20,6 +20,9 @@ defineStep('I browse/open/visit (a/an/the)( ){target}', (pageName) =>
 defineStep('I find/identify/see/spot (a/an/the)( ){target}', (qa) =>
   i.seeATarget(qa));
 
+defineStep('I find/identify/see/spot (a/an/the)( ){target} {int} time(s)', (qa, count) =>
+  i.seeATarget(qa, { count }));
+
```
mirjamverloop: This is kind of similar to some other step, this one: 

`defineStep('I should have/see/see/spot {int} {target}(s)', (count, qa) =>
  i.countTarget(count, qa));
`
Does it make sense to combine them, so we have just one step that counts?

rwd: You are quite right. I have adjusted to use the existing one.

PR #222, opened 2019-09-19
Sections on entity page

mirjamverloop:
- Introduced a BrowseSection component that collects all sections (for now: content cards and rich text)
- BrowseSection is currently used on (Browse) pages and Entity pages


pages/entity/_type/_.vue
```diff
@@ -37,6 +37,10 @@
           :show-content-tier-toggle="false"
           :total-results="search.totalResults"
         />
+        <BrowseSection
```
lbiedinger: move this down into it's own row.

mirjamverloop: It has its own row now and I have fixed some basic styling.

components/browse/BrowseSection.vue
```diff
@@ -0,0 +1,44 @@
...
+  import ContentCardSection from './ContentCardSection';
+
+  export default {
+    components: {
+      RichText,
+      ContentCardSection
+    },
+
+    props: {
+      sections: {
```
rwd: Given that it accepts a collection of sections, this component should be named `BrowseSections.vue`, i.e. plural.

PR #223, opened 2019-09-20
Entity section retrieval

lbiedinger:


pages/entity/_type/_.vue
```diff
@@ -141,6 +144,18 @@
...
       }
 
+      let contentfulClient;
+      if (query.mode === 'preview' && process.env['CTF_CPA_ACCESS_TOKEN']) {
+        contentfulClient = createClient(query.mode);
+      } else {
+        contentfulClient = createClient();
+      }
+      const setLocale = app.i18n.locale;
+      const isoLookUp = (code) => {
```
rwd: As this duplicates code from pages/index.vue, I suggest creating a new plugin i18n.js which would accept the value of `app.i18n` as its parameter.

pages/entity/_type/_.vue
```diff
@@ -141,6 +144,18 @@
         }));
       }
 
+      let contentfulClient;
+      if (query.mode === 'preview' && process.env['CTF_CPA_ACCESS_TOKEN']) {
```
rwd: Let's not check this here, but instead replace the whole if/else, both here and in pages/index.vue, with just:
```js
contentfulClient = createClient(query.mode);
```

If anywhere should be responsible for testing for the presence of the preview API token, it's the plugin, but we could even omit it from there.

PR #227, opened 2019-09-23
Entity page: 3 cards per row, Search page: 4

mirjamverloop:


components/search/SearchInterface.vue
```diff
@@ -187,6 +188,10 @@
         type: Number,
         default: 24
       },
+      perRow: {
+        type: Number,
+        default: 3
```
rwd: As the main search shows 4, let's default this to 4.

PR #228, opened 2019-09-23
Click outside of facet dropdown applies them

lbiedinger:


components/search/SearchInterface.vue
```diff
@@ -322,6 +322,12 @@
         this.$store.commit('search/setView', this.$route.query.view);
       }
     },
+
+    mounted() {
+      this.$root.$on('bv::dropdown::hidden', (dropdown) => {
```
rwd: Instead of this, can't we use `@hidden` on the `b-dropdown` template element in FormDropdown.vue?

lbiedinger: We can Indeed, and that's updated.

components/search/FacetDropdown.vue
```diff
@@ -5,6 +5,7 @@
     class="mr-2"
     :data-type="type"
     data-qa="search facet"
+    @hidden="applySelection()"
```
rwd: The parentheses are not needed here.

PR #229, opened 2019-09-23
Rich text sections on CMS pages

monkemedia:


pages/index.vue
```diff
@@ -13,23 +13,33 @@
     />
     <section class="container">
       <div class="mt-3 w-100">
-        <ContentCardSection
-          v-for="section in page.hasPart"
-          :key="section.sys.id"
-          :section="section"
-        />
+        <template v-if="isRichText">
```
rwd: A page may contain multiple sections of different types. This therefore needs to be beneath the `v-for` on `page.hasPart` and test the type of each section.

pages/index.vue
```diff
@@ -13,23 +13,33 @@
...
+          <RichText
+            :headline="page.hasPart[0].fields.headline"
+            :text="page.hasPart[0].fields.text"
+          />
+        </template>
+        <template v-else>
```
rwd: We should stop assuming that sections will be of a particular type, by here using `v-if-else` (type is cardGroup), and having no `v-else`. The content would not be rendered for any other sections, but the rest of the page would not fail to load.

pages/index.vue
monkemedia: So you want the Rich text component in the contentCardSection component?

rwd: No.

We need something like:
```html
<template
  v-for="section in page.hasPart"
  :key="section.sys.id"
>
  <RichText
    v-if="isRichText"
    :headline="section.fields.headline"
    :text="section.fields.text"
  />
  <ContentCardSection
    v-else-if="isCardGroup"
    :section="section"
  />
</template>
```

pages/index.vue
rwd: See code sample at https://github.com/europeana/incubator-portal-vue-nuxt/pull/229/files#r327126030

pages/index.vue
monkemedia: Thank you Rich. I now follow

pages/index.vue
```diff
@@ -60,7 +72,9 @@
             return;
           }
           return {
-            page: response.items[0].fields
+            page: response.items[0].fields,
+            isRichText: response.items[0].fields.hasPart[0].sys.contentType.sys.id === 'richText',
```
rwd: These are always looking at the first hasPart but there may be multiple sections of different types.

`isRichText` and `isCardGroup` need to be methods taking one of the section objects as a parameter, then inspecting its `sys.contentType.sys.id`

PR #234, opened 2019-09-24
Created search bar pill component

monkemedia:


components/search/SearchBarPill.stories.js
```diff
@@ -0,0 +1,14 @@
+import { storiesOf } from '@storybook/vue';
+
+import SearchBarPill from './SearchBarPill.vue';
+
+storiesOf('Search page', module)
```
rwd: There are no aria labels on the links in Storybook. Why is that?

It would be good to see the default coming through on one of the pills, and a custom aria label on the other.

Each pill should be here as its own "story".

.storybook/config.js
```diff
@@ -8,7 +8,16 @@ import '../plugins/vue-filters'
 import '../assets/scss/style.scss';
 
 Vue.use(BootstrapVue);
-
+Vue.prototype.$t = function () {}
```
rwd: The entire code addition to this file needs to be linted.

.storybook/config.js
```diff
@@ -8,7 +8,16 @@ import '../plugins/vue-filters'
...
+Vue.component('NuxtLink', {
+  props:   ['to'],
+  methods: {
```
rwd: What is this method for?

components/search/SearchBarPill.vue
```diff
@@ -0,0 +1,65 @@
...
+
+    data() {
+      return {
+        limitCharacters: 20
+      };
+    },
+
+    computed: {
+      truncatedText() {
+        return this.text.length > this.limitCharacters ? this.text.slice(0, this.limitCharacters) + '...' : this.text;
```
rwd: Use the ellipsis character `…` instead of three period characters.

components/search/SearchBarPill.vue
```diff
@@ -0,0 +1,65 @@
...
+        type: String,
+        required: true
+      },
+      removeLinkTo: {
+        type: Object,
+        required: true
+      },
+      removeLinkLabel: {
+        type: String,
+        default() {
```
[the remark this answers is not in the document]
monkemedia: Sadly not. It will break Storybook

.storybook/config.js
monkemedia: This is so Nuxt Links work within Storybook otherwise it breaks.

.storybook/config.js
monkemedia: Removed, and now uses Vue-i18n. Seems to do the trick :)

components/search/SearchBarPill.vue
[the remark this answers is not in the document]
rwd: I do not like adapting our code to suit auxiliary tooling. If the more concise format is in fact valid from the Vue/Nuxt point of view, then could you see if you can get it working with Storybook.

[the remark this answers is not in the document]
rwd: Does the introduction of VueI18n to storybook help?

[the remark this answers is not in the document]
monkemedia: Sadly not. 

If we don't use a function, it breaks the application, not just in StoryBook.

Someone else also had the same issue https://stackoverflow.com/questions/53170638/vuejs-default-properties-set-from-vue-i18n

.storybook/config.js
monkemedia: apologies, miss read this comment. Method is so that it can work in conjunction with StoryBook actions. Some info about it here https://medium.com/js-dojo/a-guide-on-using-storybook-with-nuxt-js-1e0018ec51c9#3977

For some reason the action console isnt showing... I will investigate

monkemedia: StoryBook actions now work so you can see the method at work now :)

components/search/SearchBarPill.vue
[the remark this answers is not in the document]
rwd: Understood.

.storybook/config.js
rwd: Understood.

.storybook/addons.js
```diff
@@ -1,2 +1,3 @@
+import '@storybook/addon-actions';
 import '@storybook/addon-actions/register';
```
mirjamverloop: It seems unnecessary to import both @storybook/addon-actions and @storybook/addon-actions/register, but I might be wrong. Could you explain this?

monkemedia: Actions addon places a console area at the bottom of the page so the developer can see what objects/strings are being passed when you click on a button etc. As for the `/register` import, that was already there, so I didn't want to touch it

mirjamverloop: That makes sense, thanks!

PR #238, opened 2019-09-26
Use Vuex store for search & search within entities

rwd:
New:
* Searching from an entity page searches within that entity

Changed:
* Use Vuex for full search handling and data storage.

Fixed:
* Nuxt loader shows when paginating, faceting, etc
* Back button works after paginating, faceting, etc
* Using pagination links returns to top of page


components/search/SearchForm.vue
```diff
@@ -35,14 +35,24 @@
       };
     },
     computed: {
+      onSearchPage() {
```
lbiedinger: onSearchPage suggests this is only used on the /search page. What about onSearchablePage, to also acknowledge the entity pages?

rwd: Done.

components/search/SearchForm.vue
```diff
@@ -35,14 +35,24 @@
...
+          return this.onSearchablePage ? this.$store.state.search.query : '';
         },
         set(value) {
           this.inputQuery = value;
         }
       },
+      routePath() {
+        if (this.onSearchablePage) {
+          return this.$route.path;
+        } else {
```
monkemedia: Don't need an ELSE here as you have a return in the IF statement

PR #239, opened 2019-09-26
Warning on conflicting order

monkemedia:


components/search/SearchInterface.vue
```diff
@@ -116,9 +116,9 @@
 
 <script>
   import AlertMessage from '../../components/generic/AlertMessage';
+  import SearchResults from '../../components/search/SearchResults';
```
lbiedinger: To save future confusion, let's add an inline comment here: `...; // Sorted before InfoMessage to prevent Conflicting CSS sorting warning`

monkemedia: Done

PR #241, opened 2019-09-26
Disable language selector by default

rwd:
Env var feature toggle

### DIFF
--- a/tests/features/common/language-selector.feature
+++ b/tests/features/common/language-selector.feature
@@ -1,11 +1,17 @@
 Feature: Select language
 
-  Scenario: Change language
+Scenario: Change language
 
-    When I visit `/de/search?query=`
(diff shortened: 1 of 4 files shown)


tests/features/common/language-selector.feature
```diff
@@ -1,11 +1,17 @@
 Feature: Select language
 
-  Scenario: Change language
+Scenario: Change language
 
-    When I visit `/de/search?query=`
```
monkemedia: Should we comment this out instead, at least when put the language selector back in, we will have the same tests as before?

rwd: I did comment the feature test out (see below). Is that not what you mean?

monkemedia: My bad. sorry

PR #243, opened 2019-09-27
Entity page: filter by theme if set

rwd:


plugins/europeana/entity.js
```diff
@@ -118,14 +118,23 @@ export function getEntitySlug(entity) {
  * @return {Object} related entities
  */
 export function relatedEntities(type, id, params) {
+  let apiParams = {
```
lbiedinger: Changing the relatedEntities logic shouldn't be in scope for this.
The related entities logic is known to not be very useful and the mechanism is likely going to be replaced, so we shouldn't attempt to add additional logic that takes into account the theme param.

plugins/contentful.js
```diff
@@ -12,13 +12,9 @@ const previewConfig = {
   host: 'preview.contentful.com'
 };
 
-export function createClient(mode) {
+function createClient(mode) {
```
lbiedinger: this can be further simplified:
```
export default function createClient(mode) {
  ...
```
and remove the line at the  end of the file.

store/entity.js
```diff
@@ -1,9 +1,39 @@
...
+  async init({ commit }) {
+    // TODO: account for potential pagination if > 1,000 entries
+    await contentfulClient.getEntries({
+      'content_type': 'entityPage',
+      'fields.genre[exists]': 'true',
+      'include': 0,
+      'limit': 1000
+    })
+      .then((response) => {
+        // console.log(`entity page response: ${JSON.stringify(response)}`);
```
lbiedinger: remove comment

plugins/europeana/entity.js
rwd: Fair comment. I have removed the logic for applying theme for related entities. I have however kept the change to use `entityQuery` to keep consistent with that used for the entity search results if it has no theme set. From my testing a few entities, this does not affect the related entitities returned.

PR #244, opened 2019-09-27
Show the entity name as pill

monkemedia:


components/search/SearchForm.vue
```diff
@@ -3,32 +3,54 @@
...
+      <template
+        v-if="pillLabel"
+        v-slot:prepend
       >
-    </b-button>
+        <SearchBarPill
+          :text="pillLabel"
+          :remove-link-label="$t('removeEntityFilter', { entityLabel: pillLabel })"
+          :remove-link-to="{
+            path: localePath({
```
lbiedinger: instead of having multiline code in the template here, should this be a computed property or method?

rwd: Agreed, let's do this already.

components/search/SearchForm.vue
```diff
@@ -3,32 +3,54 @@
...
-        :alt="$t('search')"
+    <b-input-group>
+      <template
+        v-if="pillLabel"
+        v-slot:prepend
       >
-    </b-button>
+        <SearchBarPill
+          :text="pillLabel"
+          :remove-link-label="$t('removeEntityFilter', { entityLabel: pillLabel })"
```
rwd: This should be supplied from the store too. The search form component should not need to know anything about entities.

rwd: On second thoughts, all that should be needed is to rename the translation and placeholder in lang/en.js as its content is entity-agnostic anyway.

So change it in lang/en.js to: `"removeFilter": "Remove {filterLabel} filter"`

PR #250, opened 2019-09-30
Move locales into JSON file

rwd:


nuxt.config.js
```diff
@@ -62,158 +62,7 @@ module.exports = {
...
-          iso: 'sl-SI'
-        },
-        {
-          name: 'Svenska',
-          code: 'sv',
-          file: 'sv.js',
-          iso: 'sv-SE'
-        }
-      ],
+      locales: require('./plugins/i18n/locales.json'),
```
monkemedia: would it be neater if you created a locales const at the top of the page?

rwd: Indeed, and done.

nuxt.config.js
```diff
@@ -47,7 +48,7 @@ module.exports = {
   /*
   ** Plugins to load before mounting the App
   */
-  plugins: ['~/plugins/vue-filters', '~/plugins/i18n'],
+  plugins: ['~/plugins/vue-filters', '~/plugins/i18n.js'],
```
lbiedinger: Is this change needed?
Shouldn't it be kept consistent between the two plugins loaded here?

rwd: It is needed because of the existence of a plugins/i18n directory. Without it there is a warning about ambiguity of what is being required.

PR #258, opened 2019-10-01
External links new window

monkemedia:


assets/img/icons/link-external-white.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lbiedinger: why not use the link-external.svg we already have and just change `fill="#1a1a1a"` to `fill="#FFFFFF"?

components/generic/SmartLink.vue
```diff
@@ -45,7 +54,17 @@
           });
         }
         return this.destination;
+      },
+      isExternalLink() {
+        return parseDomain(this.destination);
```
lbiedinger: this won't be enough. We'll need to check that the domain is included in a list. Either it needs to match the Domain of the currently loaded page, or more likely we'll want to be able to set this via the .env file, something like:
`INTERTNAL_DOMAINS="europeana.eu, eanadev.org"`
Then we can set it for the various deployements.

assets/img/icons/link-external-white.svg
monkemedia: Following on from your comment above:

The problem with having SVGs as  tags is that we are very limited with what we can do with them. For example, if we wanted to change the color, or even make it larger, we would have to create a new SVG and change it in the source code. That's why I believe it would be better to be using either inline SVG's or use something like Fontawesome for all our icons.

components/generic/SmartLink.vue
monkemedia: The only issue I would see with a whitelist is that content editors wouldn't be able to update it if they were to introduce a new external link

rwd: As discussed elsewhere, it should just use the current domain.

components/generic/SmartLink.vue
```diff
@@ -9,13 +9,16 @@
   <b-link
     v-else
     :href="path"
-    :class="linkClass"
+    :target="isExternalLink ? '_blank' : '_self'"
+    :class="[{ 'is-external-link' : isExternalLink }, linkClass]"
```
monkemedia: Kept this here for future development

assets/scss/style.scss
```diff
@@ -7,3 +7,5 @@
 @import "form.scss";
 @import "hero.scss";
 @import "licenses.scss";
+@import "icons.scss";
+@import "external-link-icon.scss"
```
mirjamverloop: Order these alphabetically.

components/generic/SmartLink.vue
```diff
@@ -45,6 +48,14 @@
...
         }
         return this.destination;
+      },
+      isExternalLink() {
+        if (!process.browser) return;
+
+        const destDomain = parseDomain(this.destination);
+        const currentDomain = parseDomain(window.location.href);
+
+        return destDomain && currentDomain && destDomain.domain + destDomain.tld !== currentDomain.domain + currentDomain.tld;
```
mirjamverloop: This is hard to read, is there another way to write it? Maybe using brackets would help?

assets/scss/external-link-icon.scss
```diff
@@ -0,0 +1,20 @@
...
+  &.card-link {
+    p:last-child {
+      &:after {
+        @include external-icon;
+        margin-left: 4px;
+        vertical-align: middle;
+      }
+    }
+  }
+}
```
mirjamverloop: This is duplication of code, as there is already the icons.scss. I am more inclined to say: stick to that CSS and naming convention. So for example, add a `<span class="icon-external-link"></span>`. And I might prefer to have the icon after the main title on the cards.

monkemedia: To achieve this you would have to inject the span element with javascript, thats why I have taken the approach of using pseudo-classes here. The span is ideal for when we can manually put the icon next to a title (for example), but when it comes to dynamically generated content, using classes is probably the best way to go. 

For example:

Cards components - the icon will likely appear after the last paragraph
Footer - the icon will appear after the link text

Added classes to the SmartLink gives us a lot of flexibility on where we want the icon to appear.

package.json
```diff
@@ -53,12 +53,15 @@
     "express": "^4.16.4",
     "express-ipfilter": "^1.0.1",
     "http-auth": "^3.2.3",
+    "install": "^0.13.0",
```
rwd: What's this?

package.json
```diff
@@ -53,12 +53,15 @@
...
     "lodash": "^4.17.15",
     "morgan": "^1.9.1",
     "nightwatch-accessibility": "^1.8.0",
+    "npm": "^6.11.3",
```
rwd: What's this?

assets/scss/icons.scss
```diff
@@ -0,0 +1,26 @@
+@font-face {
+  font-family: 'icomoon';
+  src:  url('../fonts/icomoon.woff?plxcjd') format('woff');
+  font-weight: normal;
+  font-style: normal;
+  font-display: block;
+}
+
+[class^="icon-"], [class*=" icon-"] {
+  /* use !important to prevent issues with browser extensions that change fonts */
```
rwd: What's the purpose of this?

components/generic/SmartLink.vue
```diff
@@ -45,6 +54,14 @@
           });
         }
         return this.destination;
+      },
+      isExternalLink() {
```
rwd: Could I propose some changes:
```js
      isExternalLink() {
        if (this.useRouterLink) return false;

        if (!process.browser) return;

        const destDomain = parseDomain(this.destination);
        const dest = [destDomain.domain, destDomain.tld].join('.');
        const currentDomain = parseDomain(window.location.href);
        const current = (currentDomain === null ? window.location.hostname : [currentDomain.domain, currentDomain.tld].join('.'));

        return dest !== current;
      }
```
... to:
1. first see if it's a router link as they are always internal
2. join domain and TLD with `.` before comparison as that's how they are in FQDNs
3. fallback to the full hostname from window.location if parse domain could not identify it, which makes it work with localhost

package.json
monkemedia: This was attached when I installed `parse-domain`

package.json
monkemedia: This was attached when I installed `parse-domain`

assets/scss/icons.scss
monkemedia: This style sheet is given to you when you use icomoon. At the moment, we are not using classes, but at some point we will be, so this will need to be here. I can delete it for now if you wish.

package.json
monkemedia: I deleted it, and parse domain still works :)

package.json
monkemedia: I deleted it, and parse domain still works :)

components/generic/SmartLink.vue
monkemedia: I have made the changes here. Much much cleaner thank you. I left out `if (this.useRouterLink) return false;` as this logic is already being applied in the HTML and this computed property will only get called when `useRouterLink` is false.

assets/scss/style.scss
mirjamverloop: Can you move icons.scss to after hero.scss?

assets/scss/external-link-icon.scss
mirjamverloop: I see your point, you are right. I still think however that we should be able to reduce duplication of code and/or make use of icomoons naming conventions for classes. I played around a bit, there is a branch here: https://github.com/europeana/portal.js/tree/feat/EC-3835-external-links-reuse-css

What do you think of this approach?

monkemedia: @mirjamverloop can you please show me what you have done please?

PR #259, opened 2019-10-01
Nice up list view

mirjamverloop:
- Updated list view design
- Removed some unused styling / code
- Moved footer css to its own file


components/generic/ContentCard.vue
```diff
@@ -16,7 +16,7 @@
       />
       <b-card-body>
         <b-card-title>
-          {{ title }}
+          {{ title.length > 90 ? title.substring(0, 90) + $t('formatting.ellipsis') : title }}
```
rwd: Instead of having this logic in the template, and also duplicating it in the search result component, let's make it a Vue filter named `truncate` in plugins/vue-filters.js. SearchBarPill.vue could also be made to use it.

mirjamverloop: @rwd good suggestion, thanks. I have added the ellipsis as an argument on the truncate filter. Does that make sense to you?

components/search/SearchResult.vue
```diff
@@ -3,31 +3,28 @@
...
         :data-field-name="key"
         data-qa="result field"
       >
-        <pre v-if="!Array.isArray(value)">
-          <code>{{ value }}</code>
-        </pre>
-        <template v-else-if="value.length === 1">
-          {{ value[0] }}
+        <template v-if="Array.isArray(value) && value.length === 1">
+          {{ key === 'dcTitle' ? $options.filters.truncate(value[0], 90, $t('formatting.ellipsis')) : value[0] }}
```
rwd: Let's break this up:
```js
          <template v-if="key === 'dcTitle'">
            {{ value[0] | truncate(90, $t('formatting.ellipsis')) }}
          </template>
          <template v-else>
            {{ value[0] }}
          </template>
```

PR #265, opened 2019-10-02
Optimise images from Contentful

rwd:


plugins/vue-filters.js
```diff
@@ -16,3 +16,10 @@ Vue.filter('truncate', (val, char, ellipsis) => {
   return val.length > char ? val.substring(0, char) + ellipsis : val;
 });
 
+Vue.filter('optimisedImageUrl', (imageUrl) => {
+  const hostname = imageUrl.match(/\/\/([^/]+)\//)[1];
+  if (hostname === 'images.ctfassets.net') {
+    return imageUrl + '?fm=jpg&fl=progressive&q=25';
```
lbiedinger: q=25 will be too much, it results in quite noticeable artifacts.

rwd: Increased to 50.

components/browse/BrowseContentCard.vue
```diff
@@ -35,12 +39,15 @@
           } else {
```
monkemedia: Don't need an else here as we have a return above it

components/browse/BrowseContentCard.vue
```diff
@@ -35,12 +39,15 @@
...
             return this.fields.image;
           }
-        } else if (this.fields.image && this.fields.image.fields && this.fields.image.fields.file) {
+        } else if (this.imageIsContentfulAsset) {
           return this.fields.image.fields.file.url;
         } else {
```
monkemedia: Same here

pages/index.vue
```diff
@@ -4,7 +4,8 @@
   >
     <HeroBanner
       v-if="page.primaryImageOfPage"
-      :hero-image="page.primaryImageOfPage.fields.image.fields.file.url"
+      :image-url="page.primaryImageOfPage.fields.image.fields.file.url"
+      :image-content-type="page.primaryImageOfPage.fields.image.fields.file.contentType"
       :headline="page.primaryImageOfPage.fields.headline"
       :description="page.primaryImageOfPage.fields.description"
       :identifier="page.primaryImageOfPage.fields.identifier"
```
monkemedia: I wonder if it would make better reading if we used computed properties here? What are your thoughts?

rwd: Good thinking.

PR #271, opened 2019-10-04
Image compare

monkemedia:


components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+      },
+
+      imageRight: {
+        type: String,
+        required: true
+      },
+
+      imageLeftText: {
+        type: String,
+        default: 'This is an image'
```
rwd: Let this be blank by default.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+      },
+
+      imageRightText: {
+        type: String,
+        default: 'This is an image'
```
rwd: Let this be blank by default.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+    computed: {
+      leftImageClip() {
+        return {
+          clip: `rect(auto, ${this.imageWidth * this.sliderPosition}px, auto, auto)`
+        };
+      },
+
+      sliderBarPosition() {
+        return {
+          left: `${this.imageWidth * this.sliderPosition - this.sliderWidth / 2}px`
```
rwd: Let's move the calculation out into a `const` and just have that in the `${}`, to improve readability.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+        imageWidth: 0,
+        sliderWidth: 0,
+        sliderPosition: 0.5
+      };
+    },
+
+    computed: {
+      leftImageClip() {
+        return {
+          clip: `rect(auto, ${this.imageWidth * this.sliderPosition}px, auto, auto)`
```
rwd: Let's move the calculation out into a `const` and just have that in the `${}`, to improve readability.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+      ref="slider"
+      class="compare-image__slider"
+      :style="sliderBarPosition"
+    >
+      <span class="compare-image__slider__bar" />
+      <button
+        :class="{ 'is-active' : dragging }"
+        class="compare-image__slider__handle"
+      >
+        <span class="sr-only">Slider Handle</span>
```
rwd: Why is this span here?

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,225 @@
...
+    &__slider {
+      width: $slider-width;
+      height: 100%;
+      position: absolute;
+      top: 0;
+      justify-content: center;
+      align-items: center;
+      display: flex;
+
+      // &__bar {
```
rwd: Remove this comment block.

components/generic/CompareImageSlider.vue
monkemedia: This is for accessibility. Using the class `sr-only` (screen reader only) hides the text from the browser but allows screen readers to let the user know what the component is.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+  <figure
+    ref="container"
+    class="compare-image"
+    data-qa="compare image"
+  >
+    <img
+      ref="imageLeft"
+      :src="imageLeft"
+      :alt="imageLeftText"
+      class="compare-image__image-left"
```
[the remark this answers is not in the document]
monkemedia: Removed

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+      justify-content: center;
+      align-items: center;
+      display: flex;
+
+      .slider-handle {
+        display: flex;
+        position: absolute;
+        background: rgba(255, 255, 255, 0.5);
+        width: $slider-width;
+        height: 48px;
```
mirjamverloop: If you rename the $slider-width in something more generic, it can be used for this height as well.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+    .image-right {
+      width: 100%;
+    }
+
+    .image-left {
+      position: absolute;
+      top: 0;
+      left: 0;
+    }
+
```
mirjamverloop: You might be able to clean this up slightly and get rid of the classes, eg: 

```
img {
    width: 100%;

    &:nth-child(1) {
      position: absolute;
      top: 0;
      left: 0;
    }
  }
```

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+    .slider {
+      width: $slider-width;
+      height: 100%;
+      position: absolute;
+      top: 0;
+      justify-content: center;
```
mirjamverloop: I guess you are fine just using the align-items here.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+      align-items: center;
+      display: flex;
+
+      .slider-handle {
+        display: flex;
```
mirjamverloop: I am not sure if you need this display: flex.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  $slider-width: 48px;
+
+  .compare-image {
+    position: relative;
+    overflow: hidden;
```
mirjamverloop: It might be good to introduce a height or max-height, for example if images are of different sizes and to avoid having a compare thing that is too large in height.

components/generic/CompareImageSlider.vue
mirjamverloop: I am wondering if it makes sense here. When using a screen reader it is good to know that there is this image comparison component, but using a screen reader you cannot control the handler I guess. And most people that use a screen reader cannot actual see the comparison.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+      <span class="slider-bar" />
+      <button
+        :class="{ 'is-active' : dragging }"
+        class="slider-handle"
+        data-qa="compare image slider handler"
+      >
+        <span class="sr-only">Slider Handle</span>
+      </button>
+    </div>
+  </figure>
```
mirjamverloop: I am not sure if it is part of this task, so it might be worth discussing with PO, but I can imagine that there is a paragraph with description here as well. That says something like: "On the left, you see Person A in here twenties. On the right, you can see Person A in here fifties".

tests/unit/components/generic/CompareImageSlider.spec.js
```diff
@@ -0,0 +1,86 @@
+import { createLocalVue, shallowMount } from '@vue/test-utils';
+import BootstrapVue from 'bootstrap-vue';
+import sinon from 'sinon';
+import CompareImageSlider from '../../../../components/generic/CompareImageSlider.vue';
+
+const localVue = createLocalVue();
+localVue.use(BootstrapVue);
+
+const factory = () => shallowMount(CompareImageSlider, {
```
lbiedinger: to test the setImageWidth call do this:
```
const factory = (setImageWidthSpy) => shallowMount(CompareImageSlider, {
  localVue,
  propsData: {
    imageLeft: 'img/landscape.jpg',
    imageRight: 'img/portrait.jpg'
  },
  methods: { setImageWidth: setImageWidthSpy }
});

describe('components/generic/CompareImageSlider', () => {

  it('calls `setImageWidth` method when browser has been resized', () => {
    const setImageWidth = sinon.spy();
    factory(setImageWidth);

    global.window.dispatchEvent(new Event('resize'));

    setImageWidth.should.have.callCount(2);
  });

  ...
```
This won't execute the actual function any more, but sets up the spy before mounting, so we can refer to it later. If this interferes with the other tests, then we could have a specific factory definition just for that one example.

components/generic/CompareImageSlider.vue
monkemedia: Are we assuming all users who are blind are 100% blind?

components/generic/CompareImageSlider.vue
monkemedia: The reason for keeping it simple like this is that there is currently no design for the compare image component. The approach allows us to quickly apply new changes when the designs come through.

components/generic/CompareImageSlider.vue
monkemedia: This is required to keep the handler vertically centered

components/generic/CompareImageSlider.vue
monkemedia: Should we compare images that do not have the same dimensions? @rwd What are your thoughts?

components/generic/CompareImageSlider.vue
monkemedia: This can be done in a different ticket as its not part of the criteria

components/generic/CompareImageSlider.stories.js
```diff
@@ -0,0 +1,21 @@
+import { storiesOf } from '@storybook/vue';
+
+import CompareImageSlider from './CompareImageSlider.vue';
+
+storiesOf('Generic', module)
+  .add('Compare Image Slider', () => ({
+    components: { CompareImageSlider },
+    data() {
+      return {
+        imageLeft: 'https://www.fillmurray.com/640/360',
```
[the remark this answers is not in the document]
monkemedia: Would we compare images like this? @rwd thoughts?

tests/unit/components/generic/CompareImageSlider.spec.js
monkemedia: BOOM!!!! Thank you Lutz. Works a treat :)

components/generic/CompareImageSlider.vue
lbiedinger: We'll need to handle different size images. They may not always be drastically different dimensions, but should always be made to fit.

components/generic/CompareImageSlider.stories.js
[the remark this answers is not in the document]
lbiedinger: This is more about the fact that we want to avoid using external dependencies.
However using a landscape vs. portrait comparison is also a good example to show how we will handle this case.

components/generic/CompareImageSlider.vue
mirjamverloop: No, you are right. And not all screen readers users are not using a mouse. Leave it the way it is.

components/generic/CompareImageSlider.vue
```diff
@@ -0,0 +1,224 @@
...
+        }
+      }
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  $slider-width: 48px;
+
+  .compare-image {
```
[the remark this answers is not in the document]
lbiedinger: Not to be implemented now. We'll leave it to whoever is creating the compare sliders to make sure the images have the correct dimensions.

components/generic/CompareImageSlider.vue
mirjamverloop: There is no difference in styling when you delete it. As far as I can see.

components/generic/CompareImageSlider.vue
mirjamverloop: I agree with you, but I foresee changes to the handle/button and not necessarily to the dimensions and positioning of the images.

PR #278, opened 2019-10-14
Contentful migrations

rwd:


lbiedinger:
We should ignore the similar blocks of code linting for the migrations directory.

rwd:
@lbiedinger regarding:
> We should ignore the similar blocks of code linting for the migrations directory.
I agree, but consulting the [CodeClimate documentation on Excluding Files and Folders](https://docs.codeclimate.com/docs/excluding-files-and-folders):
> Exclusions can only be made at the global level (excluding code from all analysis) or at the plugin-level (excluding code only from specific third-party plugins). Currently, exclusions cannot be made for individual maintainability checks.
This suggests to me that we would either exclude migrations/ from all checks, or have to disable the similar lines check outright on all code, neither of which we want.

lbiedinger:
You're right, I thought there was something like `/* eslint-disable */`, but not for codeclimate checks.

PR #279, opened 2019-10-14
Hero banner as image with attribution

rwd:


pages/index.vue
```diff
@@ -35,6 +39,12 @@
       hero() {
         return this.page.primaryImageOfPage ? this.page.primaryImageOfPage.fields : null;
       },
+      heroDescription() {
+        return this.hero.description || this.page.description;
+      },
+      heroHeadline() {
+        return this.page.headline || this.hero.headline || this.page.name;
+      },
```
mirjamverloop: I guess we are going to remove the hero banner at some point, but should hero and page go in the same order in both heroDescription and heroHeadline?

rwd: Good point. Done in 0b73a23

PR #280, opened 2019-10-14
Replace parse-domain

monkemedia:


components/generic/SmartLink.vue
```diff
@@ -56,14 +54,10 @@
...
-        if (destDomain === null) {
-          return false;
-        }
-        const dest = [destDomain.domain, destDomain.tld].join('.');
-        const currentDomain = this.$store.state.request.domain;
-        const current = currentDomain;
-        return dest !== current;
+        const path = this.destination;
+        if (typeof path !== 'string' || path.startsWith('/')) return false;
+        const internalDomain = process.env.INTERNAL_LINK_DOMAIN;
```
rwd: The env var needs to be documented in /.env.example

components/generic/SmartLink.vue
```diff
@@ -56,14 +54,10 @@
...
+        return !path.includes(internalDomain);
```
rwd: This could in theory return false positives, if for instance `process.env.INTERNAL_LINK_DOMAIN=.europeana.eu` and `this.destination = 'http://www.example.org/pages/www.europeana.eu';`

Let's test the actual hostname ends with the domain in the env var thus:
```js
      isExternalLink() {
        const path = this.destination;
        const hostnamePattern = /\/\/([^/:]+)/;
        if (typeof path !== 'string' || !hostnamePattern.test(path)) return false;
        const hostname = path.match(hostnamePattern)[1];
        const internalDomain = process.env.INTERNAL_LINK_DOMAIN;
        return !hostname.endsWith(internalDomain);
      }
```

components/generic/SmartLink.vue
rwd: Let's make `internalDomain` a data property, defaulting to the env var. It will then only need to be set once, and this computed function can then more easily be unit tested.

PR #282, opened 2019-10-15
Initial setup exhibition page

mirjamverloop:


pages/exhibition/_.vue
```diff
@@ -0,0 +1,68 @@
+<template>
```
lbiedinger: this will need to be in pages/exhibitions/_.vue

pages/exhibition/_.vue
```diff
@@ -0,0 +1,68 @@
...
+            return;
+          }
+          return {
+            page: response.items[0].fields
+          };
+        })
+        .catch((e) => {
+          error({ statusCode: 500, message: e.toString() });
+        });
+    }
```
lbiedinger: add:
```
head() {
      return {
        title: this.page.headline
      };
    }
```

pages/exhibition/_.vue
```diff
@@ -0,0 +1,123 @@
...
+              v-html="credits"
+            />
+            <!-- eslint-enable vue/no-v-html -->
+          </article>
+        </b-col>
+      </b-row>
+      <b-row>
+        <b-col>
+          <!-- TODO: add links to chapters, remove h2 -->
+          <h2>Chapters</h2>
```
monkemedia: Should be localised

lbiedinger: This is only a placeholder, it'll be replaced by an actual component when we implement EC-3886. I don't think it's worth adding to the string to the locale file if it'll be removed later.

pages/exhibition/_.vue
```diff
@@ -0,0 +1,123 @@
...
+        return this.page.name;
+      },
+      lead() {
+        return this.page.headline;
+      },
+      heroImage() {
+        return this.hero ? this.hero.image.fields.file : null;
+      },
+      mainContent() {
+        return marked(this.page.text);
```
rwd: As there are two computed methods here, which solely convert markdown to HTML (and will be plenty elsewhere), let's introduce a Vue filter for this so that in the template we can just do:
```html
            <div
              v-html="text | markup"
            />
```

lbiedinger: this doesn't work so nicely due to this: https://github.com/vuejs/vue/issues/4080

PR #293, opened 2019-10-17
Add primaryImageOfPage to blogPosting

rwd:


pages/blog/_.vue
```diff
@@ -5,6 +5,19 @@
         cols="12"
         md="9"
       >
+        <HeroImage
+          v-if="hero"
+          :image-url="heroImage.url"
+          :image-content-type="heroImage.contentType"
+          :identifier="hero.identifier"
+          :citation="hero.citation"
```
lbiedinger: we can get rid of identifier and citation here too.

pages/blog/_.vue
```diff
@@ -5,6 +5,19 @@
...
+          :rights-statement="hero.license"
+          :name="hero.name"
+          :provider="hero.provider"
+          :creator="hero.creator"
+          :url="hero.url"
+          class="blog"
```
lbiedinger: this class isn't actually needed, it's just something left over from when I tried to style the name output on the hero image.

PR #294, opened 2019-10-17
Enforce locale in URLs

rwd:
* Introduces a custom router middleware to handle locale detection and enforcement in URLs
* Adds Nightwatch configuration to support testing browsers having different preferred languages
* Simplifies size-limit config to just client & server bundles


middleware/l12n.js
```diff
@@ -0,0 +1,46 @@
...
+// package but with additional redirection handling for requests to URL paths
+// without the language in them.
+
+export default ({ app, route, redirect, req }) => {
+  // Is there a locale in the URL path already?
+  const routePathLocale = route.path.match(/^\/([a-z]{2})(\/|$)/);
+  if (routePathLocale) {
+    // Store it in the cookie, indicating user's current preference e.g. from
+    // using language selector.
+    app.$cookies.set('i18n_locale_code', routePathLocale[1]);
```
lbiedinger: this should also check the route matches the known list of locales, otherwise you can set a broken cookie.
for example visit:
http://localhost:3000/rf/about-us
then
http://localhost:3000/explore

PR #297, opened 2019-10-18
Image comparison section

rwd:
* Extract attribution into its own component for reuse in different places
* In the browse sections, detect imageComparison entries and display the image compare component for them

### DIFF
--- a/assets/scss/licenses.scss
+++ b/assets/scss/licenses.scss
@@ -1,41 +1,3 @@
-figcaption {
-  background: $white;
-  border-top-left-radius: $border-radius-small;
-  border-top-right-radius: $border-radius-small;
-  bottom: 0;
-  font-size: 0.625rem;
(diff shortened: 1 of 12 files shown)


assets/scss/licenses.scss
```diff
@@ -45,3 +45,41 @@
...
+  background: $white;
+  border-top-left-radius: $border-radius-small;
+  border-top-right-radius: $border-radius-small;
+  bottom: 0;
+  font-size: 0.625rem;
+  left: 1rem;
+  line-height: 1;
+  padding: 0.5rem;
+
+  &.is-external-link:after {
```
lbiedinger: I'm not seeing this icon, I think it may be wrapping onto the next line.

rwd: Fixed in ab1950d

PR #306, opened 2019-10-23
Update styling attributions

mirjamverloop:


components/generic/CompareImageSlider.vue
```diff
@@ -155,6 +156,18 @@
         this.dragging = false;
       },
 
+      showHideAttribution() {
+        if (this.dragging) {
+          if (this.sliderPosition < 0.2) {
+            this.hideAttribution = 'left';
+          } else if (this.sliderPosition > 0.8) {
```
monkemedia: I'm wondering if the 0.2 and 0.8 should be dynamic, as these will change when you reduce the size of the browser window ie Mobile.

components/generic/CompareImageSlider.vue
```diff
@@ -155,6 +156,18 @@
         this.dragging = false;
       },
 
+      showHideAttribution() {
+        if (this.dragging) {
```
monkemedia: Is this required since this method is only called when the user is dragging. Please look at 

```
if (this.dragging) {
          this.sliderPosition = pos / this.imageWidth;
          this.showHideAttribution();
}
```

components/generic/CompareImageSlider.vue
mirjamverloop: I kept the values static for now. And moved the attributions to below the images for small screens.

components/generic/CompareImageSlider.vue
mirjamverloop: Good suggestion, removed the check here.

components/generic/CompareImageSlider.vue
```diff
@@ -72,6 +71,11 @@
     },
 
     props: {
+      hideAttribution: {
```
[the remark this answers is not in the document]
mirjamverloop: I have updated this.

PR #309, opened 2019-10-24
Record API plugin: gather isShownBy and hasView into media

rwd:


pages/record/_.vue
```diff
@@ -15,13 +15,13 @@
         class="mb-3 px-0"
       >
         <MediaPresentation
```
lbiedinger: let's add a v-if here.
otherwise for http://localhost:3000/en/record/2048441/item_BMXIEQWQ2YHKW2S5C3XQCH2VISBNSV5E
you'll get an error.

PR #310, opened 2019-10-24
Layout rich media card

mirjamverloop:
- Includes preparations for image media card


plugins/vue-filters.js
```diff
@@ -28,3 +28,17 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType) => {
   }
   return imageUrl;
 });
+
+Vue.filter('inCurrentLanguage', (val, currentLocale) => {
```
rwd: This needs to account for the fact that the keys of `val` may be three-letter codes.

For instance, on the record /2023813/o_12654, the title is in French but with the code `fra` so it is not detected or rendered when viewing the site at /fr.

I think we will therefore need to do something like:
* Extend the locale definitions in /plugins/i18n/locales.json to also include the three-letter codes in a new property named `isoAlpha3`
* In this filter, find the three-letter code for `currentLocale` and also prepend that onto `languageKeys`

pages/record/_.vue
```diff
@@ -9,31 +9,53 @@
...
+              :url="edmIsShownBy.about"
+              :width="edmIsShownBy.ebucoreWidth"
+              :height="edmIsShownBy.ebucoreHeight"
+              class="mb-3"
+            />
+            <p
+              v-if="description"
+              class="description"
+            >
+              {{ description | inCurrentLanguage($i18n.locale) }}
```
rwd: For description (only) we want to honour white space to permit presentation of paragraphs, e.g. in the record /125/contributions_42580010_c414_0137_3824_6eee0af7162b

I would suggest trying the CSS `white-space: pre-wrap;` which has the additional benefit of presenting poetry as intended.

pages/record/_.vue
```diff
@@ -91,8 +125,47 @@
     },
     head() {
       return {
-        title: this.$t('record')
+        title: (this.title && this.$options.filters.inCurrentLanguage(this.title, this.$i18n.locale)) ?
```
rwd: This can end up calling inCurrentLanguage with the same parameters twice. Store the return value before using it.

pages/record/_.vue
```diff
@@ -9,31 +9,53 @@
...
-          :url="edmIsShownBy.about"
-          :width="edmIsShownBy.ebucoreWidth"
-          :height="edmIsShownBy.ebucoreHeight"
-        />
+        <div class="card p-3">
+          <div
+            class="card-grid"
+            :class="isRichMedia && 'card-grid-richmedia'"
+          >
+            <h1
```
rwd: As this may be in a language other than the user is viewing the UI, we need to include the `lang` attribute here.

Given that we will need to do this on every metadata field, or at least those not in the same language as the UI, I wonder if we should have a new component to handle this, or refactor components/record/LangMap.vue for this purpose.

pages/record/_.vue
```diff
@@ -64,13 +87,24 @@
...
         error: null,
         image: null,
         fields: null,
         media: null,
+        title: null,
         edmIsShownBy: {}
       };
     },
+    computed: {
+      isRichMedia() {
```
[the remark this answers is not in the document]
mirjamverloop: I have moved it to a plugin "media.js", the isRichMedia check for video and oEmbed as the PDF is currently an image with a link.

plugins/vue-filters.js
```diff
@@ -28,3 +28,17 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType) => {
...
+  let languageKeys = ['en', 'eng', 'def', 'und'];
+  if (currentLocale !== 'en') {
+    languageKeys.unshift(currentLocale);
+  }
+
+  for (let key of languageKeys) {
+    if (val[key]) {
+      return val[key][0];
```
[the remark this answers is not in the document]
mirjamverloop: I have updated the Vue filter in order to output all values. There are subtitles with multiple values as well, eg: 150/_REB01______________________000250438

plugins/vue-filters.js
mirjamverloop: I have updated the locales and the filter so it accepts 2 and 3 character codes.

pages/record/_.vue
mirjamverloop: `white-space: pre-wrap;` adds some additional (extra whitespace before the first line of text and indentation - always - of the first line) styling, which I couldn't get around. I tried to match the behaviour on the current portal and updated the filter accordingly.

pages/record/_.vue
mirjamverloop: I have made this part of the filter. The filter returns the text output and the lang code that can be used on the HTML element. 

The language of the document `html lang="en"` was always English, even if you switched languages, so I made a fix for that too.

plugins/media.js
```diff
@@ -0,0 +1,20 @@
...
+    (mimeType === 'video/webm') ||
+    ((mimeType === 'video/mp4') && (codec === 'h264'));
+}
+
+export function isOEmbed(oembedUrl) {
+  return oEmbeddable(oembedUrl);
+}
+
+// TODO: as the pdf is currently just an image with a link, it is not marked as "rich media", this might change in the future
+export function isRichMedia(mimeType, codec, oembedUrl) {
```
rwd: Let's change the param name from `oembedUrl` to `mediaUrl` as it may or may not be oEmbeddable.

components/generic/LanguageSelector.vue
```diff
@@ -35,6 +35,13 @@
           return locale.code === this.$i18n.locale;
         });
       }
+    },
+    head() {
+      return {
+        htmlAttrs: {
+          lang: this.$i18n.locale
```
rwd: This does not seem to change the attribute if I switch to another language.

To have the desired effect, we also need to remove from nuxt.config.js the lines https://github.com/europeana/portal.js/blob/master/nuxt.config.js#L17-L19

pages/record/_.vue
```diff
@@ -9,32 +9,67 @@
...
+                :lang="titlesInCurrentLanguage.mainTitle.code"
+              >
+                {{ titlesInCurrentLanguage.mainTitle.value }}
+              </h1>
+              <!-- eslint-disable vue/no-v-html -->
+              <p
+                v-if="titlesInCurrentLanguage.altTitle"
+                :lang="titlesInCurrentLanguage.altTitle.code"
+                class="font-weight-bold"
+                v-html="titlesInCurrentLanguage.altTitle.value"
```
rwd: I don't think this needs to be in v-html. It should just be the text content of the element.

pages/record/_.vue
```diff
@@ -66,13 +102,40 @@
...
         image: null,
         fields: null,
-        media: null
+        media: null,
+        title: null,
+        edmIsShownBy: {}
       };
     },
     computed: {
+      titlesInCurrentLanguage() {
```
rwd: I suggest having two computed properties, one for mainTitle and one for altTitle. There may be a little duplication of logic, but it improves readability.

pages/record/_.vue
```diff
@@ -97,8 +160,46 @@
     },
     head() {
       return {
-        title: this.$t('record')
+        title: this.titleInCurrentLanguage ? this.titleInCurrentLanguage.value : this.$t('record')
```
rwd: This never returns anything other than the translated "Record".

plugins/europeana/record.js
```diff
@@ -37,7 +37,11 @@ function parseRecordDataFromApiResponse(response) {
     media: providerAggregation.webResources.filter((webResource) => {
       return (webResource.about === providerAggregation.edmIsShownBy) ||
         (providerAggregation.hasView || []).includes(webResource.about);
-    })
+    }),
+    title: providerProxy.dcTitle,
+    edmIsShownBy: providerAggregation.webResources.find((webResource) => {
```
rwd: This needs to be re-removed. edm:isShownBy is included in the media array as the first item.

pages/record/_.vue
```diff
@@ -66,13 +102,40 @@
...
       return {
+        altTitle: null,
+        description: null,
         error: null,
         image: null,
         fields: null,
-        media: null
+        media: null,
+        title: null,
+        edmIsShownBy: {}
```
rwd: This needs to be re-removed. edm:isShownBy is included in the media array as the first item.

pages/record/_.vue
```diff
@@ -66,13 +102,40 @@
...
+        }
+      },
+      descriptionInCurrentLanguage() {
+        if (!this.description) {
+          return false;
+        }
+        return this.$options.filters.inCurrentLanguage(this.description, this.$i18n.locale);
+      },
+      isRichMedia() {
+        return isRichMedia(this.edmIsShownBy.ebucoreHasMimeType, this.edmIsShownBy.edmCodecName, this.edmIsShownBy.about);
```
rwd: Refer to `selectedMedia` now instead of `edmIsShownBy`.

plugins/vue-filters.js
```diff
@@ -28,3 +30,25 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType) => {
   }
   return imageUrl;
 });
+
+Vue.filter('inCurrentLanguage', (val, locale) => {
+  let languageKeys = ['en', 'eng', 'def', 'und'];
+  const currentLocale = locales.find(l => l.code === locale);
+
+  if (currentLocale.code !== 'en' || currentLocale.isoAlpha3 !== 'eng') {
```
rwd: I think that this is redundant:
```js
 || currentLocale.isoAlpha3 !== 'eng'
```

plugins/vue-filters.js
```diff
@@ -28,3 +30,25 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType) => {
...
+  }
+
+  for (let key of languageKeys) {
+    if (val[key]) {
+      let htmlLang = '';
+      if (key !== 'def' && key !== 'und') {
+        const langCode = key.length === 3 ? locales.find(l => l.isoAlpha3 === key).code : key;
+        htmlLang = currentLocale.code !== langCode ? langCode : '';
+      }
+      return { 'code': htmlLang, 'value': val[key].join('\n\n').replace(/\n/g, '<br/>') };
```
rwd: The newline to `<br/>` conversion should only occur on descriptions (for now). I suggest making this a new filter and calling it from the record page view on description only. That will also save us having to use `v-html` all over the place.

plugins/vue-filters.js
```diff
@@ -28,3 +30,25 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType) => {
...
+Vue.filter('inCurrentLanguage', (val, locale) => {
+  let languageKeys = ['en', 'eng', 'def', 'und'];
+  const currentLocale = locales.find(l => l.code === locale);
+
+  if (currentLocale.code !== 'en' || currentLocale.isoAlpha3 !== 'eng') {
+    languageKeys.unshift(currentLocale.code);
+    languageKeys.unshift(currentLocale.isoAlpha3);
+  }
+
+  for (let key of languageKeys) {
```
rwd: Could you add a comment explaining what all of this is doing as I can imagine it being tricky to scan when it's not fresh in our minds.

pages/record/_.vue
mirjamverloop: This has been updated with the rework on the headings.

pages/record/_.vue
mirjamverloop: With the new insights, as we discussed today, I kept one computed property for titles.

plugins/vue-filters.js
mirjamverloop: I have created a new filter.

plugins/vue-filters.js
mirjamverloop: I have added some comments to the filter.

components/generic/LanguageSelector.vue
mirjamverloop: It works fine for me, can you check again?

lbiedinger: it works only when there's a language selector on the page, as in it won't work without having `ENABLE_LANGUAGE_SELECTOR=1`

PR #312, opened 2019-10-30
Move image from left to right

mirjamverloop:


components/search/SearchResult.vue
```diff
@@ -44,6 +34,17 @@
         </ul>
       </div>
     </b-media-body>
+    <b-media-aside class="media-image">
+      <b-img
+        v-if="result.edmPreview"
+        slot="aside"
+        :src="result.edmPreview"
+        alt=""
```
monkemedia: Should we populate the alt tag with a property?

mirjamverloop: Eventually, yes. The alt would ideally be a description of the actual image, something like "Painting of a lady with a red dress and a green dog". We don't have such descriptions yet, that is why we left it blank when we set this up a couple of months ago. 

As this ticket was just about the styling, I would like to suggest to keep it like this for now. Is that ok?

PR #313, opened 2019-10-30
Added conditionals for related card, move CSS to cards.scss

mirjamverloop:


assets/scss/cards.scss
```diff
@@ -1,3 +1,94 @@
...
+  a {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .card-img {
+    background-position: center center;
+    background-size: cover;
+    border-top-left-radius: 4px;
+    border-top-right-radius: 4px;
```
monkemedia: Could we use shorthand here? `border-radius: 4px 4px 0 0`

components/generic/ContentCard.vue
```diff
@@ -1,8 +1,10 @@
 <template>
   <b-card
-    class="mb-4 text-left"
+    class="mb-4 text-left content-card"
     data-qa="content card"
     no-body
+    :class="isRelated && 'related-card'"
```
monkemedia: Can you explain whats going on here? 'related-card' will only show if 'isRelated' equals true?

mirjamverloop: Yes, indeed. See also https://michaelnthiessen.com/conditional-class-binding-vue#guard-expressions

assets/scss/cards.scss
mirjamverloop: Good idea, I have updated it to a shorthand version.

components/generic/ContentCard.vue
monkemedia: I wonder if it's better to take the standard approach which I think is easier to read

`:class="{ 'related-card' : isRelated }"`

PR #317, opened 2019-10-31
EC-3502 - clear all filters

monkemedia:


components/search/SearchInterface.vue
```diff
@@ -295,7 +302,34 @@
...
 </script>
+
+<style lang="scss" scoped>
+  @import "./assets/scss/variables.scss";
+  @import "./assets/scss/icons.scss";
+
+  .clear-all {
+    background: none;
+    border: none;
+    color: $lightblue;
```
lbiedinger: this causes an accessibility violation due to colour contrast. Changing it to $blue resolves it and is still more or less in line with the design.

components/search/SearchInterface.vue
```diff
@@ -295,7 +302,34 @@
           }
         }
         return updated;
+      },
+      clearFilters() {
+        let qf = this.$route.query.qf;
```
lbiedinger: this doesn't work when only the REUSABILITY facet is set or only one single other facet.
When it's only reusability it'll be null and when it's a single facet it's a string and won't have a .filter method.
Either use `qf = this.$route.query.qf || [];` and a check that it's an array, or perhaps do: `let qf = [].concat(this.$route.query.qf);` and then remove any nil values.

components/search/SearchInterface.vue
```diff
@@ -33,6 +33,13 @@
             :selected="selectedFacets[facet.name]"
             @changed="changeFacet"
           />
+          <button
+            v-if="Object.keys(selectedFacets).length"
```
lbiedinger: this will also show the button when only the content tier toggle is active, but the button won't do anything.

lbiedinger: if the count is 1 we could check `contentTierActiveState` which should be available as a computed porperty here.

lbiedinger: ...in fact I reckon this should check for only any of the defaultFacets values being in 'this.selectedFacets.' So probably will need it's own computed property or method.

tests/features/search/faceting.feature
```diff
@@ -139,3 +139,17 @@ Feature: Search faceting
...
+  Scenario: Clear filters using using `clear all filter` button
+    When I visit the `search page`
+    And I click the `COUNTRY dropdown button`
+    And I check the "France" "COUNTRY" checkbox
+    And I click the `COUNTRY apply button`
+    And I click the `TYPE dropdown button`
+    And I check the "IMAGE" "TYPE" checkbox
+    And I click the `TYPE apply button`
+    And I wait 2 seconds
+    Then I should be on `/en/search?page=1&qf=COUNTRY%3A%22France%22&qf=TYPE%3A%22IMAGE%22&query=&view=grid`
```
lbiedinger: this isn't quite part of the thing we should test here. I reckon get rid of this line and move the `Then` to the last line, since that's the only expectation of this test really.

components/search/SearchInterface.vue
```diff
@@ -33,6 +33,14 @@
             :selected="selectedFacets[facet.name]"
             @changed="changeFacet"
           />
+          <button
+            v-if="hasDefaultFacets()"
```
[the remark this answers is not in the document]
monkemedia: I believe you can only remove the parentheses when the method is called from an event ie. @click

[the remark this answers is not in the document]
rwd: They are definitely not needed, so please remove them.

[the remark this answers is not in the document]
monkemedia: Ok

PR #319, opened 2019-11-04
Retrieve related entities from Entity API

rwd:


pages/record/_.vue
```diff
@@ -131,17 +132,30 @@
...
         title: null
       };
     },
     computed: {
+      europeanaAgents() {
+        return (this.agents || []).filter((agent) => agent.about.startsWith('http://data.europeana.eu/agent/'));
+      },
+      europeanaConcepts() {
+        return (this.concepts || []).filter((concept) => concept.about.startsWith('http://data.europeana.eu/concept/'));
+      },
```
monkemedia: Could these ever start with `https://`?

rwd: Certainly not in the foreseeable future, and if they did, it would be a transitioned change we would know about well in advance and could account for then.

PR #320, opened 2019-11-04
Added header with title and headline for when hero image isn't present

monkemedia:


pages/index.vue
```diff
@@ -17,6 +17,15 @@
       :url="hero.url"
     />
     <b-container>
+      <header v-if="!hero">
+        <h2>{{ page.name }}</h2>
+        <p
+          v-if="headline"
```
rwd: This should be page.headline

pages/index.vue
```diff
@@ -17,6 +17,15 @@
       :url="hero.url"
     />
     <b-container>
+      <header v-if="!hero">
+        <h2>{{ page.name }}</h2>
```
rwd: This needs to be h1, not h2.

pages/index.vue
```diff
@@ -17,6 +17,15 @@
       :url="hero.url"
     />
     <b-container>
+      <header v-if="!hero">
```
mirjamverloop: Could you limit the width of the header? On small screens this if fine, but on larger screens it can be limited to 75% for readability.

PR #324, opened 2019-11-05
Core metadata section

lbiedinger:
To get a more user friendly(and complete) display of the records metada:

* In the record plugin deepmerge the provider & europeana proxy metadata.
* In the record plugin, perform an entity lookup on all metadata fields.
  * When an entity was detected, override the field value with the entity information.
* In the record plugin return two sets of fields: coreFields & fields this is to allow expandable sections later.
* The lookup for preferred metadata has been refactored to handle these entity objects.
* Introduction of a new EntityField component, which will render links to entity pages where entities are europeana entities.
* Also additional logic to only render 'lang' attributes when the language is different than that of the page.


plugins/europeana/record.js
```diff
@@ -12,39 +13,71 @@ function parseRecordDataFromApiResponse(response) {
...
+  return !checkNull(value);
+}
+
+/**
+ * Update
+ * @param field
+ * @param entities
+ * @returns {*}
+ */
+function dereferenceEntities(fields, entities) {
```
lbiedinger: `lookupEntities` ?

plugins/europeana/utils.js
```diff
@@ -11,3 +11,61 @@ export function apiError(error) {
...
+  let languageKeys = ['en', 'eng', 'def', 'und']; // predefined set of preferred/fallback languages
+  const currentLocale = locales.find(l => l.code === locale);
+
+  if (currentLocale.code !== 'en') { // if language code is not en, add both 2 and 3 character language code to preferred languages
+    languageKeys.unshift(currentLocale.code);
+    languageKeys.unshift(currentLocale.isoAlpha3);
+  }
+
+  let returnVal = { values: [] };
+  for (let key of languageKeys) { // loop through all language key to find a match
```
lbiedinger: languageKeys can be a distinct function to simplify the logic above out of this function.

components/record/MetadataField.vue
```diff
@@ -56,5 +90,14 @@
   ul {
     list-style: none;
     padding: 0;
+    li {
+      display: inline;
+      &:after {
```
mirjamverloop: You can have it even like this: `&:not(:last-child):after` and then remove the `&:last-child:after`

pages/record/_.vue
```diff
@@ -88,10 +99,32 @@
...
+            :name="name"
+            :field-data="value"
+            class="mb-3"
+          />
+        </div>
       </b-col>
       <b-col
         cols="12"
         lg="3"
+        style="background-color: #FFF"
```
mirjamverloop: Could you remove this, else it will probably be merged in.

components/record/EntityField.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <li
+    v-if="locale !== null"
+    :lang="locale"
```
mirjamverloop: The conditional for whether or not displaying the lang attribute introduces a lot of new duplication of code. I am wondering if it would work if you use `:lang="locale ? locale : false "` instead.

components/record/MetadataField.vue
```diff
@@ -1,52 +1,86 @@
 <template>
   <div
-    v-if="value"
+    v-if="hasValuesForLocale"
```
mirjamverloop: There is something wrong with this check, see for example /en/record/145/URN_RS_NAE_905fa4f7_4345_4d57_bb09_c3910aa3cf06cho. Where the label is shown for "type of object", plus an empty `<ul>`. While there shouldn't be anything shown in my opinion.

components/record/EntityField.vue
lbiedinger: right, https://vuejs.org/v2/guide/syntax.html#Attributes
It should be fine already just setting it to whatever is in the code/locale key, since I'm setting it to null when it's not required and only to an empty string for undefined

pages/record/_.vue
```diff
@@ -88,6 +81,27 @@
             :europeana-identifier="identifier"
           />
         </div>
+        <div
+          class="card p-3 mb-3"
```
mirjamverloop: Now there is always a card, even is there is no coreFields (or fields for the next one). This is an example where you have an empty card: en/record/2048012/Athena_Plus_ProvidedCHO_Openbaar_Kunstbezit_Vlaanderen_OKV1963_6

components/record/MetadataField.vue
lbiedinger: the logic for this has been fixed, but still needs to be refactored to reduce cognitive complexity.

components/record/MetadataField.vue
```diff
@@ -1,52 +1,79 @@
...
+            :about="value.about"
+            :locale="value.code"
+            data-qa="entity value"
+          />
+        </template>
+        <li
+          v-else
+          :key="index"
+          :lang="langMappedValues.code"
+          data-qa="value"
```
rwd: `"literal value"` perhaps?

pages/record/_.vue
```diff
@@ -100,27 +114,18 @@
...
-      </b-col>
-    </b-row>
   </b-container>
 </template>
 
 <script>
   import EntityCards from '../../components/entity/EntityCards';
   import MediaActionBar from '../../components/record/MediaActionBar';
   import AlertMessage from '../../components/generic/AlertMessage';
-  import WebResources from '../../components/record/WebResources';
```
rwd: Given this component is no longer used, delete it?

plugins/europeana/entity.js
```diff
@@ -258,6 +258,28 @@ export function getEntityDescription(entity) {
...
+  return !!uri.match(/^http:\/\/data\.europeana\.eu\/(concept|agent|place)\/base\/\d{1,9}$/);
+}
+
+/**
+ * From a URI split params as required by the portal
+ * @param {String} A URI to check
+ * @return {{type: String, identifier: string}} Object with the portal relevant identifiers.
+ */
+export function entityParamsFromUri(uri) {
+  const matched = uri.match(/^http:\/\/data\.europeana\.eu\/(concept|agent|place)\/base\/(\d{1,9})$/);
```
rwd: Why is the numeric ID limited to 9 digits?

plugins/europeana/record.js
```diff
@@ -12,39 +13,77 @@ function parseRecordDataFromApiResponse(response) {
...
+ * will match any literal values in  the 'def' key to about fields
+ * in any of the entities and return the related object instead of
+ * the plain string.
+ * @param field
+ * @param entities
+ * @return {Object[]} The fields with any entities as JSON objects
+ */
+function lookupEntities(fields, entities) {
+  let returnVal = fields;
+  for (const key of Object.keys(returnVal)) {
```
rwd: `for (const key in returnVal) {`

plugins/europeana/record.js
```diff
@@ -12,39 +13,77 @@ function parseRecordDataFromApiResponse(response) {
...
 
+function checkNull(value) {
+  return value === undefined;
+}
+function checkNotNull(value) {
+  return !checkNull(value);
+}
+
+/**
+ * Update an array of fields, in order to find linked entity data.
```
rwd: The param is not an array

plugins/europeana/record.js
```diff
@@ -12,39 +13,77 @@ function parseRecordDataFromApiResponse(response) {
...
+ * will match any literal values in  the 'def' key to about fields
+ * in any of the entities and return the related object instead of
+ * the plain string.
+ * @param field
```
rwd: "fields"

plugins/europeana/record.js
```diff
@@ -12,39 +13,77 @@ function parseRecordDataFromApiResponse(response) {
 
   const providerAggregation = edm.aggregations[0];
   const europeanaAggregation = edm.europeanaAggregation;
-  const providerProxy = edm.proxies.find((proxy) => {
-    return proxy.europeanaProxy === false;
-  });
+  const entities = [].concat(edm.concepts, edm.places, edm.agents).filter(checkNotNull);
```
rwd: I recommend reducing these into an object keyed by URI so that they can be looked up thus by `lookupEntities` making `matchEntity` redundant and eliminating its repeated scanning through all entities:

```js
const entities = [].concat(edm.concepts, edm.places, edm.agents)
  .filter(checkNotNull)
  .reduce((memo, entity) => {
    memo[entity.about] = entity;
    return memo;
  }, {});
```

plugins/europeana/record.js
```diff
@@ -12,39 +13,77 @@ function parseRecordDataFromApiResponse(response) {
...
+ * Update an array of fields, in order to find linked entity data.
+ * will match any literal values in  the 'def' key to about fields
+ * in any of the entities and return the related object instead of
+ * the plain string.
+ * @param field
+ * @param entities
+ * @return {Object[]} The fields with any entities as JSON objects
+ */
+function lookupEntities(fields, entities) {
+  let returnVal = fields;
```
rwd: This is redundant. Just use `fields` throughout this function.

plugins/europeana/utils.js
```diff
@@ -11,3 +11,69 @@ export function apiError(error) {
...
   return apiError;
 }
+
+const locales = require('../i18n/locales.json');
+
+function isEntity(value) {
+  return !!value && !!value.about;
+}
+
+function entityValues(values, locale) {
```
rwd: Reduce the cognitive complexity of this function by splitting thus:
```diff
diff --git a/plugins/europeana/utils.js b/plugins/europeana/utils.js
index 0ddd8dc..fdb4270 100644
--- a/plugins/europeana/utils.js
+++ b/plugins/europeana/utils.js
@@ -19,22 +19,18 @@ function isEntity(value) {
 }
 
 function entityValues(values, locale) {
-  let entityValues = [];
-  for (const value of values || []) {
-    if (isEntity(value)) {
-      let entityValue = {};
-      if (value.prefLabel)  {
-        let mappedValues = langMapValueForLocale(value.prefLabel, locale);
-        mappedValues['values'] = mappedValues['values'].length >= 1 ? mappedValues['values'] : [value.about];
-        entityValue = mappedValues;
-        entityValue['about'] = value.about;
-      } else {
-        entityValue = { code: '', values: [value.about], about: value.about };
-      }
-      entityValues = entityValues.concat(entityValue);
-    }
+  return (values || []).filter((value) => isEntity(value))
+    .map((value) => entityValue(value, locale));
+}
+
+function entityValue(value, locale) {
+  if (value.prefLabel)  {
+    const entityValue = langMapValueForLocale(value.prefLabel, locale);
+    entityValue.about = value.about;
+    if (entityValue.values.length === 0) entityValue.values = [value.about];
+    return entityValue;
   }
-  return entityValues;
+  return { code: '', values: [value.about], about: value.about };
 }
 
 /**
```

components/record/MetadataField.vue
```diff
@@ -1,52 +1,79 @@
...
+    <ul>
+      <template
+        v-for="(value, index) of langMappedValues.values"
+      >
+        <template
+          v-if="value.about"
+        >
+          <EntityField
+            v-for="(nestedValue, nestedIndex) of value.values"
+            :key="index + '_' + nestedIndex"
```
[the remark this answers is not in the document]
lbiedinger: to differentiate it from index in:
`      <template
        v-for="(value, index) of langMappedValues.values"
      >
`

pages/record/_.vue
lbiedinger: I'm going to leave the translations in for now as we may very well need them when implementing the carousel.

plugins/europeana/entity.js
lbiedinger: no reason other than it didn't seem like there would be entity IDs that high, I've however changed it to alow any number of digits.

pages/record/_.vue
lbiedinger: I think this is out of scope for now. It would mean moving the language lookups up to the record page instead of containing it to the metadata field component. Also for an initial implementation this will highlight how metadata is/isn't displayed and we know that we'll need to refactor this once it goes through another iteration of design where we'll probably be very unlikely to end up with the situation where there is no metadata at all to show.

components/record/EntityField.vue
```diff
@@ -0,0 +1,45 @@
...
+      SmartLink
+    },
+    props: {
+      value: {
+        type: String,
+        default: null
+      },
+      about: {
+        type: String,
+        default: null,
```
rwd: There is no use in having a default for a required prop.

components/record/MetadataField.vue
```diff
@@ -1,60 +1,100 @@
...
+      <template
+        v-for="(value, index) of langMappedValues.values"
+      >
+        <template
+          v-if="value.about"
+        >
+          <li
+            v-for="(nestedValue, nestedIndex) of value.values"
+            :key="index + '_' + nestedIndex"
+            :lang="locale"
```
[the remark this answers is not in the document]
lbiedinger: this has been addressed and an issue with entities that have prefLabels but no matching language has been fixed also.

components/record/MetadataField.vue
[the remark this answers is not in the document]
rwd: Indeed it is necessary.

PR #334, opened 2019-11-11
Number of selected "more filters"

monkemedia:


components/search/MoreFacetsDropdown.vue
```diff
@@ -45,7 +53,28 @@
...
+    data() {
+      return {
+        selected: {}
+      };
+    },
+
+    computed: {
+      selectedAmount() {
+        let sum = 0;
+        for (const key of Object.keys(this.selected)) {
```
lbiedinger: `        for (const key in this.selected) {`

components/search/MoreFacetsDropdownOptions.vue
```diff
@@ -8,6 +8,7 @@
     <b-form-checkbox-group
       class="option-group"
       plain
+      @change="selected"
```
lbiedinger: while this could be okay for now to demonstrate the selection working, this will need to be refactored to only update when the 'apply' button is clicked, not when an option is selected.

components/search/MoreFacetsDropdown.vue
monkemedia: `this.selected` is not iterable

assets/fonts/icomoon-config.json
```diff
@@ -129,21 +129,29 @@
     },
     {
       "selection": [
+        {
+          "order": 1,
+          "id": 2,
+          "prevSize": 28,
+          "code": 59653,
+          "name": "circle",
```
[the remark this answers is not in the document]
monkemedia: I have made this change, but the issue is we have less control over the size of the interpunct. If i try to increase the size of it, the line height becomes way to big causing the position to be out of line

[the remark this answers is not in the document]
lbiedinger: we could add 'font-weight: bold;' to make it thicker.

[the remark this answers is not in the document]
lbiedinger: Also the changes here should be removed if we're not making use of the 'circle' icon anywhere.

components/search/MoreFacetsDropdown.vue
lbiedinger: if you use 'in' instead of 'of' it'll work.

assets/scss/dropdown.scss
```diff
@@ -6,12 +6,25 @@
...
+    &.is-active {
+      > button {
+        color: $blue;
+      }
+    }
+
+    .selected-amount {
+      padding-left: .25rem;
+
+      &:before {
```
lbiedinger:
```
      &:before {
        content: '·';
        font-weight: bold;
      }
```

components/search/MoreFacetsDropdown.vue
```diff
@@ -45,7 +56,28 @@
...
+
+    data() {
+      return {
+        selected: {}
+      };
+    },
+
+    computed: {
+      selectedAmount() {
+        let sum = 0;
```
rwd: This whole function can be reduced to just:
```js
return Object.values(this.selected).flat().length;
```

components/search/MoreFacetsDropdown.vue
```diff
@@ -45,7 +56,28 @@
...
+        for (const key in this.selected) {
+          sum += this.selected[key].length;
+        }
+        return sum;
+      }
+    },
+
+    methods: {
+      updateSelected(value) {
```
rwd: This isn't so easy to scan.

Let's rewrite it as:
```js
      updateSelected(facetName, selectedFields) {
        Vue.set(this.selected, facetName, selectedFields);
      }
```
... with the accompanying change to `selected`, below.

components/search/MoreFacetsDropdownOptions.vue
```diff
@@ -74,6 +75,12 @@
         isActive: false,
         limitTo: 9
       };
+    },
+
+    methods: {
+      selected(value) {
+        this.$emit('selectedOptions', { [this.name]: [ ...value ] });
```
rwd: This isn't so easy to scan.

Let's rewrite it as:
```js
        this.$emit('selectedOptions', this.name, value);
```
... with the accompanying change to `updateSelected`, above.

components/search/MoreFacetsDropdown.vue
monkemedia: Sadly the flat() method doesn't work in IE Edge  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#Browser_compatibility

components/search/MoreFacetsDropdown.vue
monkemedia: Yeah much cleaner.

components/search/MoreFacetsDropdown.vue
rwd: How about this then:
```js
return [].concat(...Object.values(this.selected)).length;
```

monkemedia: Done

PR #335, opened 2019-11-13
Show or hide extended information

mirjamverloop:


pages/record/_.vue
```diff
@@ -212,10 +226,23 @@
     },
     async mounted() {
       this.relatedEntities = await searchEntities(this.europeanaEntityUris, { wskey: process.env.EUROPEANA_ENTITY_API_KEY });
+
+      if (process.browser) {
+        if (localStorage.extendedMetadataIsOpen && JSON.parse(localStorage.extendedMetadataIsOpen)) {
```
lbiedinger: clicking on the button throws an error for me:
`JSON.parse: unexpected character at line 1 column 1 of the JSON data` in firefox
`Unexpected token u in JSON at position 0` in chrome

mirjamverloop: @lbiedinger I have added an additional check.

plugins/europeana/record.js
```diff
@@ -35,13 +35,54 @@ function parseRecordDataFromApiResponse(response) {
       dcPublisher: proxyData.dcPublisher,
       dcSubject: proxyData.dcSubject,
       dcType: proxyData.dcType,
-      dcTermsMedium: proxyData.dctermsMedium
+      dcTermsMedium: proxyData.dcTermsMedium
     }, isUndefined), entities),
     fields: lookupEntities(omitBy({
```
rwd: Let's move this out into its own function, and likewise `coreFields`.

lang/en.js
```diff
@@ -25,11 +25,51 @@ export default {
...
       "edmDataProvider": "Providing institution",
-      "edmRights": "License of the media in this record (unless otherwise specified)"
+      "edmRights": "License of the media in this record (unless otherwise specified)",
+      "dcRights": "Rights",
+      "dcDate": "Date",
+      "dcTermsPublished": "Publication Date",
+      "dcTermsIssued": "Issue date",
+      "dcTermsTemporal": "Temporal",
+      "dcCoverage": "Place-Time",
+      "dcTermsTOC": "Description",
```
rwd: This is not a valid field. dcTermsTableOfContents is, but that should have the label "Table of contents" and there are no records with data for it anyway, so perhaps just remove it from here.

lang/en.js
```diff
@@ -25,11 +25,51 @@ export default {
...
+      "dcTermsSpatial": "Places",
+      "edmCurrentLocation": "Current Location",
+      "edmUgc": "Source",
+      "dcTermsProvenance": "Provenance",
+      "dcSource": "Provenance",
```
rwd: This should be labelled "Source"

lang/en.js
```diff
@@ -25,11 +25,51 @@ export default {
...
+      "dcDuration": "Duration",
+      "dcMedium": "Medium",
+      "dcFormat": "Format",
+      "dcLanguage": "Language",
+      "dcTermsIsPartOf": "Is Part of",
+      "europeanaCollectionName": "Collection Name",
+      "dcRelation": "Relations",
+      "dcTermsReferences": "References",
+      "dcTermsHasPart": "Consists of",
+      "dcTermsHasVersion": "Version",
```
rwd: "Has version"

lang/en.js
```diff
@@ -74,7 +114,12 @@ export default {
   "depiction": "A depiction of {title}",
   "showLess": "Show less",
   "showMore": "Show more",
-  "record": "Record",
+  "record": {
+    "extendedInformation": "Extended Information",
```
rwd: "Extended information"

lang/en.js
```diff
@@ -25,11 +25,52 @@ export default {
...
+      "dcTermsIsReferencedBy": "Is Referenced by",
+      "dcTermsIsReplacedBy": "Is Replaced by",
+      "dcTermsIsRequiredBy": "Is Required by",
+      "edmHasMet": "Has Met",
+      "edmIncorporates": "Incorporates",
+      "edmIsDerivativeOf": "Is Derivative of",
+      "edmIsRepresentationOf": "Is Representation of",
+      "edmIsSimilarTo": "Is Similar to",
+      "edmIsSuccessorOf": "Is Successor of",
+      "proxies.edmRealizes": "Realises",
```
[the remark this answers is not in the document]
lbiedinger: no

plugins/europeana/record.js
```diff
@@ -19,37 +19,120 @@ function parseRecordDataFromApiResponse(response) {
...
+ * @param {Object[]} proxyData To take the fields from.
+ * @param {Object[]} edm To take additional fields from.
+ * @param {Object[]} entities Entities in order to perform entity lookups
+ * @return {Object[]} Key value pairs of the metadata fields.
+ */
+function extraFields(proxyData, edm, entities) {
+  const providerAggregation = edm.aggregations[0];
+  const europeanaAggregation = edm.europeanaAggregation;
+  return lookupEntities(omitBy({
+    dcTermsCreated: proxyData.dctermsCreated,
```
rwd: Why are this object's dcterms properties named like `dcTermsCreated` instead of `dctermsCreated`?

If there is no good reason, correct that for consistency. Likewise in `coreFields`.

plugins/europeana/entity.js
```diff
@@ -302,8 +302,9 @@ export function getEntityDescription(entity) {
  * @param {String} A URI to check
```
rwd: This should be:
```
* @param {string} string URI to check
```

Also, the new param should be documented.

plugins/europeana/record.js
lbiedinger: just from what the fields where before.

plugins/europeana/record.js
```diff
@@ -29,27 +29,110 @@ function parseRecordDataFromApiResponse(response) {
...
+/**
+ * Retrieves the "Core" fields which will always be displayed on record pages.
+ *
+ * @param {Object[]} proxyData All core fields are in the proxyData.
+ * @param {Object[]} entities Entities in order to perform entity lookups
+ * @return {Object[]} Key value pairs of the metadata fields.
+ */
+function coreFields(proxyData, entities) {
+  return lookupEntities(omitBy({
+    dcContributor: proxyData.dcContributor, // Plus rdaGr2DateOfBirth & rdaGr2DateOfDeath
```
rwd: What exactly does this comment and the same one on the following line mean? If irrelevant, remove them.

lbiedinger: On the Ruby portal we show additional information for the agents, it's out of scope for this PR.

PR #339, opened 2019-11-13
Rights statement in media action bar

mirjamverloop:


pages/record/_.vue
```diff
@@ -188,6 +189,15 @@
       isRichMedia() {
         return isRichMedia(this.selectedMedia.ebucoreHasMimeType, this.selectedMedia.edmCodecName, this.selectedMedia.about);
       },
+      rightsStatement() {
```
rwd: I'd be inclined to add a second computed property to DRY this up, like:
```js
      edmRights() {
        return this.selectedMedia.webResourceEdmRights ? this.selectedMedia.webResourceEdmRights : this.fields.edmRights;
      },
      rightsStatement() {
        if (this.edmRights) return langMapValueForLocale(this.edmRights, this.$i18n.locale).values[0];
        return false;
      },
```

PR #340, opened 2019-11-14
Move language selector to footer

mirjamverloop:


components/generic/LanguageSelector.vue
```diff
@@ -47,10 +45,7 @@
 </script>
 
 <style lang="scss" scoped>
-  @import "./assets/scss/variables.scss";
-
-  .dropdown {
-    margin-top: 20px;
+  /*.dropdown {
```
monkemedia: could this now be removed?

mirjamverloop: Yes, you are right, thanks!

tests/unit/components/PageFooter.spec.js
```diff
@@ -34,4 +34,15 @@ describe('components/search/PageFooter', () => {
...
+  it('contains the language selector', () => {
+    const wrapper = factory();
+    wrapper.setProps({
+      enableLanguageSelector: true
+    });
+    const selector = wrapper.find('[data-qa="language selector"]');
+
+    selector.isVisible().should.equal(true);
+  });
+
```
lbiedinger: this raises a linting error at the start of the unit tests
```
 Module Warning (from ./node_modules/@vue/cli-plugin-eslint/node_modules/eslint-loader/index.js):
  error: Trailing spaces not allowed (no-trailing-spaces) at tests/unit/components/PageFooter.spec.js:47:1:
    45 |     selector.isVisible().should.equal(true);
    46 |   });
  > 47 |   
       | ^
    48 | });
    49 | 

```

mirjamverloop: That has been removed, @lbiedinger

PR #346, opened 2019-11-15
Similar items component

monkemedia:


components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,38 @@
...
+    <b-col>
+      <b-card-group
+        class="card-deck-4-cols similar-items"
+        deck
+      >
+        <ContentCard
+          v-for="(item, index) in items"
+          :key="index"
+          :title="item.name"
+          :texts="texts"
```
rwd: Remove this.

components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,38 @@
...
+  <section class="row py-4 px-2 bg-white">
+    <b-col>
+      <b-card-group
+        class="card-deck-4-cols similar-items"
+        deck
+      >
+        <ContentCard
+          v-for="(item, index) in items"
+          :key="index"
+          :title="item.name"
```
rwd: As we are passing `title` down, let's make it `title` on the items too.

components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,38 @@
...
+
+  export default {
+    name: 'SimilarItems',
+
+    components: {
+      ContentCard
+    },
+
+    props: {
+      items: {
```
rwd: This needs a `linkGen` prop too. It should default to `(identifier) => identifier`, and it should be called on the ContentCard component iterator, like: `:url="linkGen(identifier)"`.

When we wire this all up, the parent component will pass in an actual link generation function.

components/generic/ContentCard.vue
```diff
@@ -76,6 +84,10 @@
       isRelated: {
         type: Boolean,
         default: false
+      },
+      thumbnail: {
```
rwd: Can we just reuse the existing `imageUrl` prop in here and also then get rid of the new `b-img` element? If so, let's.

components/generic/SimilarItems.vue
monkemedia: Sorry, not sure what you mean here

components/generic/ContentCard.vue
monkemedia: Do you think where possible we should be using images instead of background images?

components/generic/SimilarItems.vue
rwd: Change the property name on items to be `title`, i.e. `item.title`

components/generic/ContentCard.vue
rwd: Perhaps but let's consider that at a later date.

components/generic/SimilarItems.vue
rwd: After further consideration we decided just to hard-code the item page route in the component.

components/generic/SimilarItems.stories.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+storiesOf('Generic', module)
+  .addDecorator(StoryRouter())
+  .add('Similar items', () => ({
+    i18n,
+    components: { SimilarItems },
+    data() {
+      return {
+        items: [
```
rwd: Thumbnails for the items in this context will always be 200px wide but at the source image's original aspect ratio, so will rarely actually be square.

In the storybook therefore we need to use both portrait and landscape images to be sure they display as intended in this component.

Also, let's reuse the images we have in .storybook/assets/img/

components/generic/SimilarItems.stories.js
```diff
@@ -0,0 +1,48 @@
...
+
+const i18n = {
+  locale: 'en',
+  messages: {
+    en: {}
+  }
+};
+
+storiesOf('Generic', module)
+  .addDecorator(StoryRouter())
```
rwd: Pass this a named route to prevent console warnings:
```js
  .addDecorator(StoryRouter({}, {
    routes: [
      { name: 'record-all', path: '/record/*' }
    ]
  }))
```

components/generic/SimilarItems.stories.js
```diff
@@ -0,0 +1,48 @@
...
+  .add('Similar items', () => ({
+    i18n,
+    components: { SimilarItems },
+    data() {
+      return {
+        items: [
+          {
+            identifier: 'en/entity/person/60404-johannes-vermeer',
```
rwd: Replace the identifiers with some matching the actual format, e.g. '/123/abc'

components/generic/SimilarItems.stories.js
monkemedia: We can still control them so that they will always be a square. Maybe good to speak to the designer to see what they say?

rwd: Indeed, and that's what we need to do, and you have done here. :)

I was just noting that we want to reflect that in the Storybook by using images that are not in fact square, but display as such.

components/generic/ContentCard.vue
```diff
@@ -15,6 +15,7 @@
         :aria-label="title"
         :style="!isRelated && cardImageStyle"
         class="card-img"
+        :class="{ 'is-square' : isSquare }"
```
mirjamverloop: The similar items will always be square, that styling can be done based on its context, so I don't think we need this additional class.

components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <section class="row py-4 px-2 bg-white">
```
mirjamverloop: The similar items are going to be in a card. So could you remove the bg-white class and add the card class instead. 
`<section class="row py-4 px-2 card">`

components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,81 @@
...
+
+      @media (max-width: $bp-medium) {
+        flex: 100%;
+        margin-bottom: 1rem;
+      }
+    }
+
+    .card-link {
+      display: initial;
+    }
```
mirjamverloop: Could you add a hover on the link that makes the text blue?

components/generic/SimilarItems.vue
```diff
@@ -0,0 +1,81 @@
...
+
+    .card-img {
+      border: 0;
+      padding: 0;
+      border-radius: 0;
+      width: 100%;
```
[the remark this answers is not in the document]
monkemedia: Done

components/generic/ContentCard.vue
monkemedia: Yes, you are right, however, this would only work if we were actually using `` tags. Because we are using background images, the image will fill the container, causing the image not to appear square.

components/generic/SimilarItems.vue
monkemedia: done

components/generic/SimilarItems.vue
monkemedia: Done

components/generic/ContentCard.vue
mirjamverloop: I don't see what you mean. You can move the after to card-img and get rid of this additional is-square class in the similar items component, right?  
Eg: 

```
  .card-img {
      border-radius: 0;
     &:after {
        content: '';
        display: block;
        padding-bottom: 100%;
      }
    }
    
   // remove this
   // .is-square {
   // }
```

monkemedia: Apologies, I thought you mean to remove the class completely. I have applied your changes for you

PR #350, opened 2019-11-15
Favour Contentful entity page title & description

rwd:
On entity pages:
* Favour the title and description from Contentful
* If absent, use the values from the Entity API
* URL slugs also use the title from Contentful, but always in English
  * This necessitates making an additional request to the Contentful Delivery API to get the English title, if the user is using a different language. Contentful does not offer a way to request multiple locales, it's either just one or all.

### DIFF
--- a/pages/entity/_type/_.vue
+++ b/pages/entity/_type/_.vue
@@ -101,7 +101,15 @@
         return (!this.entity || !this.entity.depiction) ? null : entities.getWikimediaThumbnailUrl(this.entity.depiction.id);
       },
       description() {
-        return entities.getEntityDescription(this.entity);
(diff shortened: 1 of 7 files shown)


pages/entity/_type/_.vue
```diff
@@ -140,7 +153,7 @@
 
       const contentfulClient = createClient(query.mode);
 
-      return axios.all([
+      const requests = [
```
lbiedinger: I think this assignment is causing all the requests to be made twice, you can observe this by going to: http://localhost:3000/fr/entity/topic/1669-color-negatives and with the network inspect open clicking the first entity pill that links to the photograph page.

PR #355, opened 2019-11-19
Styling navigation and header

monkemedia:


components/PageHeader.vue
```diff
@@ -19,25 +19,28 @@
             data-qa="logo"
           >
         </b-navbar-brand>
-        <div class="navbar-nav ml-auto w-100 col-md-6 col-lg-6 p-0 pt-3 pt-md-0">
+        <div class="navbar-nav w-100 col-xl-4 p-0 pt-3 pt-md-0">
```
mirjamverloop: Does this "navbar-nav" class do anything here? It is a bit confusing as this is not the navigation.

PR #358, opened 2019-11-20
Entity depiction override

rwd:


pages/entity/_type/_.vue
```diff
@@ -98,14 +99,28 @@
...
       description() {
         return this.editorialDescription ? this.editorialDescription : entities.getEntityDescription(this.entity);
       },
+      editorialAttribution() {
+        return this.page.primaryImageOfPage['en-GB'].fields.url['en-GB'];
+      },
+      // Depiction from the Contentful entry
+      editorialDepiction() {
+        if (!this.page || !this.page.primaryImageOfPage) return null;
+        const image = this.page.primaryImageOfPage['en-GB'].fields.image['en-GB'].fields.file['en-GB'];
```
lbiedinger: While it's not possible to publish broken content, it can be broken in preview mode and also it's possible to delete a file(you'll get a warning, but can chose to ignore it) that is used in an image with attribution component. See this entity page and linked image with attribution, where I manually deleted the file in the test environment: http://localhost:3000/en/entity/topic/19-pamphlet
We should have a more graceful fallback, maybe just check the other values in this chain on the line above and return null when there's no file.

pages/entity/_type/_.vue
```diff
@@ -98,14 +99,34 @@
...
         return (!this.entity || !this.entity.depiction) ? null : entities.getWikimediaThumbnailUrl(this.entity.depiction.id);
       },
+      depictionLinkTitle() {
+        return this.editorialDepiction ? this.$t('goToRecord') : this.$t('entityDepictionCredit');
+      },
       description() {
         return this.editorialDescription ? this.editorialDescription : entities.getEntityDescription(this.entity);
       },
+      editorialAttribution() {
+        return this.page.primaryImageOfPage['en-GB'].fields.url['en-GB'];
```
monkemedia: Worth putting the 'en-GB' string into Vues Data Object? ie. `EN-GB: 'en-GB'`

rwd: They have now in fact been removed due to the reworking of the Contentful API requests to not include all locales in a single entry.

PR #367, opened 2019-11-21
Add hash to URL on selecting more facets

monkemedia:


rwd:
This is not quite the solution we need as it only works the first time one uses it in Chromium.

PR #372, opened 2019-11-22
Error pages

monkemedia:


layouts/error.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div
+    data-qa="error page"
+  >
+    <b-container>
+      <b-row>
+        <b-col>
+          <h1>Error {{ error.statusCode }}</h1>
```
rwd: While I take the point that the error messages are not translated, let's still translate "Error" here to save us having to do so later, and bearing in mind the rest of the UI is translated. Also, don't output the status code in the h1. Just "Error" (localised) will suffice.

layouts/error.vue
```diff
@@ -0,0 +1,27 @@
...
+          <p>{{ error.message }}</p>
```
rwd: Let's give this `class="lead"` as it deserves more prominence.

layouts/error.vue
```diff
@@ -0,0 +1,27 @@
...
+<script>
+  export default {
+    name: 'ErrorPage',
+
+    props: {
+      error: {
+        type: Object,
+        required: true
+      }
+    }
```
rwd: Could you add a head property with the title also set to "Error" (localised).

PR #384, opened 2019-11-27
Lokalise: Translations update

EuropeanaDevelopment:
This pull request was initiated by Lokalise (user Lutz Biedinger) at 2019-11-27 17:46:49


lang/en.js
```diff
@@ -21,14 +21,15 @@ export default {
...
       "dctermsCreated": "Creation date",
-      "dctermsMedium": "Medium",
       "edmCountry": "Providing country",
       "edmDataProvider": "Providing institution",
       "edmRights": "License of the media in this record (unless otherwise specified)",
-      "dcRights": "Rights",
-      "dcDate": "Date",
       "dcPublisher": "Publisher",
       "dcSubject": "Subject",
+      "dcTermsMedium": "Medium",
```
rwd: This needs to be removed from Lokalise.

lang/en.js
```diff
@@ -189,6 +180,30 @@ export default {
...
+    "extendedInformation": "Extended information",
+    "hideAll": "Hide all information",
+    "record": "Record",
+    "showAll": "Show all information",
+    "similarItems": "Similar items",
+    "view": {
+      "pdf": "View PDF"
+    }
+  },
+  "contentfulManual": {
```
rwd: What is this for?

lbiedinger: These were all strings that are not in the project, but Ash wanted to have them translated for the demo. So he got the translations done in lokalise and then copied them to contentful manually.

lang/en.js
lbiedinger: Done, but won't affect this PR, it'll be gone from the next one.

PR #388, opened 2019-11-29
Non-facet filters counted in more filters dropdown button

rwd:


components/search/SearchInterface.vue
```diff
@@ -1,4 +1,4 @@
-<template>
+defaultFacetNames<template>
```
lbiedinger: what's 'defaultFacetNames' doing here?

rwd: It's an unwanted guest and has been duly evicted.

PR #389, opened 2019-12-02
Date range input component

monkemedia:


components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
...
+      end: {
+        type: String,
+        default: ''
+      }
+    },
+
+    data() {
+      return {
+        form: {
+          name: this.name,
```
rwd: name will never change as a result of form input, so does not need to be a data attribute.

components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
...
+    name: 'DateFilter',
+
+    props: {
+      name: {
+        type: String,
+        required: true
+      },
+
+      start: {
+        type: String,
```
rwd: This should be type `Date`

components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
...
+        default: ''
+      },
+
+      end: {
+        type: String,
```
rwd: This should be type `Date`

components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <b-form inline>
+    <legend>
+      {{ name }}
```
rwd: name will be a field name like `proxy_dcterms_issued`. It needs to be made human-readable and localised by way of adding a computed method like:
```js
      facetName() {
        return this.$t(`facets.${this.name}.name`);
      }
```

components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
...
+          end: this.end
+        }
+      };
+    },
+
+    watch: {
+      form: {
+        deep: true,
+        handler() {
+          this.$emit('dateFilter', this.form);
```
rwd: This should emit name as a distinct parameter first, but the final parameter is fine to be an object with start and end properties.

I don't think it needs to go on watch either. It should be possible to just observe the `change` event on each date input and have that call a method which emits the required data.

components/search/DateFilter.stories.js
```diff
@@ -0,0 +1,38 @@
...
+      dateFilter: {
+        legend: 'Date range',
+        startDate: 'Start date',
+        endDate: 'End date'
+      }
+    }
+  }
+};
+
+storiesOf('Search', module)
```
rwd: Could you add an action method to this so that it observes the event emitted and logs to the Storybook UI the arguments sent to the event handler?

components/search/DateFilter.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <b-form inline>
```
rwd: This component will be embedded in a form already, rendered by an ancestral component, so does not need its own form element.

components/search/DateFilter.stories.js
```diff
@@ -0,0 +1,38 @@
...
+  .add('Date Filter', () => ({
+    i18n,
+    components: { DateFilter },
+    data() {
```
rwd: Is this necessary? Shouldn't the data function in the component handle this?

monkemedia: This is fake data pre-populate the form

components/search/DateFilter.vue
monkemedia: Native date picker only excepts strings

components/search/DateFilter.vue
monkemedia: Native date picker only excepts strings

components/search/DateFilter.stories.js
```diff
@@ -0,0 +1,50 @@
...
+  locale: 'en',
+  messages: {
+    en: {
+      dateFilter: {
+        startDate: 'Start date',
+        endDate: 'End date'
+      },
+      facets: {
+        'proxy_dcterms_issued': {
+          name: 'Example'
```
rwd: Let's make this 'Date Issued'

PR #391, opened 2019-12-02
Establish list of facets not supporting quotes

rwd:


components/search/SearchInterface.vue
```diff
@@ -312,7 +312,7 @@
             }
           } else {
             for (const facetValue of selectedValues) {
-              if (defaultFacetNames.includes(facetName)) {
+              if (this.enquoteFacet(facetName)) {
```
monkemedia: Would this be cleaner, as you are not repeating the push?

```
const quoted = this.enquoteFacet(facetName) ? "${facetValue}" : ${facetName}
queryUpdates.qf.push(`${facetName}:${quoted}`);
```

rwd: Indeed it would, and that's been done.

PR #392, opened 2019-12-02
Range queries

lbiedinger:


plugins/europeana/search.js
```diff
@@ -81,6 +81,30 @@ function fieldsForSearchResult(item) {
   return fields;
 }
 
+/**
+ * Construct a range query from two values, if keys are omitted they will default to '*'
+ * @param {Object[]} values An object containing 'start' and 'end' values
+ * @return {string} The range as a value that can be used by the API
+ */
+export function rangeToQueryParam(values) {
+  const start = values['start'] ? values['start'] : '*';
```
rwd: Use `values.start` and `values.end`

plugins/europeana/search.js
```diff
@@ -81,6 +81,30 @@ function fieldsForSearchResult(item) {
...
+}
+
+/**
+ * Deconstruct a range query string value into the upper and lower bounds.
+ * From/to values that are '*' will default to null.
+ * @param {string} paramValue The value as a string for a qf or query as used in the search API request
+ * @return {Object} Object with start and end keys
+ */
+export function rangeFromQueryParam(paramValue) {
+  const matches = paramValue.match(/^\[(.*) TO (.*)\]$/);
```
rwd: I think that the subpatterns should be `([^ ]+)` instead of `(.*)` as otherwise this will match `[ TO ]` which is _not_ a range query.

And could you add a unit test to ensure this case returns `null`.

PR #399, opened 2019-12-04
Accesibility for suggest dropdown

lbiedinger:


tests/features/common/language-selector.feature
```diff
@@ -14,6 +14,7 @@ Scenario: Cookie notice has been dismissed
 Scenario: Cookie notice has not been dismissed
   Given I am on the `English home page`
   And I accept cookies
+  And I wait 1 seconds
```
rwd: "second"

PR #400, opened 2019-12-04
Searchbox accessibility: remove duplicated button label

Abbe98:
Having both an image alt and a off screen element causes browser to label the main search button with both. Causing screen readers and other assistive tools to announce "search" twice.

Accessibility tree Firefox:
image


mirjamverloop:
Hi! Thank you for suggestion and the pull request. This has been handled elsewhere and I will, therefore, close the pull request. 

Have a good day!

PR #408, opened 2019-12-11
Specific date checkbox

mirjamverloop:


report.20191210.155319.4601.0.001.json
```diff
@@ -0,0 +1,807 @@
+
```
lbiedinger: this file shouldn't be in the PR

components/search/DateFilter.vue
```diff
@@ -107,12 +124,24 @@
...
+      },
+      specific() {
+        this.form.specific = this.specific;
       }
     },
 
     methods: {
       emitDateForm() {
-        this.$emit('dateFilter', this.name, this.form);
+        Vue.nextTick(() => { // Change event triggers before v-model has updated. This resolves the issue
```
rwd: The contents of this function can be reduced to simply:
```js
this.$emit('dateFilter', this.name, this.form);
```
... with the changes to the event handler in the parent component noted below.

components/search/MoreFiltersDropdown.vue
```diff
@@ -148,8 +153,18 @@
       },
 
       dateFilterSelected(facetName, dateRange) {
```
rwd: Change this function's logic to:
```js
        let dateQuery = [];

        if (dateRange.specific) {
          if (dateRange.start) {
            dateQuery = [dateRange.start];
          }
        } else if (dateRange.start || dateRange.end) {
          dateQuery = [rangeToQueryParam(dateRange)];
        }

        this.isCheckedSpecificDate = dateRange.specific;

        this.updateSelected(facetName, dateQuery);
```
It's a bit more concise and supports the above changes to emitDateForm.

components/search/FacetFieldLabel.vue
```diff
@@ -53,6 +61,17 @@
         let subtype = this.fieldValue.split('/')[1];
 
         return subtype.replace(/^x-/, '').toUpperCase();
+      },
+
+      dateLabel() {
```
rwd: This is redundant and can be removed.

PR #409, opened 2019-12-12
Output data provider name, linking to edm:isShownAt

monkemedia:


components/record/MediaActionBar.vue
```diff
@@ -67,3 +99,18 @@
...
+<style lang="scss" scoped>
+  @import '../../assets/scss/icons.scss';
+
+  .view-at {
+    .is-external-link {
+      &:after {
+        content: '\e900';
+        @extend .icon-font;
+        margin-left: 4px;
+        display: inline-block
```
rwd: Add semi-colon.

components/record/MediaActionBar.vue
```diff
@@ -67,3 +99,18 @@
...
+
+<style lang="scss" scoped>
+  @import '../../assets/scss/icons.scss';
+
+  .view-at {
+    .is-external-link {
+      &:after {
+        content: '\e900';
+        @extend .icon-font;
+        margin-left: 4px;
```
rwd: Can we use rem units here?

lang/en.js
```diff
@@ -359,7 +359,9 @@ export default {
     "exhibitions": "Exhibitions"
   },
   "actions": {
-    "download": "Download"
+    "download": "Download",
+    "viewAt": "View at",
```
rwd: It can not be assumed that in all languages the provider name and link will go at the end of this text, so this needs a {provider} placeholder to take the link.

components/record/MediaActionBar.vue
```diff
@@ -1,35 +1,55 @@
...
+      </b-col>
+      <b-col
+        cols="6"
+        class="d-flex align-items-center justify-content-end"
+      >
+        <span
+          data-qa="provider name"
+          class="is-size-4 view-at"
+        >
+          {{ isShownAt ? $t('actions.viewAt') : $t('actions.providedBy', { provider: providerName }) }}
```
rwd: When there is no link, the provider name still needs to be marked up with the language code in the lang attribute, on a span element.

pages/record/_.vue
```diff
@@ -257,6 +260,10 @@
       rightsStatement() {
         if (this.edmRights) return langMapValueForLocale(this.edmRights, this.$i18n.locale).values[0];
         return false;
+      },
+      dataProvider() {
+        const edmDataProvider = this.coreFields.edmDataProvider;
+        return langMapValueForLocale(edmDataProvider);
```
rwd: This needs to be passed the user's locale, i.e.
```js
        return langMapValueForLocale(edmDataProvider, this.$i18n.locale);

```

components/record/MediaActionBar.vue
```diff
@@ -57,6 +77,18 @@
       rightsStatement: {
         type: String,
         default: null
+      },
+      providerName: {
```
rwd: Let's name this dataProviderName as provider and data provider are two distinct properties of records.

components/record/MediaActionBar.vue
```diff
@@ -57,6 +77,18 @@
...
+        type: String,
+        default: null
+      },
+      lang: {
```
rwd: This should be renamed to be more specific, e.g. `dataProviderLang`

PR #411, opened 2019-12-12
Return empty array when no description texts found

mirjamverloop:


components/entity/EntityCards.vue
```diff
@@ -6,7 +6,7 @@
       :key="entity.id"
       :is-related="true"
       :title="entity.prefLabel.en"
-      :texts="getEntityDescription(entity, $i18n.locale).values"
+      :texts="getEntityDescription(entity, $i18n.locale) ? getEntityDescription(entity, $i18n.locale).values : []"
```
rwd: Let's make a computed property `hasDescription` to test whether there is anything to render.

rwd: Actually I have refactored a different way in the following commit in order to prevent repeating the function call.

PR #418, opened 2019-12-18
Localise all entity API metadata

monkemedia:


components/browse/BrowseChip.vue
```diff
@@ -3,26 +3,34 @@
...
       title: {
-        type: String,
-        default: ''
+        type: Object,
+        default: () => {}
+      }
+    },
+
+    computed: {
+      browseChipTitle() {
```
rwd: Let's name this `localisedTitle`

pages/entity/_type/_.vue
```diff
@@ -172,9 +173,9 @@
         };
       },
       title() {
-        if (!this.entity) return this.$t('entity');
-        if (this.editorialTitle) return this.editorialTitle;
-        return this.entity.prefLabel[this.$store.state.i18n.locale];
+        if (!this.entity) return this.titleFallback(this.$t('entity'));
```
rwd: This will always be in the user's UI language, so does not need the lang attribute.

pages/entity/_type/_.vue
```diff
@@ -172,9 +173,9 @@
...
+        if (this.editorialTitle) return this.titleFallback(this.editorialTitle);
```
rwd: This may be in the user's UI language or may be in English, but we do not know which from the Contentful response, so should not have the lang attribute.

pages/entity/_type/_.vue
```diff
@@ -292,12 +293,21 @@
       this.$store.commit('search/setPill', this.title);
     },
 
+    methods: {
+      titleFallback(title) {
```
rwd: Given the two comments above, this is not needed.

pages/entity/_type/_.vue
```diff
@@ -292,12 +293,21 @@
...
+        return {
+          values: [title],
+          code: this.$i18n.locale
```
rwd: This should be `null`

monkemedia: done

components/browse/BrowseChip.vue
monkemedia: done

PR #424, opened 2019-12-20
Social media links in footer

monkemedia:


components/generic/LinkGroup.vue
```diff
@@ -21,7 +21,17 @@
           :destination="link.url"
           :link-class="linkClass"
         >
-          {{ link.text }}
+          <template v-if="link.icon">
+            <span class="icon-background">
+              <i
```
rwd: Is `i` the right element to use here?

components/generic/LinkGroup.vue
```diff
@@ -21,7 +21,17 @@
...
+                :class="link.icon"
+                :aria-label="link.text"
```
rwd: Should this be on the `title` attribute instead so that it's visible on hover?

components/generic/LinkGroup.vue
mirjamverloop: An `i` is commonly used for icons, a `span` would be appropriate too. I don't have a preference.

components/generic/LinkGroup.vue
mirjamverloop: I have changed this into a title.

PR #436, opened 2020-01-08
Update styling of related cards

mirjamverloop:
- Updated styling of the related/entity cards
- Automated entity cards are now styles as related/entity cards
- Renamed the CSS "related-card" to "entity-card" and prop "isRelated" to "isEntity".

### DIFF
--- a/components/browse/BrowseContentCard.vue
+++ b/components/browse/BrowseContentCard.vue
@@ -24,8 +24,8 @@
         default: () => {}
       },
       cardType: {
-        type: Object,
-        default: () => {}
+        type: String,
+        default: null
       }
     },
     computed: {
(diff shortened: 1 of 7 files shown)


components/browse/BrowseContentCard.vue
```diff
@@ -21,6 +22,10 @@
       fields: {
         type: Object,
         default: () => {}
+      },
+      cardType: {
+        type: Object,
```
rwd: Let's make this a string and have the parent component be responsible for drilling down into the object to extract it.

components/browse/BrowseContentCard.vue
```diff
@@ -69,6 +74,13 @@
           }
         }
         return texts;
+      },
+      isEntity() {
+        if (this.cardType && this.cardType.sys.id === 'automatedEntityCard') {
```
rwd: Just return the value of the conditional and remove all following lines from this function.

PR #440, opened 2020-01-09
View media new tab

mirjamverloop:


components/record/MediaPresentation.vue
```diff
@@ -134,7 +121,14 @@
           this.oEmbedData = { error: err };
         });
       }
+    },
+
+    methods: {
+      imageLink(pdf) {
+        return pdf ? this.media.about : this.$options.filters.proxyMedia(this.media.about, this.europeanaIdentifier) + '&disposition=inline';
```
rwd: Update the proxyMedia filter to accept arbitrary additional parameters as an optional third param, like:
```js
Vue.filter('proxyMedia', (mediaUrl, europeanaId, params = {}) => {
```

Then pass disposition to the filter in this call.

components/record/MediaPresentation.vue
```diff
@@ -134,7 +127,14 @@
           this.oEmbedData = { error: err };
         });
       }
+    },
+
+    methods: {
+      imageLink() {
```
rwd: As this accepts no parameters it can be a computed property.

components/record/MediaPresentation.vue
```diff
@@ -4,19 +4,10 @@
   >
     <MediaImage
       v-if="displayImage"
-      :link="imageLink"
       :src="imageSrc"
+      :link="imageLink()"
```
rwd: If/when imageLink becomes a computed property (see below), the parentheses can be removed.

plugins/vue-filters.js
```diff
@@ -48,9 +48,14 @@ Vue.filter('convertNewLine', (val) => {
   return val.replace(/\n/g, '<br/>');
 });
 
-Vue.filter('proxyMedia', (mediaUrl, europeanaId) => {
+Vue.filter('proxyMedia', (mediaUrl, europeanaId, params = {}) => {
   const proxyUrl = new URL('https://proxy.europeana.eu');
   proxyUrl.pathname = europeanaId;
   proxyUrl.searchParams.append('view', mediaUrl);
+
+  Object.keys(params).forEach(name => {
```
rwd: Let's use the native for...in:
```js
  for (const name in params) {
    proxyUrl.searchParams.append(name, params[name]);
  }
```

rwd: This functional change should be unit tested.

plugins/media.js
```diff
@@ -4,6 +4,13 @@ export function isPDF(media) {
   return media.ebucoreHasMimeType === 'application/pdf';
 }
 
+export function isImage(media) {
```
rwd: Unit test this.

components/record/MediaPresentation.vue
```diff
@@ -97,8 +84,14 @@
           iiif: Number(process.env.ENABLE_IIIF_MEDIA)
         });
       },
-      isPDF() {
-        return isPDF(this.media);
+      imageMediaType() {
```
rwd: I don't think it makes sense to do this here just to pass it into MediaImage.vue

Let's refactor so that:
1. MediaImage.vue takes the same props as MediaPresentation.vue, i.e. imageSrc, media and europeanaIdentifier
2. imageLink is moved into MediaImage.vue
3. isPDF and isImage are imported in MediaImage.vue, not in MediaPresentation.vue, and used to determine the i18n key directly instead of setting this additional media type property

PR #450, opened 2020-01-15
Legacy entity page URL redirects

rwd:


middleware/legacy.js
```diff
@@ -1,31 +1,46 @@
...
-    const redirectPath = stringifyPathMatch(legacyRecordPageMatch);
-    return redirect({
-      path: redirectPath,
-      query: route.query
-    });
+  (routePath) => {
+    const legacyRecordPagePattern = /^(\/[a-z]{2})?(\/record\/[0-9]+\/[a-zA-Z0-9_]+)\.html$/;
+    const legacyRecordPageMatch = routePath.match(legacyRecordPagePattern);
+    return legacyRecordPageMatch ? stringifyPathChunks(legacyRecordPageMatch.slice(1)) : null;
+  },
```
lbiedinger: why not have this remove '.html' from the end of _any_ path? 
```const legacyHTMLPattern = /^(.*)?.html$/;```

In conjunction with the removal of /portal, record pages will then be in the right format, and as all the checks are ran anyway this will simplify the path for future checks.

PR #455, opened 2020-01-16
Lokalise: Translations update

EuropeanaDevelopment:
This pull request was initiated by Lokalise (user Lutz Biedinger) at 2020-01-16 15:22:09


lang/en.js
```diff
@@ -82,15 +82,16 @@ export default {
...
-        "permission": "Permission",
+        "open": "Yes",
+        "restricted": "Yes, with conditions",
+        "permission": "Maybe, seek permission",
         "uncategorized": "Uncategorized"
       }
     },
     "THEME": {
-      "name": "Collection",
+      "name": "Category",
```
rwd: Is this intentional?

lang/en.js
```diff
@@ -82,15 +82,16 @@ export default {
...
       "options": {
+        "all": "All items",
```
rwd: This is not used and should be removed.

lang/en.js
```diff
@@ -394,7 +393,7 @@ export default {
   },
   "contentfulManual": {
```
rwd: Is this ever used? I see no reference to it in the codebase.

lang/en.js
lbiedinger: I've removed the key from lokalize.

lang/en.js
lbiedinger: yes

lang/en.js
rwd: For reference, it's a place for translations in Lokalise which will then be input manually into Contentful, for now.

PR #457, opened 2020-01-17
Redirect search results

rwd:
* Refactor `legacy` middleware to be more modular
* Introduce module for redirecting search results URLs with different query parameter naming


middleware/legacy/rules/search.js
```diff
@@ -0,0 +1,133 @@
...
+        queryFacetParameterMappings.default(query, field, value[field]);
+      }
+    }
+  },
+  range(query, key, value) {
+    for (const field in value) {
+      const rangeBegin = value[field]['begin'];
+      const rangeEnd = value[field]['end'];
+      if (rangeBegin === rangeEnd) {
+        query.qf.push(`${field}:${rangeBegin}`);
```
lbiedinger: these need to account for empty strings and add '*'

middleware/legacy/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+  for (const rule of rules) {
+    const redirectRoute = rule(route, query);
+
+    if (redirectRoute) {
+      if (Array.isArray(redirectRoute.path)) redirectRoute.path = stringifyPathChunks(redirectRoute.path);
+      if (!redirectRoute.query && query) redirectRoute.query = query;
+
+      // TODO: instead of returning here, should we keep looping over other rules
+      //       so multiple rules get applied all at once?
```
lbiedinger: - [x] yes

components/search/SearchInterface.vue
```diff
@@ -210,6 +210,7 @@
       api() {
         return this.userParams.api;
       },
+      // TODO: do not assume filters are fielded, e.g. `qf=whale`
```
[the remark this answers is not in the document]
rwd: Yes. If one adds that to the URL, then it does not behave as expected, nor get displayed as expected.

middleware/legacy/rules/search.js
rwd: Indeed. That's done, with unit tests.

middleware/legacy/index.js
rwd: Actually, with the latest versions of the rules, I think it wouldn't work, as the response from the first match won't include /portal. I've removed the comment instead.

PR #458, opened 2020-01-20
Attributed image url validation, Exhibition publication date

lbiedinger:
For EC-4151 & EC-4149

Add publication date to exhibitions, allow blank URLs in Image with Attribution

### DIFF
--- a/migrations/0015-exhibition-publication-and-attributed-image-validation.js
+++ b/migrations/0015-exhibition-publication-and-attributed-image-validation.js
@@ -1,4 +1,16 @@
 module.exports = function(migration) {
+  const exhibitionPage = migration.editContentType('exhibitionPage');
+
+  exhibitionPage
+    .createField('datePublished')
+    .name('Publish at')
+    .type('Date')
+    .localized(false)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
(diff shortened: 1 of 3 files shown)


migrations/0015-exhibition-publication-and-attributed-image-validation.js
```diff
@@ -0,0 +1,32 @@
...
+    .type('Date')
+    .localized(false)
+    .required(true)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  const imageWithAttribution = migration.editContentType('imageWithAttribution');
+
+  imageWithAttribution
```
rwd: This is awkward to scan for the changes applied as it has the full field definition. Just run the commands that result in actual changes to the field.

migrations/0015-exhibition-publication-and-attributed-image-validation.js
```diff
@@ -0,0 +1,32 @@
+module.exports = function(migration) {
+  const exhibitionPage = migration.editContentType('exhibitionPage');
+
+  exhibitionPage
```
rwd: This will put the new field at the end of the ones that already exist. I suggest moving it to immediately after `primaryImageOfPage` to be more consistent with the blog post content type.

PR #459, opened 2020-01-20
Request additional Fashion facets

lbiedinger:


lang/en.js
```diff
@@ -82,6 +82,18 @@ export default {
     "COUNTRY": {
       "name": "Providing country"
     },
+    "proxy_dc_format": {
+      "name": "Technique"
+    },
+    //"proxy_dc_format": {
```
rwd: Remove this. We have decided this is out of scope for now.

store/collections/fashion.js
```diff
@@ -1,6 +1,6 @@
 import { defaultFacetNames } from '../search';
 
-const fashionFacetNames = ['CREATOR'].concat(defaultFacetNames);
+const fashionFacetNames = ['CREATOR', 'proxy_dc_format', 'proxy_dcterms_medium', 'proxy_dc_type'].concat(defaultFacetNames);
```
rwd: Shouldn't these new facets all be `.en`? They are on Portal.rb

lbiedinger: yes, this is causing some issues with the labelling I'll need to look into though.

lbiedinger: I've  pushed changes to escape the translation keys. This is only applied to the more facet dropdown elements. I've not applied the same convention of facet name -> translation key modifications on date facets or the standard dropdowns, as they won't contain these kind of keys for now.

An alternative approach to this may be additional nesting in the locale file, but this would move the translations to different depth depending on the actual facet name and seems inconsistent to me.

PR #460, opened 2020-01-21
Theme-specific facet labels

rwd:


plugins/vue/i18n.js
```diff
@@ -0,0 +1,34 @@
+// I18n function to localise into active locale if translation exists, else
+// into app fallback locale, else return null.
+function translateOrNull(key, values) {
```
lbiedinger: Where is values being passed in here? What does it do?

rwd: It's to mimic the signatures of the functions from Vue I18n which may or may not have placeholder values. Included here for the purpose of future-proofing.

PR #461, opened 2020-01-22
Shallow mount search components without Bootstrap

rwd:


tests/unit/components/search/SearchForm.spec.js
```diff
@@ -22,9 +20,10 @@ const router = new VueRouter({
   ]
 });
 const routerPush = sinon.spy(router, 'push');
-const factory = (options = {}) => mount(SearchForm, {
+const factory = (options = {}) => shallowMount(SearchForm, {
   localVue,
   router,
+  stubs: ['b-button', 'b-input-group', 'b-button', 'b-form', 'b-form-input'],
```
mirjamverloop: `b-button` is in here twice

rwd: Removed.

PR #462, opened 2020-01-22
Filter fashion facet value by designer suffix

mirjamverloop:


tests/unit/store/collections/fashion.spec.js
```diff
@@ -4,12 +4,32 @@ describe('store/collections/fashion', () => {
...
+      ];
+
       it('includes CREATOR in facet param', () => {
         store.getters.apiParams(state).facet.should.startWith('CREATOR,');
       });
+
+      it('filters the Fashion facets', () => {
+        const beforeFilter = payload[1][0].fields.length;
+        store.mutations.filter(state, payload);
+        state.facets[0].fields.length.should.not.eq(beforeFilter);
```
rwd: Let's test what it should actually be, i.e. 2.

store/collections/fashion.js
```diff
@@ -20,6 +21,18 @@ export const mutations = {
   enable(state) {
     state.enabled = true;
   },
+  filter(state, payload) {
```
rwd: Unless it's envisaged this will ever operate on other fields, which I can't anticipate at present, then it's not necessary or useful to send through an array as payload. Just sends the facets, then set `state.facets` directly below.

store/collections/fashion.js
```diff
@@ -20,6 +21,18 @@ export const mutations = {
...
+    const facets = payload[1];
+    facets.forEach((facet, index) => {
+      facets[index] = facet;
+
+      if (facet.name === 'CREATOR') {
+        facets[index]['fields'] = facets[index].fields.filter(creator => creator.label.indexOf('(Designer)') !== -1);
```
rwd: Use `String.prototype.endsWith()`, just for readability.

store/collections/fashion.js
```diff
@@ -20,6 +21,18 @@ export const mutations = {
   enable(state) {
     state.enabled = true;
   },
+  filter(state, payload) {
+    const facets = payload[1];
+    facets.forEach((facet, index) => {
+      facets[index] = facet;
```
rwd: What's this doing?

store/collections/fashion.js
rwd: Also, let's rename it to `filterFacets`.

store/collections/fashion.js
mirjamverloop: A leftover, so I have removed it.

PR #468, opened 2020-01-24
More static page redirects

rwd:


middleware/legacy/rules/static-pages.js
```diff
@@ -3,34 +3,52 @@
...
-  '/collections/newspapers': { path: '/entity/topic/18-newspaper' },
-  '/collections/photography': { path: '/entity/topic/48-photography' },
-  '/collections/sport': { path: '/entity/topic/114-sport' },
-  '/exhibitions/foyer': { path: '/exhibitions' },
-  '/explore/galleries': { path: '/galleries' }
+  '/collections/industrial-heritage': '/entity/topic/129-industrial-heritage',
+  '/collections/manuscripts': '/entity/topic/17-manuscripts',
+  '/collections/maps': '/entity/topic/151-maps-and-geography',
+  '/collections/migration': '/entity/topic/128-migration',
+  '/collections/migration/collection-days-stories.html': '/migration/explore-migration-stories-from-events-in-these-cities',
```
lbiedinger: this doesn't work. Probably due to having '.html' in here.

middleware/legacy/rules/static-pages.js
```diff
@@ -3,34 +3,52 @@
...
+  '/collections/migration/collection-days': '/europeana-migration-collection-days',
```
lbiedinger: this one works, but only when visiting: '[host]/portal/en/collections/migration/explore-migration-stories', realistically people would be coming from 'https://www.europeana.eu/portal/en/collections/migration/explore-migration-stories.html' which icludes 'html'

PR #470, opened 2020-01-27
Store resettable filters

rwd:
And:
* clear all collection-specific filters when switching collections
* supply Fashion facet field labels


components/search/SearchInterface.vue
```diff
@@ -350,16 +351,17 @@
...
         for (const name in selected) {
           filters[name] = selected[name];
+        }
 
-          // Remove collection-specific filters when collection is changed
-          // TODO: this is rather crude and will not scale well; improve
-          if (name === 'THEME' && !selected['api']) {
-            filters['api'] = null;
+        // Remove collection-specific filters when collection is changed
+        if (selected[this.THEME]) {
```
lbiedinger: This isn't catching the case where the theme is reset. Meaning you can end up on a page like this:
http://localhost:3000/en/search?page=1&qf=proxy_dc_format.en%3A%22Technique%3A%20embroidery%22&query=&view=grid

PR #471, opened 2020-01-28
Next/previous chapter buttons

mirjamverloop:
- Previous/next buttons on the chapter pages
- Styled
- Unit and feature tests


components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,90 @@
...
+        const nextIndex = this.currentChapterIndex() + 1;
+        return this.getChapter(nextIndex);
+      },
+      previousChapter() {
+        const prevIndex = this.currentChapterIndex() - 1;
+        return this.getChapter(prevIndex);
+      }
+    },
+    methods: {
+      currentChapterIndex() {
```
rwd: As this accepts no arguments, it could be a computed property.

components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,90 @@
...
+        if (this.chapters.length <= 1) return false;
+        return this.chapters.findIndex(chapter => chapter.fields.identifier === this.currentChapter);
+      },
+      getChapter(index) {
+        return this.chapters[index] ?
+          { 'name': this.chapters[index].fields.name, 'url': this.chapterUrl(this.chapters[index]) } : null;
```
rwd: The properties of this object can just be symbols rather than strings.

components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,90 @@
...
+</template>
+
+<script>
+  export default {
+    props: {
+      exhibitionIdentifier: {
+        type: String,
+        required: true
+      },
+      chapters: {
```
rwd: As this component is concerned only with navigation, this property should already be an array of objects having only `name` and `url` properties, assembled by the parent component.

pages/exhibition/_exhibition/_chapter.vue
```diff
@@ -54,15 +58,24 @@
...
   export default {
     components: {
       BrowseSections,
       ExhibitionChapters,
+      ExhibitionChaptersNavigation,
       HeroImage
     },
     computed: {
+      chapterNavigation() {
+        let navigation = [];
```
rwd: Use `map` here like:
```js
        return this.chapters.map((chapter) => {
          return {
            identifier: chapter.fields.identifier, name: chapter.fields.name, url: this.chapterUrl(chapter.fields.identifier)
          };
        });
```

components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,89 @@
...
+      }
+    },
+    data() {
+      return {
+        currentChapter: this.$route.params.chapter
+      };
+    },
+    computed: {
+      currentChapterIndex() {
+        if (this.chapterNavigation.length <= 1) return false;
```
rwd: This does not do what it is intended to, because `false + 1 // => 1` and `false - 1 // => -1`.

Remove this line and nothing breaks, even on exhibitions with only a single chapter.

components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,89 @@
...
+        return this.getChapter(nextIndex);
+      },
+      previousChapter() {
+        const prevIndex = this.currentChapterIndex - 1;
+        return this.getChapter(prevIndex);
+      }
+    },
+    methods: {
+      getChapter(index) {
+        return this.chapterNavigation[index] ? this.chapterNavigation[index] : null;
```
rwd: This conditional is a bit redundant as if there is no such index it'll return `undefined` anyway which has the same effect in the `v-if` conditions. Just return `this.chapterNavigation[index]`.

components/exhibition/ExhibitionChaptersNavigation.vue
```diff
@@ -0,0 +1,88 @@
...
+        return this.getChapter(nextIndex);
+      },
+      previousChapter() {
+        const prevIndex = this.currentChapterIndex - 1;
+        return this.getChapter(prevIndex);
+      }
+    },
+    methods: {
+      getChapter(index) {
+        return this.chapterNavigation[index];
```
rwd: This method now seems like overkill given its simplicity. Just have the computed methods return the relevant chapter themselves.

tests/unit/components/exhibition/ExhibitionChaptersNavigation.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  identifier: 'astilbe',
+  url: 'en/exhibition/flowers/astilbe'
+},
+{
+  name: 'Calls for education',
+  identifier: 'calls-for-education',
+  url: 'en/exhibition/flowers/calls-for-education'
+}];
+
+const factory = () => mount(ExhibitionChaptersNavigation, {
```
rwd: Use `shallowMount` here.

tests/unit/components/exhibition/ExhibitionChaptersNavigation.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    $route,
+    localePath: (opts) => opts
+  },
+  propsData: {
+    exhibitionIdentifier: 'flowers',
+    chapterNavigation
+  }
+});
+
+describe('components/exhibition/ExhibitionChaptersNavigation', () => {
```
rwd: There should only be one of these `describe` blocks, not three.

tests/unit/components/exhibition/ExhibitionChaptersNavigation.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  it('shows a previous and next chapter button', () => {
```
rwd: Add `context` blocks to explain when each of these expectations should occur, e.g.
```js
context('when there are both previous and next chapters', () => {
  it('has buttons for both', () => {
```

PR #479, opened 2020-02-04
Optimised images in exhibition cards

mirjamverloop:


components/generic/ContentCard.vue
```diff
@@ -90,6 +90,10 @@
         type: String,
         default: null
       },
+      imageMaxDimensions: {
+        type: Object,
+        default: () => {}
+      },
```
lbiedinger: As this takes the place of the options param when it's passed to optimisedImageUrl it would make sense to have this named imageOptimisationOptions.
That way it's more descriptive should we ever make use of additional options other than just `width, height`

PR #480, opened 2020-02-05
Share a record

mirjamverloop:
- Component for social share buttons, 
- Toggle in the action bar,
- og:title, og:type, og:image  and og:url on record pages

### DIFF
--- a/components/generic/SocialShare.vue
+++ b/components/generic/SocialShare.vue
@@ -1,33 +1,42 @@
 <template>
-  <p>
+  <div>
     <b-link
-      title="Share on Instagram"
-    >
-      <span class="icon-instagram" />
-    </b-link>
-    <b-link
-      title="Share on Facebook"
+      target="_blank"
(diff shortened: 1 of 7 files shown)


pages/record/_.vue
```diff
@@ -317,6 +326,8 @@
         }
       });
 
+      this.canonicalURL = window.location.href;
```
lbiedinger: This will be too lenient for when a user adds arbitrary stuff to the URL, like for example:
http://localhost:3000/en/record/2024909/photography_ProvidedCHO_United_Archives_02035211_jpg?param_that_does=nothing&another=one#user-supplied-tag
For now remove any params and anchor tags.

components/generic/SocialShare.vue
```diff
@@ -57,7 +57,7 @@
     },
 
     mounted() {
-      this.location = window.location.href;
+      this.location = window.location.href.split(/\?|#/)[0];
```
lbiedinger: let's pass this in as a prop from the component/page.

PR #483, opened 2020-02-07
Gallery Markdown descriptions

lbiedinger:


plugins/vue-filters.js
```diff
@@ -38,12 +40,38 @@ Vue.filter('optimisedImageUrl', (imageUrl, contentType, options = {}) => {
...
+ * This method FIRST converts markdown to HTML, then removes HTML tags.
+ * WARNING: This also means any HTML tags already present before will be stripped.
+ *
+ * As an optional parameter specific HTML tag names can be supplied. In which case,
+ * only these will be removed.
+ * @param {string} text String containing mark1down
+ * @param {string[]} tags the HTML tags to be removed.
+ * @return {String} text value with HTML breaks
+ */
+Vue.filter('stripMarkdown', (text, tags) => {
```
rwd: This is more concise and DRY:
```js
Vue.filter('stripMarkdown', (text, tags = ['']) => {
  text = marked(text); // Marked adds newlines to the end of the string, and wraps it in a <p> tag.
  for (const pattern of htmlRemovalPatternsFromTags(tags)) {
    text = text.replace(pattern, '');
  }
  return text;
});
```

PR #486, opened 2020-02-11
Gallery sharing

lbiedinger:


pages/gallery/_.vue
```diff
@@ -54,21 +65,35 @@
...
+    },
     head() {
       return {
         title: this.title,
         meta: [
           { hid: 'title', name: 'title', content: this.title },
           { hid: 'description', name: 'description', content: this.description },
           { hid: 'og:title', property: 'og:title', content: this.title },
-          { hid: 'og:description', property: 'og:description', content: this.description }
+          { hid: 'og:description', property: 'og:description', content: this.description },
```
mirjamverloop: If there is no description, the description is "undefined". Let's leave it blank instead.

pages/gallery/_.vue
```diff
@@ -54,21 +65,35 @@
...
           error({ statusCode: 500, message: e.toString() });
         });
     },
+    computed: {
+      shareMediaUrl() {
+        if (this.images.length <= 0) return null;
+        if (!this.images[0].fields.thumbnailUrl) return null;
+        return this.images[0].fields.thumbnailUrl;
+      }
+    },
```
mirjamverloop: The "computed" should go before the asyncData.

pages/gallery/_.vue
```diff
@@ -54,22 +74,30 @@
           return {
             description: response.items[0].fields.description,
             images: response.items[0].fields.hasPart,
-            title: response.items[0].fields.name
+            title: response.items[0].fields.name,
+            canonicalURL: null
```
rwd: The naming of this is inconsistent with that of `shareMediaUrl`. Rename it to `canonicalUrl`.

pages/gallery/_.vue
```diff
@@ -54,22 +74,30 @@
...
           };
         })
         .catch((e) => {
           error({ statusCode: 500, message: e.toString() });
         });
     },
+    mounted() {
+      this.canonicalURL = window.location.href.split(/\?|#/)[0];
```
rwd: This results in the og:url meta tag having the value `null` when requested solely by SSR which may impede indexing. Move it into the `asyncData` method, using `route` from the Nuxt context instead of relying on the CSR-only window location.

PR #488, opened 2020-02-13
Refactor share for SSR support

lbiedinger:


pages/gallery/_.vue
```diff
@@ -93,8 +89,7 @@
           { hid: 'description', name: 'description', content: this.description },
           { hid: 'og:title', property: 'og:title', content: this.title },
           { hid: 'og:image', property: 'og:image', content: this.shareMediaUrl },
-          { hid: 'og:type', property: 'og:type', content: 'article' },
-          { hid: 'og:url', property: 'og:url', content: this.canonicalURL }
```
mirjamverloop: This page now does not have a og:url. Is that correct?
There should be one, at least, according to https://ogp.me/#metadata .

pages/record/_.vue
```diff
@@ -80,7 +80,7 @@
         <div class="card p-3 mb-3 bg-grey">
           <MediaActionBar
             :url="selectedMedia.about"
-            :share-url="canonicalURL"
+            :share-url="canonicalUrl"
```
mirjamverloop: Is this - and the other references in this file - still valid ? The SocialShare component no longer accepts shareUrl as a prop.

pages/gallery/_.vue
lbiedinger: it does, but it doesn't update. This is still in progress

lbiedinger:
Prefer  #503

PR #493, opened 2020-02-17
Optimise images in attributed and comparisons

rwd:


components/generic/CompareImageSlider.vue
```diff
@@ -108,6 +135,7 @@
 
     data() {
       return {
+        maxWidth: 1100,
```
mirjamverloop: Where is this used?

components/generic/CompareImageSlider.vue
```diff
@@ -150,7 +178,8 @@
 
     methods: {
       setImageWidth() {
-        this.imageWidth = this.$refs.rightImage.getBoundingClientRect().width;
+        console.log(this.$refs.rightImage);
```
mirjamverloop: Remove the console.log.

components/generic/OptimisedImage.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <span>
```
mirjamverloop: Is the span necessary? It breaks the image compare. If possible remove it here, or else, the image compare code should be slightly updated.  

```diff --git a/components/generic/CompareImageSlider.vue b/components/generic/CompareImageSlider.vue
index 77221612..a0eda8d4 100644
--- a/components/generic/CompareImageSlider.vue
+++ b/components/generic/CompareImageSlider.vue
@@ -246,7 +246,7 @@
     position: relative;
     overflow: hidden;
 
-    img {
+    span {
       width: 100%;
 
       &:nth-child(1) {
```

components/generic/OptimisedImage.vue
```diff
@@ -0,0 +1,100 @@
...
+    <b-img-lazy
+      v-if="lazy"
+      :src="optimisedSrc"
+      :width="optimisedWidth"
+      :height="optimisedHeight"
```
mirjamverloop: Could you set this to blank-width and blank-height, so CSS can still do some resizing if needed.

components/generic/CompareImageSlider.vue
rwd: Nowhere.

components/generic/OptimisedImage.vue
rwd: It seems not. I will remove it.

PR #520, opened 2020-02-26
Notification banner for home page

mirjamverloop:


lang/en.js
```diff
@@ -468,5 +468,11 @@ export default {
   },
   "galleries": {
     "galleries": "Galleries"
+  },
+  "linksToClassic": {
+    "home": {
+      "text": "You're viewing the new Europeana experience.",
+      "linkText": "Go to the original Europeana"
```
rwd: This needs a full stop at the end.

pages/index.vue
```diff
@@ -53,6 +62,12 @@
       },
       heroImage() {
         return this.hero ? this.hero.image.fields.file : null;
+      },
+      onHomePage() {
+        return Boolean(Number(process.env.ENABLE_LINKS_TO_CLASSIC)) && !this.path;
+      },
+      notificationUrl() {
+        return `https://classic.europeana.eu/${this.$store.state.i18n.locale}/portal/?utm_source=new-website&utm_medium=button`;
```
rwd: This is incorrect. The locale comes after /portal, and there should not be another slash before `?`

PR #521, opened 2020-02-26
List view

mirjamverloop:
- Description in list view
- Refactor of ContentCard, now using variant (current options: default, list, mini, entity)
- Styling update of the list view
- Removal of unused code/components/styling


components/search/SearchResults.vue
```diff
@@ -69,9 +67,14 @@
     },
 
     methods: {
-      cardTexts(result) {
+      cardTexts(result, variant) {
         const texts = [result.edmDataProvider];
         if (result.dcCreator) texts.unshift(result.dcCreator);
+
+        if (variant === 'list') {
+          if (result.dcDescription) texts.unshift(result.dcDescription);
```
rwd: This needs to only be added if dcTitle is present, else it will end up duplicated from being used instead of the title.

To see this, search for `proxy_dc_description:* NOT proxy_dc_title:*`

PR #524, opened 2020-02-27
Wait for page load in feature tests

rwd:
Changes:
* In feature tests, instead of having steps to "wait X seconds", wait for the actual page to load, based on display and disappearance of Nuxt progress bar.
  * Speeds up the tests where they do not take X seconds, taking one minute off Travis runs.
  * Makes the tests more resilient by giving them up to 5 seconds for each such page load, but only taking as long as they need.
* Disable lazy loading of images in some components when running in the test env as they often cause issues, e.g. when attempting to paginate through search results and webdriver reaches the location where the pagination was, then lazyloaded images are injected and the pagination is moved by the time webdriver attempts to use it.
* Switch the "entity page" in feature tests to be a non-thematic-collection one, so that it is much faster to test against.
* Disable the auto suggest e2e tests as the feature is disabled by default and has been for some time with no ETA on restoration.

Not changed:
* There are some Newspapers collection feature tests failing, which will be addressed elsewhere.


tests/features/search/querying.feature
```diff
@@ -31,43 +31,51 @@ Feature: Search querying
...
     And I click a `search result`
     Then I see a `record page`
     And I don't see "paris" in the `search box`
 
-  Scenario: Using auto suggestion with keyboard populates search field
-    When I visit a `search page`
+  Scenario: No auto suggestion
+    Given I am on the `search page`
     And I enter "World" in the `search box`
     And I wait 1 second
```
lbiedinger: let's up this to 2 seconds. It's passing even though autosuggest is switched on, just because it's taking longer than a second to come in.

PR #525, opened 2020-02-28
Search redirects

lbiedinger:


pages/search/index.vue
```diff
@@ -1,23 +1,34 @@
...
-      <b-col>
-        <h1>{{ $t('search') }}</h1>
-      </b-col>
-      <SearchInterface
-        :per-row="4"
-      />
-    </b-row>
-  </b-container>
+  <div>
+    <NotificationBanner
```
rwd: This should only be displayed when the feature toggle `ENABLE_LINKS_TO_CLASSIC` is enabled, as with on the homepage already. That is to enable us to trivially switch it off without needing to make code-level changes & deployments.

PR #536, opened 2020-03-04
Disable download button on InC media

rwd:


assets/scss/buttons-links.scss
```diff
@@ -13,7 +13,7 @@
   font-size: 0.875rem;
   text-transform: uppercase;
 
-  &:hover {
+  &:hover:not([disabled]) {
```
lbiedinger: after this block add the following css:
```

  &:disabled {
    color: $grey;
    border-color: $grey;
  }
```
to differentiate the disabled buttons a bit more, there's not much difference between the shades of blue.

PR #540, opened 2020-03-06
Add triangle shape to hero and footer

mirjamverloop:


assets/scss/hero.scss
```diff
@@ -60,4 +80,7 @@
       font-size: 1rem;
     }
   }
+
+
+
```
lbiedinger: get rid of these extra spaces.

PR #572, opened 2020-03-23
Add label to cards

mirjamverloop:
Options: exhibition, gallery, blog

### DIFF
--- a/components/generic/ContentCard.vue
+++ b/components/generic/ContentCard.vue
@@ -147,7 +147,8 @@
     },
     data() {
       return {
-        cardImageUrl: this.imageUrl
+        cardImageUrl: this.imageUrl,
+        displayLabelTypes: 'exhibitions|galleries|blog'
       };
     },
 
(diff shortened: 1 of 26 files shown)


components/generic/ContentCard.vue
```diff
@@ -160,6 +167,15 @@
         }
       },
 
+      displayLabel() {
+        const regex = /(exhibitions|galleries|blog)/g;
```
rwd: This needs to be a bit stricter to prevent false positives. If we have a blog post about exhibitions with a slug like /blog/2020-online-exhibitions-for-the-self-isolating, the regex will detect it as an exhibition.

lang/en.js
```diff
@@ -368,6 +368,7 @@ export default {
   "newWindow": "opens in new window",
   "blog": {
     "blog": "Blog",
+    "label": "Blog",
```
rwd: Instead of adding new keys, we should use the pluralisation format, as we have for the facet filter labels like "Colour | Colours"

PR #583, opened 2020-04-01
Update card component to use encoding field

mirjamverloop:


components/browse/BrowseContentCard.vue
```diff
@@ -34,24 +35,33 @@
...
         apiConfig: 'apis/config'
       }),
+      cardFields() {
+        return this.cardType === 'automatedRecordCard' ? this.fields.encoding : this.fields;
+      },
       title() {
-        return this.fields.name;
+        if (this.cardType === 'automatedRecordCard') {
+          return this.cardFields.dcTitleLangAware;
+        }
```
lbiedinger, line 44: This won't be backwards compatible, do a check as above where it uses dcTitleLangAware if encoding is set.

components/browse/BrowseContentCard.vue
```diff
@@ -70,9 +80,17 @@
...
         // TODO: Refactor content model to set this directly, so this method can be skipped.
         let texts = [];
-        for (const field of ['description', 'creator', 'provider']) {
-          if (this.fields[field]) {
-            texts.push(this.fields[field]);
+        let textFields;
+
+        if (this.cardType === 'automatedRecordCard') {
+          textFields = ['dcDescriptionLangAware', 'dcCreatorLangAware', 'dataProvider'];
+        } else {
```
lbiedinger, line 87: Only do this when the encoding is set.

PR #589, opened 2020-04-07
Styling heading/description gallery pages

mirjamverloop:


components/generic/PageHeader.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <header class="row">
+    <b-col
+      cols="12"
+      :lg="mediaUrl ? 9 : 12"
```
mirjamverloop, line 5: Could you always set this to 9. The title and description should always stretch 9 columns.

components/generic/PageHeader.vue
```diff
@@ -0,0 +1,65 @@
...
+      </h1>
+      <!-- eslint-disable vue/no-v-html -->
+      <div
+        v-if="htmlDescription"
+        v-html="htmlDescription"
+      />
+      <!-- eslint-enable vue/no-v-html -->
+      <p v-if="rawDescription">
+        {{ rawDescription }}
+      </p>
```
mirjamverloop, line 21: Does it break when you have just 1 description and use only

`<div
        v-if="htmlDescription"
        v-html="htmlDescription"
      />`
?

monkemedia: We end up with a missing <p> tag if we just use htmldescription.

components/generic/PageHeader.vue
monkemedia: Sorry, I forgot to push my changes

components/generic/ContentHeader.vue
```diff
@@ -0,0 +1,65 @@
...
+      </h1>
+      <!-- eslint-disable vue/no-v-html -->
+      <div
+        v-if="htmlDescription"
+        v-html="htmlDescription"
+      />
+      <!-- eslint-enable vue/no-v-html -->
+      <p v-if="rawDescription">
+        {{ rawDescription }}
+      </p>
```
lbiedinger, line 21: get rid of this and just pass 'description' in as a prop to simplify the logic here.
Passing plain text through the v-html attribute will still work with plain text.

Then in the various Pages always only pass in 'description'

pages/index.vue
```diff
@@ -22,20 +22,11 @@
...
-          <p
-            v-if="page.headline"
-            class="lead"
-          >
-            {{ page.headline }}
-          </p>
-        </div>
-      </header>
+        :title="page.name"
+        :raw-description="page.headline"
```
lbiedinger, line 28: this needs to be 'description' too

monkemedia: resolved

PR #591, opened 2020-04-09
Accessibility: keyboard navigation

monkemedia:


components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,21 @@
     class="d-flex flex-wrap mb-3"
   >
     <!-- TODO: populate alt, but with what? -->
-    <b-img-lazy
+    <button
```
mirjamverloop, line 7: I do think it is valid HTML, an `` inside a button, but could you make this a link? See also the notes on the description of the ticket.

components/item/MediaActionBar.vue
```diff
@@ -82,7 +82,7 @@
       </b-col>
     </b-row>
     <b-row
-      v-else
+      v-show="showShareButtons"
```
mirjamverloop, line 85: Could you update me in why you have changed this into a v-show?

components/generic/CookieDisclaimer.vue
```diff
@@ -118,6 +118,9 @@
       justify-content: center;
       display: flex;
       cursor: pointer;
+      background-color: transparent;
+      color: $white;
+      border: none;
```
mirjamverloop, line 123: Maybe it is worth adding a class to the buttons-links.scss for a "btn-transparent" with the transparent background-color and the border set to none. So it can be used elsewhere, for example in the MediaActionBar.

monkemedia: Done

components/item/MediaActionBar.vue
monkemedia: v-if removes the element from the DOM, so when it does become finally visible, the tabbing index messes up, stopping the user from tabbing through the 3 social buttons. Using v-show, keeps the element in the DOM, but hidden from the user. Maybe we should encourage this more throughout, as I am sure its probably better for screen readers too.

components/item/MediaThumbnailGrid.vue
monkemedia: Now using anchor tags instead of buttons

components/item/MediaActionBar.vue
```diff
@@ -197,7 +197,8 @@
   }
 
   .icon-close {
-    cursor: pointer;
+    background-color: transparent;
+    border: 0;
   }
```
lbiedinger, line 202: prefer the new `btn-transparent` class to get this styling on here.

components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,23 @@
...
       :data-about="thumbnail.about"
-      :data-qa="`media thumbnail #${index + 1}`"
-      thumbnail
-      alt=""
-      class="mb-2 mr-2 rounded-0"
-      @click.native="clickThumbnail(thumbnail.about)"
-    />
+      :data-qa="`link thumbnail #${index + 1}`"
+      href="#"
+      class="thumbnail-link mb-2 mr-2 rounded-"
```
mirjamverloop, line 14: What is the `rounded-` class doing?

components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,23 @@
...
+      @click="clickThumbnail(thumbnail.about)"
+    >
+      <b-img-lazy
+        :src="thumbnail.src"
+        :data-qa="`media thumbnail #${index + 1}`"
+        thumbnail
+        :alt="$t('mediaPreview.selectMedia', { src: thumbnail.src })"
```
mirjamverloop, line 21: I am not sure if this is the best alt text. This could mean that a screen reader must read `Select media https://api.europeana.eu/api/v2/thumbnail-by-url.json?size=w200&type=IMAGE&uri=http%3A%2F%2Fwww.mimo-international.com%2Fmedia%2FGNM%2FIMAGE%2FMINe99_1309160825120_2.jpg`. 

I know that we left the text empty before, because we didn't find the best text yet. An alternative could be "Select 2nd (3th, 4th, etc) ... "

monkemedia: @lbiedinger what are your thoughts on this, as I followed your advice by using the path?

lbiedinger: Yes, we wanted to have this in here for now as to at least pass the axe tests, but do a follow up.
One thing we were talking about is that it would be useful to know the 'format' of the media item, but doing so was pushing this work well out of scope for the original ticket here.

components/item/MediaThumbnailGrid.vue
monkemedia: Will remove now

components/item/MediaThumbnailGrid.vue
monkemedia: Are you happy to merge this @lbiedinger ?

components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,23 @@
...
       :class="thumbnailImgClass(thumbnail, index)"
       :data-about="thumbnail.about"
-      :data-qa="`media thumbnail #${index + 1}`"
-      thumbnail
-      alt=""
-      class="mb-2 mr-2 rounded-0"
-      @click.native="clickThumbnail(thumbnail.about)"
-    />
+      :data-qa="`link thumbnail #${index + 1}`"
+      href="#"
```
rwd, line 13: href is not a required attribute. As this anchor element does not link anywhere, remove the href attribute.

components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,23 @@
...
-      :src="thumbnail.src"
       :class="thumbnailImgClass(thumbnail, index)"
       :data-about="thumbnail.about"
-      :data-qa="`media thumbnail #${index + 1}`"
-      thumbnail
-      alt=""
-      class="mb-2 mr-2 rounded-0"
-      @click.native="clickThumbnail(thumbnail.about)"
-    />
+      :data-qa="`link thumbnail #${index + 1}`"
```
rwd, line 12: This should something like `"media thumbnail anchor #1"` instead of `"link thumbnail #1"`

components/item/MediaThumbnailGrid.vue
```diff
@@ -4,18 +4,23 @@
...
+      href="#"
+      class="thumbnail-link mb-2 mr-2"
+      @click="clickThumbnail(thumbnail.about)"
+    >
+      <b-img-lazy
+        :src="thumbnail.src"
+        :data-qa="`media thumbnail #${index + 1}`"
+        thumbnail
+        :alt="$t('mediaPreview.selectItem', { src: index + 1 })"
```
rwd, line 21: This is poor alternate text for the _content_ of the image. Instead:
* Stick to using `alt=""`
* Move this label content up to the anchor element on an appropriate aria-x attribute

PR #593, opened 2020-04-15
Updated active navigation state with blue underline

alansutherland:


components/PageNavigation.vue
```diff
@@ -66,6 +66,17 @@
 
       &.nuxt-link-active {
         font-weight: bold;
+        &:after {
+          content: '';
+          position: absolute;
+          border-bottom: solid 3px #0771ce;
```
mirjamverloop, line 72: Would you mind updating the colour of the border? You can use the variable $blue.

components/PageNavigation.vue
```diff
@@ -66,6 +66,17 @@
 
       &.nuxt-link-active {
         font-weight: bold;
```
mirjamverloop, line 68: Could you remove the font weigh to match the design?

PR #594, opened 2020-04-15
Accessibility: focus

monkemedia:


components/search/MoreFiltersDropdownFacet.vue
```diff
@@ -111,6 +113,17 @@
...
           this.$emit('selectedOptions', this.name, this.selectedOptions);
         });
+      },
+
+      toggleShowMore() {
+        this.isActive = !this.isActive;
+
+        this.$nextTick(() => {
+          let index;
+          this.isActive ? index = this.limitTo : index = this.limitTo - 1;
```
rwd, line 123: Replace these two lines with:
```js
const index = this.limitTo - (this.isActive ? 0 : 1);
```
This has lower cognitive complexity, and `index` is never mutated so should use `const`.

PR #597, opened 2020-04-16
Initial above 992 style added

alansutherland:
ready for review

### DIFF
--- a/assets/scss/form.scss
+++ b/assets/scss/form.scss
@@ -2,7 +2,7 @@
   background-color: $offwhite;
   border: 0;
   font-weight: normal;
-  height: 2.5rem;
+  height: 2.625rem;
 
   &::placeholder {
     color: $grey;
--- a/assets/scss/form.scss
+++ b/assets/scss/form.scss
@@ -2,7 +2,7 @@
   background-color: $offwhite;
   border: 0;
   font-weight: normal;
-  height: 2.625rem;
+  height: 2.5rem;
 
   &::placeholder {
     color: $grey;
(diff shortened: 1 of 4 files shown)


alansutherland:
not ready for review yet, still some styles to add

assets/scss/form.scss
```diff
@@ -2,7 +2,7 @@
   background-color: $offwhite;
   border: 0;
   font-weight: normal;
-  height: 2.5rem;
+  height: 2.625rem;
```
mirjamverloop, line 5: Just a detail, but I think this makes the search box 2 pixels too tall, compared to the design.

components/PageNavigation.vue
```diff
@@ -63,6 +63,7 @@
       color: $mediumgrey;
       text-decoration: none;
       text-transform: uppercase;
+      font-size: $font-size-extrasmall;
```
mirjamverloop, line 66: Could you use `$font-size-small` instead? That is the size used in the design.

assets/scss/form.scss
alansutherland: @mirjamverloop I reverted to this style since reducing the height/padding of the nav was causing the tiny gap between the active link and the bottom border to show again

alansutherland: ok I've reduced the height and added margin to keep the blue underlines in the right position

components/PageNavigation.vue
alansutherland: sure I've changed to small

PR #598, opened 2020-04-20
Newspaper full text preview shows highlighted search term

monkemedia:


plugins/europeana/search.js
```diff
@@ -75,15 +75,22 @@ export function rangeFromQueryParam(paramValue) {
...
+
     return {
       europeanaId: item.id,
       edmPreview: item.edmPreview ? `${item.edmPreview[0]}&size=w200` : genericThumbnail(item.id, { type: item.type, size: 'w200' }),
       dcTitle: item.dcTitleLangAware,
       dcDescription: item.dcDescriptionLangAware,
       dcCreator: item.dcCreatorLangAware,
-      edmDataProvider: item.dataProvider
+      edmDataProvider: item.dataProvider,
+      selector: (selector && selector.selectors) ? selector.selectors : []
```
lbiedinger, line 93: I think something like this will be better:
```
/**
 * Extract search results from API response
 * @param  {Object} response API response
 * @return {Object[]} search results
 */
function resultsFromApiResponse(response) {
  const items = response.data.items;

  const results = items.map(item => {
    return {
      ...{
        europeanaId: item.id,
        edmPreview: item.edmPreview ? `${item.edmPreview[0]}&size=w200` : genericThumbnail(item.id, {
          type: item.type,
          size: 'w200'
        }),
        dcTitle: item.dcTitleLangAware,
        dcDescription: item.dcDescriptionLangAware,
        dcCreator: item.dcCreatorLangAware,
        edmDataProvider: item.dataProvider
      },
      ...(response.data.hits !== undefined ? hitForItem(response.data.hits, item.id) : {})
    };
  });

  return results;
}

function hitForItem(hits, id) {
  const selector = hits.find((hit) => id === hit.scope);
  return selector ? { selector: selector.selectors[0]  } : {};
}
```
Then in components/search/SearchResults.vue, we can just have this:
```
-          if (result.selector.length === 0 && result.dcDescription) texts.unshift(result.dcDescription);
+          if (!result.selector && result.dcDescription) texts.unshift(result.dcDescription);
```
 
And from a data point of view only items with hits will actually pass a selector object along.

components/generic/ContentCard.vue
```diff
@@ -52,6 +52,17 @@
         >
           {{ $d(new Date(datetime), 'short') }}
         </time>
+        <template v-if="hitsText.length > 0">
+          <template v-for="(hit, index) in hitsText">
```
lbiedinger, line 56: this outside loop can be removed with the changes proposed for how hits are added to each item.
Also we only ever have one 'hit' so I think 'hitText' makes more sense as a variable.

components/generic/ContentCard.vue
```diff
@@ -96,6 +107,10 @@
         type: Array,
         default: () => []
       },
+      hitsText: {
+        type: Array,
```
lbiedinger, line 111: This will have to be an object and should default to null if the changes to the search plugin are implemented.

plugins/europeana/search.js
monkemedia: Don't you think this is harder to read? Also, it seems like we are adding more code. I am happy to pass an object instead of an array though.

lbiedinger: It's more consistent with what we do elsewhere where objects are constructed.
And the `hitForItem` function is more atomic, being removed from the `resultsFromApiResponse` function.

But the main reason is to not have the key set to an empty value, but rather not added at all.
The other way to achieve this would be to remove any keys with blank values, but again the approach of appending the additional key like this is how we do it elsewhere.

components/generic/ContentCard.vue
```diff
@@ -52,6 +52,14 @@
         >
           {{ $d(new Date(datetime), 'short') }}
         </time>
+        <template v-if="hitsText">
+          <b-card-text
+            text-tag="div"
+            data-qa="highlighted search term"
+          >
+            <p>{{ hitsText.prefix }} <span class="font-weight-bold has-text-highlight">{{ hitsText.exact }}</span> {{ hitsText.suffix }}</p>
```
rwd, line 60: Let's use strong instead of span here, for better semantic markup and accessibility.

PR #602, opened 2020-04-22
Header split in to header/search/nav landmarks

alansutherland:


components/PageHeader.vue
```diff
@@ -1,32 +1,34 @@
...
+      </SmartLink>
+    </header>
+    <SearchForm
+      data-qa="search form"
+      role="search"
+      class="col-lg w-100 px-3 mr-auto mx-xl-auto"
+      :enable-auto-suggest="enableAutoSuggest"
+      :enable-suggestion-validation="enableSuggestionValidation"
+    />
+    <b-navbar
```
mirjamverloop, line 29: Axe is reporting an issue with the navbar "Landmarks must have a unique role or role/label/title (i.e. accessible name) combination" and is suggesting "The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable".

components/PageHeader.vue
```diff
@@ -1,32 +1,34 @@
...
-    class="border-bottom p-0 mb-3"
+    class="border-bottom mb-3 d-flex py-3 py-lg-1 px-lg-3 flex-column flex-lg-row"
+    data-qa="header"
   >
-    <b-navbar
-      class="pb-3 pt-3 py-lg-1 px-lg-3 flex-column flex-lg-row"
-      data-qa="header"
+    <header
+      class="col-lg-2 p-0 m-0 text-center text-lg-left navbar-brand d-flex align-items-center justify-content-center justify-content-lg-start"
+      role="banner"
```
mirjamverloop, line 9: I think the search bar can go into the landmark "banner" too. At least, that it was I get from the documentation https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_banner.

pages/exhibitions/index.vue
```diff
@@ -1,13 +1,13 @@
 <template>
   <b-container data-qa="exhibitions">
-    <header class="row">
+    <div class="row">
```
mirjamverloop, line 3: What is the reason for changing header into div? On this page and the other pages? I think it is semantically correct to use a `<header>` here.

components/PageHeader.vue
alansutherland: ok I have added aria-labels for the landmarks

components/PageHeader.vue
alansutherland: sure I will put the search in the banner, just thought all landmarks had to be top-level but checked the docs and search doesn't have to be

pages/exhibitions/index.vue
alansutherland: I'll revert this change, I thought this might break rules for only having one header landmark but it is semantically correct

PR #605, opened 2020-04-24
Hreflang is now applied via the layouts template

monkemedia:


layouts/default.vue
```diff
@@ -70,15 +70,31 @@
...
-          { hid: 'og:url', property: 'og:url', content: this.canonicalUrl }
+          { hid: 'og:url', property: 'og:url', content: this.canonicalUrl },
+          ...i18nSeo.meta
         ]
       };
+    },
+
+    methods: {
+      i18nLinksWithDefault(i18nSeo) {
+        const defaultISO =  this.$i18n.isoLocale(this.$i18n.defaultLocale);
```
lbiedinger, line 93: this still uses the current locale without updating the code in plugins/i18n.js
```
export default ({ app }) => {
  const locales = app.i18n.locales;
  // Set the function directly on the context.app object
  app.i18n.isoLocale = function(code = app.i18n.locale) {
    return locales.find(locale => locale.code === code)['iso'];
  };
};
```

layouts/default.vue
```diff
@@ -70,19 +70,31 @@
...
     head() {
+      const i18nSeo = this.$nuxtI18nSeo();
       return {
         link: [
-          { rel: 'stylesheet', href: 'https://fonts.googleapis.com/css?family=Ubuntu:300,400,700%7COpen+Sans:400italic,700italic,400,600,700&subset=latin,greek,cyrillic&display=swap', body: true }
+          { rel: 'stylesheet', href: 'https://fonts.googleapis.com/css?family=Ubuntu:300,400,700%7COpen+Sans:400italic,700italic,400,600,700&subset=latin,greek,cyrillic&display=swap', body: true },
+          ...this.i18nLinksWithDefault(i18nSeo)
         ],
+        htmlAttrs: {
+          myAttribute: 'My Value',
```
lbiedinger, line 80: remove this

nuxt.config.js
```diff
@@ -43,7 +43,7 @@ module.exports = {
   /*
   ** Global CSS
   */
-  css: [],
+  css: ['./assets/scss/style'],
```
lbiedinger, line 46: what does this do?

monkemedia: This was in layouts/default.vue. It is the overall styling for the application. Sadly it was causing issues when running tests. Unit tests could not find the font icons.. a bit strange.

To fix the issue, I placed the styling here, where it should technically be I guess.

nuxt.config.js
```diff
@@ -120,10 +120,10 @@ module.exports = {
     'cookie-universal-nuxt',
     ['nuxt-i18n', {
       locales: i18nLocales,
-      baseUrl: process.env.I18N_BASE_URL,
+      baseUrl: ({ store }) => store.getters['http/canonicalUrlWithoutLocale'],
```
rwd, line 123: This needs to use the getter `http/origin`. Otherwise it results in hrefs like:
```html
<link data-n-head="ssr" data-hid="alternate-hreflang-bg" rel="alternate" href="http://localhost:3001/search?page=1&amp;view=grid&amp;query=/bg/search?page=1&amp;view=grid&amp;query=" hreflang="bg">
```

plugins/i18n.js
```diff
@@ -5,7 +5,7 @@
 export default ({ app }) => {
   const locales = app.i18n.locales;
   // Set the function directly on the context.app object
-  app.i18n.isoLocale = function() {
-    return locales.find(locale => locale.code === app.i18n.locale)['iso'];
+  app.i18n.isoLocale = function(code = app.i18n.locale) {
```
rwd, line 8: What does this change do, i.e. is it used anywhere?

monkemedia: Lutz wanted me to add it, @lbiedinger can you update @rwd on this?

lbiedinger: this was needed for when we wanted to link to the 'default' locale. It's not used elsewhere now.

monkemedia: I shall revert it back

PR #606, opened 2020-04-27
don't use aria-label or aria-labelledby on a span or div

alansutherland:


alansutherland:
@mirjamverloop not sure if .slider-bar on /CompareImageSlider.vue needs aria-label https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html
added labels to MoreFiltersDropdown & FacetDropdown but this threw errors during testing so removed

PR #619, opened 2020-05-01
Cards containing people and are only 4 in length now show as mini cards

monkemedia:


components/browse/ContentCardSection.vue
```diff
@@ -56,7 +56,27 @@
     },
     computed: {
       cards() {
-        return this.section.fields.hasPart.filter(card => card.fields);
+        const cards = this.section.fields.hasPart.filter(card => card.fields);
+        return cards.map(card => {
+          const identifier = card.fields.identifier;
+          const id = card.sys.contentType.sys.id;
+          if (id === 'automatedEntityCard' && identifier.match(/base/) && cards.length === 4) {
```
lbiedinger, line 63: matching /base/ isn't right here. We only want to match agent/person entities.

I reckon you'll want to include entityParamsFromUri from plugins/europeana/entity.js and check that the returned type attribute === `person`. Alternatively copy the regex from that function, but modify it to only match /agent/ URIs to begin with. That might be more efficient, but I'm  not sure if that modified regex should live in this file here or deserves it's own helper function in the entity plugin.

Additionally I don't think this check should be happening in a mapping of cards. As this check of content length === 4 isn't checking that there are actually 4 people cards, it just checks that there are 4 cards, people or otherwise.
I'd suggest adding a new computed property `isPeopleSection` which only returns true/false if all 4 elements of the object returned by cards() are people cards.

Then rather than having additional logic in the browse content card, make the inclusion of the BrowseContentCard conditional based on that boolean. If it's true, instead include the ContentCards component directly here without the BrowseContentCard wrapper. All the browsContentCard does is normalize the fields. 
As we know the people cards will all be of the type automatedEntityCard, just set the ContentCard props from the known required values, like:
```
  <ContentCard
    v-if="isPeopleSection"
    v-for="card in cards"
    :title="card.fields.name"
    ...
    variant="mini"
  />
```
That'll also mean you won't have to do the modification for omitting the description as you just won't be passing it to the card component in the first place.

tests/unit/components/browse/ContentCardSection.spec.js
```diff
@@ -80,5 +80,39 @@ describe('components/browse/ContentCardSection', () => {
       moreButton.text().should.contain('Show more art');
       moreButton.attributes('href').should.eq('http://europeana.eu');
     });
+
+    it('displays mini cards if a section is exclusively people', () => {
```
lbiedinger, line 84: let's add another test where one of the entity is _not_ an agent, but a concept. Make sure       `wrapper.vm.isPeopleSection.should.equal(false);`

components/browse/ContentCardSection.vue
```diff
@@ -57,6 +73,17 @@
     computed: {
       cards() {
         return this.section.fields.hasPart.filter(card => card.fields);
+      },
+
+      isPeopleSection() {
```
lbiedinger, line 78: this can be made a bit more concise:
```
      isPeopleSection() {
        if (this.cards.length !== 4) return false;
        return this.cards.every((card) => {
          return card.fields.identifier ? entityParamsFromUri(card.fields.identifier).type === 'person' : false;
        });
      }
```

monkemedia: I have applied your changes, but decided to keep the identifier variable in, otherwise i would have repeated `card.fields.identifier` twice.

components/browse/ContentCardSection.vue
```diff
@@ -20,12 +20,25 @@
...
+            :fields="card.fields"
+            :card-type="card.sys.contentType ? card.sys.contentType.sys.id : ''"
+          />
+        </template>
+        <template v-else>
+          <ContentCard
+            v-for="card in cards"
+            :key="card.sys.id"
+            :title="card.fields.name"
+            :url="card.fields.identifier"
```
mirjamverloop, line 36: The cards here link to the identifier itself, for example, http://data.europeana.eu/agent/base/147134. It does a couple of redirects and you finally end up on the production entity. 

Could you use a correct router link here? So it performs well and so you stay on the same domain.

PR #620, opened 2020-05-01
Stickynav desktop

alansutherland:


components/PageHeader.vue
```diff
@@ -58,6 +75,69 @@
         type: Boolean,
         default: false
       }
+    },
+
+    data() {
+      return {
+        scrollPosition: 0,
+        windowWidth: 992,
+        desktopWidth: 992
```
lbiedinger, line 84: Is there a better way to get this from our css breakpoint in case we change the specific value?

components/PageHeader.vue
```diff
@@ -58,6 +75,69 @@
         type: Boolean,
         default: false
       }
+    },
+
+    data() {
+      return {
+        scrollPosition: 0,
+        windowWidth: 992,
```
lbiedinger, line 83: What's the reason for this being set to 992? Could it be null or 0?

alansutherland: it could be 0/null but then handleScroll function would not run on load.
windowWidth is set dynamically after mounted() runs, so it may be fine without it, just thought it was safer this way

components/PageHeader.vue
alansutherland: it should remain the same value as long as we keep using bootstrap since that's the standard desktop breakpoint

alansutherland: we can export the value from sass variables 
```
:export{
    $bplarge: 992px;
}
```
then import
`import { $bplarge } from '../assets/scss/variables.scss';`
but for some reason this value isn't actually picked up by the app when setting:
`desktopWidth: $bplarge`

I think it's very unlikely the value will change often, 992 is the default desktop breakpoint for bootstrap so hardcoding isn't a problem

components/PageHeader.vue
alansutherland: have tested setting to 0 and since this.getWindowWidth(); is called within mounted there's no issue

PR #627, opened 2020-05-06
POC integrate media embed

mirjamverloop:


plugins/media.js
```diff
@@ -20,6 +20,12 @@ export function isHTMLAudio(media) {
   return ['audio/flac', 'audio/ogg', 'audio/mpeg'].includes(media.ebucoreHasMimeType);
 }
 
+export function isPlayableMedia(media) {
+  return (media.ebucoreHasMimeType.startsWith('video/') ||
+    media.ebucoreHasMimeType.startsWith('audio/') ||
+    (media.ebucoreHasMimeType === 'application/dash+xml'));
```
rwd, line 26: Do we even have any such content? If not, remove this.

plugins/media.js
```diff
@@ -20,6 +20,12 @@ export function isHTMLAudio(media) {
   return ['audio/flac', 'audio/ogg', 'audio/mpeg'].includes(media.ebucoreHasMimeType);
 }
 
+export function isPlayableMedia(media) {
+  return (media.ebucoreHasMimeType.startsWith('video/') ||
```
rwd, line 24: This needs also to detect and return true for EUscreen videos. Check the URL against `http://www.euscreen.eu/item.html*`

plugins/media.js
rwd: Not yet, but we will, apparently.

rwd:
Superceded by #633.

PR #630, opened 2020-05-07
Fix errors throughout storybook

monkemedia:


.storybook/config.js
```diff
@@ -26,7 +26,10 @@ Vue.component('NuxtLink', {
   },
   template: '<a href="" @click.prevent="log()"><slot>NuxtLink</slot></a>',
 });
-
+Vue.prototype.$path = () => {}; // Mocking the path plugin
+Vue.prototype.$tFacetName = (name) => { return name }; // Mocking the tFacetName plugin
+Vue.prototype.$tNull = (key) => { return key }; // Mocking the tNull plugin
+Vue.prototype.$proxyMedia = (url, id) => { return id ? id + url : url }; // Mocking the proxyMedia plugin
```
lbiedinger, line 32: Move these up to the 'methods' section of the Vue instance:
```
Vue.mixin({
  methods: {
    localePath: (args) => args,
    $path: () => {},
    $tFacetName: (name) => { return name },
    $tNull: (key) => { return key },
    $proxyMedia: (url, id) => { return id ? id + url : url }
  }
});
```

components/generic/ImageWithAttribution.stories.js
```diff
@@ -1,14 +1,8 @@
...
+import VueI18n from 'vue-i18n';
 import ImageWithAttribution from './ImageWithAttribution.vue';
 
-const i18n = {
-  locale: 'en',
-  messages: {
-    en: {}
-  }
-};
-
```
lbiedinger, line 11: instead of getting rid of this just do:
```
const i18n = new VueI18n();
```
So it doesn't have to be initialized twice below

PR #632, opened 2020-05-08
Feat/ec 4368 stickynav desktop directive

lbiedinger:


plugins/vue-directives.js
```diff
@@ -0,0 +1,36 @@
...
+  inserted: (el) => {
+    el.scrolledVisible = true;
+    window.addEventListener('scroll', () => handleScroll(el));
+    window.addEventListener('orientationchange', () => handleOrientationChange(el));
+  }
+});
+
+const handleScroll = (el) => {
+  const newPosition = window.scrollY;
+  if (onDesktop && el.scrollPosition < newPosition && el.scrollPosition > 150 && el.scrolledVisible) {
```
monkemedia, line 13: 150 is when the scroll kicks in? If so, wonder if we should pass this value from the vue component?

plugins/vue-directives.js
```diff
@@ -0,0 +1,36 @@
...
+    el.scrolledVisible = false;
+  } else if (onDesktop && ((el.scrollPosition - 5) > newPosition || el.scrollPosition <= 150) && !el.scrolledVisible) {
+    el.setAttribute('style', 'transform: translate3d(0, 0, 0)');
+    el.scrolledVisible = true;
+  }
+  el.scrollPosition = newPosition;
+};
+
+const handleOrientationChange = (el) => {
+  console.log('orientation');
```
monkemedia, line 24: Do we need this console log?

plugins/vue-directives.js
```diff
@@ -0,0 +1,36 @@
+import Vue from 'vue';
+
+Vue.directive('visible-on-scroll', {
+  inserted: (el) => {
+    el.scrolledVisible = true;
+    window.addEventListener('scroll', () => handleScroll(el));
```
monkemedia, line 6: I can assume there are no issues with window being undefined due to SSR?

plugins/vue-directives.js
lbiedinger: this should be removed

plugins/vue-directives.js
lbiedinger: Potentially, but to limit the scope let's leave it hardcoded for now until we have an actual usecase.

plugins/vue-directives.js
alansutherland: haven't seen any ssr issues while testing, but just to be safe I've added this conditional so it'll only run client-side

   ```
if (process.browser) {
      window.addEventListener('scroll', () => handleScroll(el));
      window.addEventListener('orientationchange', () => handleOrientationChange(el));
}
```

PR #633, opened 2020-05-11
Media player: local iframe

rwd:
Notes:
* Requires setting `ENABLE_EUROPEANA_MEDIA_PLAYER=1` in the .env
* Bootstrap CSS is switched to being loaded from unpkg to ensure that it is only used in the default layout.


pages/media/index.vue
```diff
@@ -0,0 +1,64 @@
...
+      });
+    },
+
+    head() {
+      return {
+        title: 'Media player',
+
+        script: [
+          { src: 'https://code.jquery.com/jquery-3.4.1.min.js' },
+          { src: 'https://code.jquery.com/ui/1.12.1/jquery-ui.min.js' },
```
alansutherland, line 46: jquery & jquery ui are dependencies for the media player? anyway we can avoid using them and replace with something like https://umbrellajs.com/ ?

alansutherland: I can get the volume percentage displaying with a little bit of css (background-color & height), this would be an ok fix, though showing the handle too would be best

Screenshot 2020-05-13 at 17 27 21

pages/item/_.vue
```diff
@@ -339,9 +343,11 @@
         });
     },
 
-    mounted() {
+    created() {
       this.cardGridClass = this.isRichMedia && 'card-grid-richmedia';
```
lbiedinger, line 347: isRichMedia should be updated in the plugin to also check not just HTML video, but also any Playable Media in the case when the mediaPlayer is enabled.

rwd: Addressed in 9031114

PR #640, opened 2020-05-14
Reharvest automated record cards

mirjamverloop:


migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
+require('dotenv').config();
+
+const axios = require('axios');
+
+const recordOrigin = 'https://api.europeana.eu/record';
```
rwd, line 5: Let's keep this consistent with our use of "origin" elsewhere and make it just "https://api.europeana.eu" with the path included in the URL in `getRecord`.

migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
...
+      throw error;
+    });
+};
+
+module.exports = async function(migration) {
+  migration.transformEntries({
+    contentType: 'automatedRecordCard',
+    from: ['thumbnailUrl'],
+    to: ['thumbnailUrl'],
+    transformEntryForLocale: async(fromFields, currentLocale) => {
```
rwd, line 28: This does not need to be async.

migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
...
+    .then((response) => {
+      return response.data;
+    })
+    .catch((error) => {
+      throw error;
+    });
+};
+
+module.exports = async function(migration) {
+  migration.transformEntries({
```
rwd, line 24: This thumbnail URL handling would only need to be done on the first run, right? As in, once we've done so, it won't be an issue in future, but we may still want to re-run the rest of the migration?

If so, then let's split this out into its own migration.

migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
+require('dotenv').config();
+
+const axios = require('axios');
+
+const recordOrigin = 'https://api.europeana.eu/record';
+const apiKey = process.env['EUROPEANA_RECORD_API_KEY'];
+
+const getRecord = async(identifier) => {
+  const url = recordOrigin + '/search.json?profile=minimal&query=europeana_id%3A%22' + identifier + '%22&rows=1';
```
rwd, line 9: Include the URL query parameters in `params`, below.

migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
...
+
+      if (currentLocale !== 'en-GB' || !fromFields.identifier) return;
+      if (fromFields.encoding) return;
+
+      const record = await getRecord(fromFields.identifier[currentLocale]);
+      if (record.itemsCount === 1 && record.items) {
+        const encoding = record.items[0];
+        return { encoding };
+      }
+      return;
```
rwd, line 50: This is superfluous.

migrations/0028-reharvest-automated-record-cards.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+    shouldPublish: 'preserve'
+  });
+
+  migration.transformEntries({
+    contentType: 'automatedRecordCard',
+    from: ['identifier', 'encoding'],
+    to: ['encoding'],
+    transformEntryForLocale: async(fromFields, currentLocale) => {
+
```
rwd, line 41: Remove this empty line.

PR #650, opened 2020-05-27
Social share on gallery pages

mirjamverloop:
- Use the data from the encoding field to find the share image
- Updated Chromedriver


mirjamverloop:
npm mess...

PR #651, opened 2020-05-27
Social share icons on all gallery pages

mirjamverloop:
- Use data from encoding field to find the social share image 
- Check presence of social share in feature test

### DIFF
--- a/pages/galleries/_.vue
+++ b/pages/galleries/_.vue
@@ -41,10 +41,7 @@
     },
     computed: {
       shareMediaUrl() {
-        if (this.images.length <= 0) return null;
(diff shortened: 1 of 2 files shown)


pages/galleries/_.vue
```diff
@@ -42,8 +42,9 @@
     computed: {
       shareMediaUrl() {
         if (this.images.length <= 0) return null;
-        if (!this.images[0].fields.thumbnailUrl) return null;
```
rwd, line 45: This function can be rewritten as just:
```js
      shareMediaUrl() {
        return this.images.length === 0 ? null : this.imageUrl(this.images[0]);
      },
```

PR #655, opened 2020-05-29
Monolingual auto-suggest, to main search

alansutherland:


tests/unit/components/search/SearchForm.spec.js
```diff
@@ -138,16 +143,31 @@ describe('components/search/SearchForm', () => {
...
+          active: true,
+          userParams: {
+            query: ''
+          },
+          view: 'grid'
+        };
+        const wrapper = factory({ store: store(state) });
 
         wrapper.setData({
           suggestions: parsedSuggestions,
```
lbiedinger, line 156: this doesn't actually need to be set if it's setting the value for 'selectedSuggestion' already below.

tests/unit/components/search/SearchForm.spec.js
```diff
@@ -201,15 +221,20 @@ describe('components/search/SearchForm', () => {
...
-    const wrapper = factory();
+    const state = {
+      active: true,
+      userParams: {
+        query: ''
+      },
+      view: 'grid'
+    };
+    const wrapper = factory({ store: store(state) });
     wrapper.setData({ suggestions: parsedSuggestions });
```
lbiedinger, line 232: this is irrelevant for the new implementation now and can be removed.

tests/unit/components/search/SearchForm.spec.js
```diff
@@ -201,15 +221,20 @@ describe('components/search/SearchForm', () => {
...
-    it('generates agent entity URLs', () => {
-      wrapper.vm.suggestionLinkGen('http://data.europeana.eu/agent/base/59981').should.eq('/collections/person/59981-frank-sinatra');
-    });
-
-    it('generates concept entity URLs', () => {
-      wrapper.vm.suggestionLinkGen('http://data.europeana.eu/concept/base/227').should.eq('/collections/topic/227-fresco');
+    it('generates search suggestion URLs', () => {
+      const link = wrapper.vm.suggestionLinkGen('Fresco');
+      link.path.should.eq('/search');
+      link.query.query.should.eq('Fresco');
```
lbiedinger, line 237: with the state set as above you can also do:
```
      link.query.view.should.eq('grid');
```

tests/unit/components/search/SearchForm.spec.js
```diff
@@ -77,8 +77,13 @@ const entityApiSuggestionsResponse = {
   ]
 };
 const parsedSuggestions = {
-  'http://data.europeana.eu/concept/base/227': { en: 'Fresco' },
-  'http://data.europeana.eu/agent/base/59981': { en: 'Frank Sinatra' }
+  'Fresco': {
```
lbiedinger, line 80: Get rid of this block. What's here now is not what is passed in as 'suggestions' anyway.
The object passed in is still based of the entity URIs, but these are not utilized for any of the current functionality. We could potentially refactor this now, but I think the idea of linking the suggestions to entity(collection) pages is still valid so may be re-implemented later.

Also these data.europeana uris should be unique, which is why they make good keys for this array in the actual code.

tests/features/search/querying.feature
```diff
@@ -40,7 +40,7 @@ Feature: Search querying
 
   Scenario: No auto suggestion
     Given I am on the `search page`
-    And I enter "World" in the `search box`
+    And I enter "Hundred" in the `search box`
```
lbiedinger, line 43: The point of this specific scenario was to ensure the autosuggest feature was behind a feature toggle that's off by default, so the thing being searched for should be something that would yield results.

components/search/SearchForm.vue
```diff
@@ -125,7 +125,8 @@
         showSearchQuery: false,
         gettingSuggestions: false,
         suggestions: {},
-        selectedSuggestion: null
+        selectedSuggestion: null,
+        inputFocussed: false
```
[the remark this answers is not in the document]
alansutherland: yes

tests/features/search/querying.feature
alansutherland: thanks for explaining, have reverted

components/PageHeader.vue
```diff
@@ -3,7 +3,7 @@
     ref="nav-container"
     v-visible-on-scroll
     fluid
-    class="border-bottom d-flex py-0 px-lg-3 flex-column flex-lg-row align-items-center"
+    class="border-bottom d-flex py-0 px-lg-3 flex-column flex-lg-row align-items-center show"
```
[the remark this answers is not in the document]
alansutherland: this is strange, removing `show` doesn't fix it, even though it returns onDesktop as false the if statement is still running and setting the nav to translate to -100%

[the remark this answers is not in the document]
alansutherland: ok figured it out, the onDesktop function wasn't being called, it was just onDesktop not onDesktop()

components/search/SearchForm.vue
```diff
@@ -249,8 +248,8 @@
       },
 
       suggestionLinkGen(suggestion) {
+        // const query = this.updateCurrentSearchQuery(suggestion);
```
lbiedinger, line 251: Remove this comment.

components/search/AutoSuggest.vue
```diff
@@ -27,6 +27,7 @@
       :data-index="index"
       @mouseover="focus = index"
       @focus="index === focus"
+      @click="$root.$emit('emitFilterReset')"
```
lbiedinger, line 30: This messes up the state if you don't leave a collection page via the 'x' on the search pill.
Reproduce: 
* Go to a collection page
* Click on a search result
* Start typing something into the search box in order to get suggestions
* Click a suggestion
* Observe you'll be redirected to the search page.

Just get rid of this line and the reset filters call in the SearchInterface mounted section to fix this.

components/search/SearchInterface.vue
```diff
@@ -303,6 +303,9 @@
     },
     mounted() {
       this.showContentTierToast();
+      this.$root.$on('emitFilterReset', () => {
+        this.resetFilters();
+      });
```
lbiedinger, line 308: remove

components/search/AutoSuggest.vue
alansutherland: thanks, I have switched to your solution

components/search/SearchForm.vue
```diff
@@ -230,11 +230,8 @@
 
         this.gettingSuggestions = true;
 
-        // Query in the user's language, and English, removing duplicates
-        const languageParam = Array.from(new Set([this.$i18n.locale, 'en'])).join(',');
-
-        const suggestions = await getEntitySuggestions(query, {
-          language: languageParam
+        const suggestions = await getEntitySuggestions(encodeURIComponent(query), {
```
rwd, line 233: Do not URI-encode this. Instead, catch any errors from the API response and just show no suggestions.

components/search/SearchForm.vue
```diff
@@ -230,14 +231,14 @@
...
-
         const suggestions = await getEntitySuggestions(query, {
-          language: languageParam
+          language: this.$i18n.locale
         }, {
           recordValidation: this.enableSuggestionValidation
-        });
+        })
+          .catch((error) => {
+            throw apiError(error);
```
rwd, line 240: When errors are encountered from the entity suggestion endpoint, we do not want to pester the user about them, replacing the page content with an error message, but instead should just show no suggestions for their query. This should not re-throw the error, but should set suggestions to an empty array.

alansutherland: OK understood, I have instead replaced this with 
`this.isAutoSuggestActive;`
after `getEntitySuggestions`

So autosuggest is hidden if no suggestions are returned

PR #656, opened 2020-06-01
Exhibition headers

alansutherland:


assets/scss/breadcrumb.scss
```diff
@@ -1,8 +1,15 @@
 .breadcrumb {
-  background-color: transparent;
+  background-color: $white;
```
mirjamverloop, line 2: This makes the breadcrumbs on every page white, is that indeed correct?

assets/scss/breadcrumb.scss
```diff
@@ -1,8 +1,15 @@
...
   margin-top: -1rem;
-}
-
-.breadcrumb-item.active {
-  color: $mediumgrey;
+  border-top: 1px solid #e0e0e0;
+  border-bottom: 1px solid #e0e0e0;
```
mirjamverloop, line 5: Would you mind verifying with the design team, if we indeed need to introduce a new grey? Perhaps we can stick to the ones we already have. But if so, it might be worth adding a variable?

assets/scss/default.scss
```diff
@@ -1,5 +1,9 @@
+html {
+  background-color: $white;
```
mirjamverloop, line 2: Now all pages have a white background. Is that intended?

assets/scss/default.scss
```diff
@@ -1,5 +1,9 @@
...
+}
+
 body {
-  background-color: $bodygrey;
+  background-color: $white;
```
mirjamverloop, line 6: See previous comment.

assets/scss/hero.scss
```diff
@@ -54,14 +54,61 @@
     line-height: 1.5;
   }
 
-  figcaption cite {
-    margin-right: 5rem;
-  
-    @media (min-width: $bp-medium) {
-      margin-right: 1rem;
+  figcaption {
```
mirjamverloop, line 57: Could you make sure that the hero banner on the homepage for example still looks as before? The following styling updates affect the homepage too.

assets/scss/hero.scss
```diff
@@ -82,3 +129,24 @@
...
+    max-width: 51.25rem;
+  }
+  .jumbotron:after {
+    display: none;
+  }
+  article, section p {
+    max-width: 45rem;
+    margin: 0 auto;
+  }
+}
```
mirjamverloop, line 152: Could you add a new line to meet the linting standards?

assets/scss/hero.scss
```diff
@@ -82,3 +129,24 @@
...
 }
+
+.exhibition-page {
+  h1 {
+    font-size: 2.375rem;
+    margin-top: 2.5rem;
+  }
+  p.lead {
+    font-size: 1.5rem;
+    color: $slategrey;
```
mirjamverloop, line 140: Would you mind checking with the designers if there should be a new grey? See one of the other comments on greys too.

assets/scss/hero.scss
```diff
@@ -82,3 +129,24 @@
...
+  }
+  .hero-wrapper {
+    max-width: 51.25rem;
+  }
+  .jumbotron:after {
+    display: none;
+  }
+  article, section p {
+    max-width: 45rem;
+    margin: 0 auto;
```
mirjamverloop, line 150: This has some unwanted side effects. Go for example to /en/exhibitions/the-silk-and-the-blood/ravenna and look for the heading "City of mosaics", there is no longer space between the previous paragraph and this heading.

assets/scss/breadcrumb.scss
alansutherland: checked with design using $paper is fine

assets/scss/default.scss
alansutherland: design said only exhibitions pages should be white, this may be difficult to style

assets/scss/hero.scss
alansutherland: ok fixed this

assets/scss/hero.scss
alansutherland: changed to $mediumgrey

assets/scss/default.scss
alansutherland: I've fixed this

assets/scss/breadcrumb.scss
alansutherland: checked with design team, they confirmed this should affect all pages

PR #658, opened 2020-06-03
New linting rules

mirjamverloop:


assets/scss/badges.scss
```diff
@@ -13,7 +13,7 @@
   background-color: $blue;
   font-size: $font-size-small;
   font-weight: 400;
-  transition: .3s ease-in-out;
+  transition: 0.3s ease-in-out;
```
alansutherland, line 16: can we use the $standard-transition variable here?

mirjamverloop: Yes, good idea!

PR #665, opened 2020-06-10
Client-side rendering

rwd:
This work seeks to improve performance, of SSRs in particular, by:
1. Moving the rendering of certain components to client-side only, where deemed auxilliary content
2. Having the footer render client-side brings a significant improvement to every page
3. Introducing code-splitting for some components where their use is conditional

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -1,6 +1,8 @@
 <template>
   <div>
-    <CookieDisclaimer />
+    <client-only>
+      <CookieDisclaimer />
+    </client-only>
     <a
       class="skip-main"
       href="#main"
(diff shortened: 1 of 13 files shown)


layouts/default.vue
```diff
@@ -28,25 +30,41 @@
...
+    enableLanguageSelector: Boolean(Number(process.env['ENABLE_LANGUAGE_SELECTOR'])),
+    enableSuggestionValidation: Boolean(Number(process.env['ENABLE_ENTITY_SUGGESTION_RECORD_VALIDATION'])),
+    bootstrapVersion: require('bootstrap/package.json').version,
+    bootstrapVueVersion: require('bootstrap-vue/package.json').version
+  };
 
   export default {
     components: {
+      CookieDisclaimer: () => import('../components/generic/CookieDisclaimer'),
       PageHeader,
```
alansutherland, line 54: Can we also code split the header?

layouts/default.vue
```diff
@@ -28,25 +30,41 @@
         id="main"
       />
     </main>
-    <PageFooter />
+    <client-only>
```
alansutherland, line 33: Does the header need to be rendered server side or could we also only render that on the client to improve performance further?

PR #671, opened 2020-06-12
mobile nav styling

alansutherland:


tests/unit/components/PageHeader.spec.js
```diff
@@ -21,7 +21,27 @@ const store = (uiState = {}) => {
   return new Vuex.Store({
     getters,
     state: {
-      ui: uiState
+      ui: uiState,
```
rwd, line 24: I think these changes are all now redundant and should be removed.

components/PageHeader.vue
```diff
@@ -47,7 +47,7 @@
       class="p-lg-0 align-items-start justify-content-lg-end flex-column flex-lg-row d-none d-lg-block"
       role="navigation"
     >
-      <PageNavigation />
+      <PageNavigation :links="$store.state['link-group'].data.mainNavigation.links" />
```
rwd, line 50: Let's give the PageHeader a couple of computed properties using mapState, for mainNavigation and mobileNavigation so that these attributes can be reduced to e.g. `<PageNavigation :links="mainNavigation.links" />`

PR #673, opened 2020-06-12
Paragraphs narrower

alansutherland:


assets/scss/hero.scss
```diff
@@ -82,3 +82,11 @@
     }
   }
 }
+
+.exhibition-page {
+  article, section p {
```
mirjamverloop, line 87: Only the paragraphs are now 45rem wide max. Is that indeed correct? The alignment with titles, for example, is a bit off. I will add a screenshot to the ticket.

PR #675, opened 2020-06-12
header & attribution

alansutherland:


assets/scss/hero.scss
```diff
@@ -82,3 +82,75 @@
     }
   }
 }
+
+.exhibition-page {
+  h1 {
+    font-size: 2.375rem;
```
mirjamverloop, line 88: To keep it consistent with the other h1s, could you remove the font-size? 
I have verified this with the design team.

components/generic/HeroImage.vue
```diff
@@ -1,5 +1,47 @@
 <template>
+  <b-container
+    v-if="exhibition"
```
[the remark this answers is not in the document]
alansutherland: I agree, it needs a more generic name, I have renamed the prop to `compact`

components/generic/CiteAttribution.vue
```diff
@@ -1,5 +1,41 @@
...
+      >
+        {{ creator }}
+      </SmartLink>
+    </p>
+    <!-- <p>{{ $t('exhibitions.attribution.date') }}
+      TODO: data currently not provided by api
+      <span>{{ date }}</span>
+    </p> -->
+    <p>{{ $t('exhibitions.attribution.institution') }}
+      <span>{{ provider }}</span>
```
lbiedinger, line 22: this also needs to be a link as per latest discussion.

components/generic/HeroImage.vue
```diff
@@ -64,11 +109,20 @@
...
     },
+    data() {
+      return {
+        citeCollapsed: true
+      };
+    },
     computed: {
       optimisedImageUrl() {
-        return this.$options.filters.optimisedImageUrl(this.imageUrl, this.imageContentType, { width: 1920 });
+        return this.$options.filters.optimisedImageUrl(this.imageUrl, this.imageContentType, { width: this.exhibition ? 820 : 1920 });
```
rwd, line 125: When would `this.exhibition` ever be `true`?

Consider moving the image width into a prop instead, with the default of 1920.

components/generic/HeroImage.vue
```diff
@@ -1,5 +1,48 @@
...
+    <SocialShare
+      :media-url="imageUrl"
+    />
+    <b-jumbotron
+      :style="jumbotronStyle"
+      fluid
+      text-variant="white"
+      data-qa="hero banner"
+      class="mt-4"
+      @click="citeCollapsed = true"
```
rwd, line 20: Don't do this in the `@click` attributes themselves. Call event handler methods.

components/generic/CiteAttribution.vue
```diff
@@ -1,5 +1,61 @@
...
+        v-if="url !== null"
+        :destination="url"
+      >
+        {{ name }}
+      </SmartLink>
+      <span v-else>
+        {{ name }}
+      </span>
+    </p>
+    <p v-if="creator !== null">{{ $t('exhibitions.attribution.creator') }}
```
rwd, line 14: CiteAttribution should not need to know anything about exhibitions. From the perspective of this component, this is a stylistic variant. Rename accordingly.

components/generic/HeroImage.vue
alansutherland: `exhibition` was the previous name for the prop which is now called `compact`, it was changed after Mirjam requested a more generic name, I failed to update this within optimisedImageUrl. We could move the width in to a prop but `compact` is used for conditionally rendering the new header layout as well.

components/generic/HeroImage.vue
alansutherland: ok, moved to method

components/generic/CiteAttribution.vue
```diff
@@ -1,5 +1,61 @@
 <template>
-  <cite>
+  <cite v-if="extended">
+    <p v-if="name !== null">{{ $t('attribution.title') }}
```
lbiedinger, line 3: `!== null` is redundant. If the attributes are null this'll evaluate to false: `<p v-if="name">`

Same for the other values url, creator and provider

alansutherland: I have removed `!== null` I added it because previously I found that the creator tag was rendering without a value on http://localhost:3000/en/exhibitions/edible-plants-from-the-americas
but testing again with just `v-if="creator"` it is not rendering anymore.

PR #681, opened 2020-06-18
Account page

spyrosbek:
Implementation of an account-page containing user-created galleries presented in tabs.


assets/scss/tabs.scss
```diff
@@ -1,7 +1,7 @@
 .tabs {
   .card-header {
     background: $white;
-    border-bottom: 2px solid $offwhite;
+    border-bottom: 2px solid $paper;
```
mirjamverloop, line 4: I have checked with the design team, as both places where tabs are used have a different color. Their suggestion is to use #e7e7e9 instead. Would you mind updating the border color?

PR #683, opened 2020-06-18
More eslint style rules

rwd:
* Set many more eslint style rules
* Move CI linting to Travis instead of CodeClimate, the latter having its own locked versions of linting packages
* Update eslint and Vue plugin versions
* Fix linting errors


rwd:
@lbiedinger 
> These rules don't seem to propagate to code-climate properly.
That was due to CodeClimate having locked versions of eslint and its plugins. Addressed by moving the linting to Travis.

PR #684, opened 2020-06-18
ELK logging if configured

lbiedinger:


Jenkinsfile
```diff
@@ -37,14 +37,28 @@ pipeline {
       environment {
         CF_APP_NAME="portaljs${env.CF_SPACE == 'production' ? '' : '-' + env.CF_SPACE}"
         CTF_CPA_ACCESS_TOKEN=credentials("portaljs.${env.CF_SPACE}.contentful.cpa")
+        LOGSTASH_CONNECTION=credentials("elk_connection_string")
```
rwd, line 40: This isn't really a credential as it's just a username and hostname, plus it's the same for all jobs on our stack.

Let's therefore move it to a [global Jenkins env var](https://jenkins.eanadev.org/configure) (as with `CF_API` and `CF_ORG`), named `ELK_LOGSTASH_SSH_DESTINATION`.

Jenkinsfile
```diff
@@ -37,14 +37,28 @@ pipeline {
...
         HTTP_DIGEST_ACL=credentials("portaljs.${env.CF_SPACE}.http.digest.acl")
       }
       steps {
+        sh 'if [[ "${LOGSTASH_CONNECTION}" != "" ]]; then sed -i "s|services:|services:\\n  - log-drainer|" manifest.yml; fi'
```
rwd, line 44: I would be inclined to put the name of the log-drainer service into a global env var in Jenkins too, named `CF_LOG_DRAINER_SERVICE_NAME`

Jenkinsfile
```diff
@@ -37,14 +37,28 @@ pipeline {
...
         sh 'cf blue-green-deploy ${CF_APP_NAME} -f manifest.yml --delete-old-apps'
       }
     }
+    stage('Register with ELK') {
+      environment {
+        CF_APP_NAME="portaljs${env.CF_SPACE == 'production' ? '' : '-' + env.CF_SPACE}"
+        LOGSTASH_CONNECTION=credentials("elk_connection_string")
+      }
+      steps {
+      sh 'if [[ "${LOGSTASH_CONNECTION}" != "" ]]; then echo "No ELK configuration detected"; exit 0; fi'
```
rwd, line 56: It seems like this should be in a `when` block on the `stage`

manifest.yml
```diff
@@ -6,6 +6,7 @@ memory: 1024M
 health-check-type: process
 # health-check-http-endpoint: /robots.txt
 stack: cflinuxfs3
+services:
```
rwd, line 9: Does this still successfully deploy with both `cf push` and `cf bgd` if no services are added?

lbiedinger: no, i've changed it so this is written in along with the the name when present.

Jenkinsfile
```diff
@@ -12,8 +12,10 @@ pipeline {
     CF_HOME='/home/node' // Revert override from Jenkins global env
     CF_API="${env.CF_API}"
     CF_LOGIN=credentials('portaljs.cloudfoundry.login')
+    CF_LOG_DRAINER_SERVICE="${CF_LOG_DRAINER_SERVICE_NAME}"
```
rwd, line 15: This should be `    CF_LOG_DRAINER_SERVICE="${env.CF_LOG_DRAINER_SERVICE_NAME}"`

Jenkinsfile
```diff
@@ -40,11 +42,25 @@ pipeline {
...
         sh 'sed -i "s|env:|env:\\n  HTTP_DIGEST_ACL: ${HTTP_DIGEST_ACL}|" manifest.yml'
         sh 'cf blue-green-deploy ${CF_APP_NAME} -f manifest.yml --delete-old-apps'
       }
     }
+    stage('Register with ELK') {
+      when {
+        expression { return env.LOGSTASH_CONNECTION }
+      }
+      environment {
+        CF_APP_NAME="portaljs${env.CF_SPACE == 'production' ? '' : '-' + env.CF_SPACE}"
```
rwd, line 56: Move this up to the top level environment instead, along with the other declaration of it.

Jenkinsfile
```diff
@@ -40,11 +42,25 @@ pipeline {
...
+      }
+      steps {
+        sh 'APP_GUID="$(cf app ${CF_APP_NAME} --guid | head -1)"'
+        sh 'RESULT=(ssh ${LOGSTASH_CONNECTION} /etc/logstash/register_appname_guid.sh ${APP_NAME} ${APP_GUID})'
```
rwd, line 60: 1. Why is the result of this command being assigned to `RESULT`?
2. This is very specific to our current Jenkins config. The path to that shell script should not be in this Pipeline script.

lbiedinger: 1. This was copied from the API deployment script. I've confirmed with Patrick it's not needed.
2. Yes, as mentioned in Jira I also wasn't sure about this. I've now moved the entire command to a new variable(ELK_REGISTRATION_COMMAND), this removes the need for LOGSTASH_CONNECTION. So that's been removed.

Jenkinsfile
lbiedinger: done, although this assignment depends on the space being set which also happens in that block, so I hope this doesn't cause any issues.

deploy/cloud-foundry/Jenkinsfile
```diff
@@ -57,6 +60,16 @@ pipeline {
         }
       }
     }
+    stage('Register with ELK') {
```
rwd, line 63: As discussed on Slack, and verified on our Jenkins cluster, the ELK registration command will fail when run from inside the Docker agent.

Set a stage-specific agent so that this command is run on the Jenkins node's native shell.

deploy/cloud-foundry/Jenkinsfile
```diff
@@ -54,6 +57,17 @@ pipeline {
...
+      agent any // Don't run in docker container.
+      when {
+        expression { return env.ELK_REGISTRATION_COMMAND }
```
rwd, line 63: I am not sure this will work as intended.

It may need to be:
```groovy
when {
  not {
    environment name: 'ELK_REGISTRATION_COMMAND', value: ''
  }
}
```

PR #687, opened 2020-06-22
fix IIIF margin

alansutherland:


alansutherland:
@lbiedinger code climate issues have been fixed

PR #693, opened 2020-06-25
Initial set up transcriptions

mirjamverloop:


tests/unit/components/browse/ContentCardSection.spec.js
```diff
@@ -96,7 +96,7 @@ describe('components/browse/ContentCardSection', () => {
       wrapper.setProps({
         section: {
           hasPartCollection: {
-            items:[
+            items: [
```
lbiedinger, line 99: this was just flagged by the linter, nothing to do with this PR

pages/item/_.vue
```diff
@@ -135,7 +135,7 @@
                   </b-card-text>
                 </b-tab>
                 <b-tab
-                  v-if="showTranscription"
+                  v-if="mediaTranscription.length >= 1"
```
alansutherland, line 138: Can we use the strict equality operator `>==`

mirjamverloop: I am not sure if this exists in JavaScript.

alansutherland: @mirjamverloop it exists in Javascript, it's works the same as >= but avoids type conversion. so for example 
1 == "1" // true
1 === "1" // false
https://stackoverflow.com/a/359509/2474466

PR #699, opened 2020-06-30
Entity listing pages

alansutherland:


pages/collections/_type/index.vue
```diff
@@ -0,0 +1,110 @@
...
+        <b-card-group
+          class="card-deck-4-cols"
+          deck
+          :data-qa="`${ this.$route.params.type} listing page`"
+        >
+          <ContentCard
+            v-for="entity in entities"
+            :key="entity.id"
+            :title="entity.prefLabel[$i18n.locale]"
+            :url="entity.id"
```
lbiedinger, line 17: This results in an external URL to 'data.europeana.eu/...', these then redirect to the production portal collections page.

See how the components/entity/entityCards component converts 'data.europeana.eu/...' enitty URIs to portal URLs by means of a helper method `entityRoute`. However that component shows the cards in the wrong style, so just port the logic from there into the page here.

Alternatively, to avoid this page template from getting too big make a new component 'entityListingCards' which then combines the style from here as well as the logic from the entityCards component to make sure the links go to collection pages and are styled correctly.

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,110 @@
...
+        return {
+          name: 'collections-index',
+          params: {
+            pathMatch: this.$route.params.type
+          }
+        };
+      }
+    },
+    asyncData({ query, params, redirect, error, app }) {
+      const currentPage = pageFromQuery(query.page);
```
lbiedinger, line 76: Add a block here:
```
      if (!['persons', 'topics'].includes(params.type)) {
        return  error({ statusCode: 404, message: 'unknown collection type' });
      }
```

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,110 @@
...
+      if (currentPage === null) {
+        // Redirect non-positive integer values for `page` to `page=1`
+        query.page = '1';
+        return redirect(app.$path({ name: 'collections-type', params: { type: this.$route.params.type }, query }));
```
lbiedinger, line 80: `this.$route.params.type` can be `params.type`

tests/features/collections/topics.feature
```diff
@@ -0,0 +1,16 @@
...
+    And I see a `pagination navigation`
+    And I am on an accessible page
+
+  Scenario: Open a topic
+
+    When I open `/en/collections/topics`
+    Then I see a `content card` in the `topics listing page`
+    And I click a `content card`
+    And I wait for the page to load
+    Then I should not be on the `topics listing page`
```
lbiedinger, line 16: I doubt this'll work yet as the 'topics listing page' isn't defined yet.
Instead use
```
  Then I see an `entity page`
```
All the entity pages should have that as a qa element.

However for that to work see the changes to the contentCard/entityCard urls first, otherwise these links open in a new window and the test fails.

pages/collections/_type/index.vue
alansutherland: I've copied the logic from entityCards, it doesn't make the component to big, think it's a better approach than making a new entityListingCards component

tests/features/collections/topics.feature
alansutherland: @lbiedinger I have updated the feature test but it's still failing on this scenario
`3) Scenario: Open a topic # tests/features/collections/topics.feature:9
   ✔ Before # tests/features/config/cucumber.conf.js:73
   ✖ And I click a `content card` # tests/features/support/step-definitions.js:68
       Timed out while waiting for element <[data-qa="content card"]> to be present for 5000 milliseconds. - expected "visible" but got: "not found" (5086ms)`

pages/collections/_type/index.vue
lbiedinger: makes sense.

assets/scss/cards.scss
```diff
@@ -153,6 +153,7 @@
         overflow: hidden;
         display: -webkit-box;
         -webkit-line-clamp: 3;
+        text-transform: capitalize;
```
rwd, line 156: I don't think we should do this. It will affect all contexts, not just the new entity listing page, and may well be gramatically incorrect in some languages.

lang/en.js
```diff
@@ -77,6 +77,11 @@ export default {
   },
   "edmIsShownAtLinkAlt": "View at the provider's website",
   "error": "Error",
+  "entity": {
+    "index": {
+      "topics": "Subjects"
```
rwd, line 82: Change the keys so that this is at `pages.collections.topics.title`

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,127 @@
...
+          class="card-deck-4-cols"
+          deck
+          :data-qa="`${ this.$route.params.type} listing page`"
+        >
+          <ContentCard
+            v-for="entity in entities"
+            :key="entity.id"
+            :title="entity.prefLabel[$i18n.locale]"
+            :url="entityRoute(entity)"
+            :image-url="typeof entity.isShownBy !== 'undefined' ? entity.isShownBy.thumbnail : '/'"
```
rwd, line 18: This is too much logic to run in a template attribute. Move to a method.

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,127 @@
...
+          :limit="perPage"
+          :total-results="total"
+          :per-page="perPage"
+          :link-gen="paginationLink"
+        />
+      </b-col>
+    </b-row>
+  </b-container>
+</template>
+<script>
```
rwd, line 38: Add a blank line between template and script.

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,127 @@
...
+      <b-col cols="12">
+        <b-card-group
+          class="card-deck-4-cols"
+          deck
+          :data-qa="`${ this.$route.params.type} listing page`"
+        >
+          <ContentCard
+            v-for="entity in entities"
+            :key="entity.id"
+            :title="entity.prefLabel[$i18n.locale]"
```
rwd, line 16: This needs to be using our lang map lookup logic with fallback to English, i.e. `langMapValueForLocale` from plugins/europeana/utils.js

Otherwise some will have no title.

Examples, for English UI:
* Concept 572 on page 3 should have title "Progressiu" from `ca` lang
* Concept 702 on page 4 should have title "Alternative" from `de` lang

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,127 @@
...
+  <b-container>
+    <ContentHeader
+      :title="$tc('entity.index.' + this.$route.params.type, 2)"
+    />
+    <b-row class="flex-md-row pb-5">
+      <b-col cols="12">
+        <b-card-group
+          class="card-deck-4-cols"
+          deck
+          :data-qa="`${ this.$route.params.type} listing page`"
```
rwd, line 11: Remove the space before `this`

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,127 @@
...
+      const currentPage = pageFromQuery(query.page);
+      if (!['persons', 'topics'].includes(params.type)) {
+        return  error({ statusCode: 404, message: 'unknown collection type' });
+      }
+      if (currentPage === null) {
+        // Redirect non-positive integer values for `page` to `page=1`
+        query.page = '1';
+        return redirect(app.$path({ name: 'collections-type', params: { type: params.type }, query }));
+      }
+      const entityIndexParams = {
```
rwd, line 66: Include an additional parameter to limit the fields returned to those we need:
```js
fl: 'skos_prefLabel.*,isShownBy'
```

pages/collections/_type/index.vue
alansutherland: The ContentCard component already uses langMapValueForLocale for displayTitle so isn't this logic already implemented?

pages/collections/_type/index.vue
alansutherland: adding this param all thumbnails aren't rendered/returned and I get these console errors
```
[vuex] duplicate namespace http/ for the namespaced module http
[vuex] duplicate getter key: http/canonicalUrl
[vuex] duplicate getter key: http/canonicalUrlWithoutLocale
[vuex] duplicate getter key: http/origin
```

pages/collections/_type/index.vue
rwd: Yes, so change this line to:
```
            :title="entity.prefLabel"
```
Then the examples will have the fallback languages used.

alansutherland: ok I see, thanks

pages/collections/_type/index.vue
rwd: You are quite right.

Try this instead:
```js
fl: 'skos_prefLabel.*,isShownBy,isShownBy.thumbnail'
```

The vuex errors are unrelated. Try restarting Nuxt.

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,132 @@
...
+      thumbnail(entity) {
+        if (typeof entity.isShownBy !== 'undefined') return entity.isShownBy.thumbnail;
+      }
+    },
+    head() {
+      return {
+        title: this.$tc(`entity.index.${this.$route.params.type}.title`, 2)
+      };
+    },
+    watchQuery: ['query', 'page']
```
rwd, line 130: Remove `query` from here.

pages/collections/_type/index.vue
```diff
@@ -0,0 +1,132 @@
...
+          }
+        };
+      },
+      thumbnail(entity) {
+        if (typeof entity.isShownBy !== 'undefined') return entity.isShownBy.thumbnail;
+      }
+    },
+    head() {
+      return {
+        title: this.$tc(`entity.index.${this.$route.params.type}.title`, 2)
```
rwd, line 127: The string from the lang file does not include pluralisation, so there is no need to use `$tc`. Use `$t` instead, without the numeric argument.

This also duplicates the code from the `ContentHeader` title attribute, so instead return it as `title` from `asyncData`, then use that in both places.

lang/en.js
```diff
@@ -77,6 +77,13 @@ export default {
   },
   "edmIsShownAtLinkAlt": "View at the provider's website",
   "error": "Error",
+  "entity": {
+    "index": {
+      "topics": {
+        "title": "Subjects"
```
rwd, line 83: Change the keys so that this is at `pages.collections.topics.title`

plugins/europeana/entity.js
```diff
@@ -299,3 +299,26 @@ export function entityParamsFromUri(uri) {
   const type = getEntityTypeHumanReadable(matched[1]);
   return { id, type };
 }
+
+/**
+ * Return all entity subjects of type concept / agent
+ * @param {Object} params additional parameters sent to the API
+ */
+export function getEntityIndex(params = {}) {
```
rwd, line 307: This duplicates much of `searchEntities`. Refactor so that:
* The function currently named `searchEntitites` is named `findEntities`, updating any existing references to it
* The function currently named `getEntityIndex` is named `searchEntitites`
* `findEntities` calls `searchEntities` with the entity URIs gathered into the `query` property of the `params` argument
* `findEntities` returns just the entities from the response of `searchEntities`, not the total

lang/en.js
alansutherland: this has been changed to pages.collections.topics.title

plugins/europeana/entity.js
alansutherland: @rwd I have refactored findEntities and SearchEntities, findEntities is working on item pages, but there’s one issue

SearchEntities requires a type as one of it’s params if I view http://localhost:3000/en/item/2048001/Athena_Plus_ProvidedCHO_KIK_IRPA__Brussels__Belgium__AP_10325114
I need to send `type: 'person'` as the param

however, viewing http://localhost:3000/en/item/2048210/europeana_fashion_2005_6_0396 
I need to send `type: 'topic'` as the param

on the collections page we can get this info from `$route.params.type`
on items pages this is not possible since it's not included in the route params

checking the method `getRecord` on the item page and logging what data is returned from result.record I can see that the record api does return a type field, however, its value is ‘IMAGE’

rwd: @alansutherland try this:
```diff
diff --git a/pages/collections/_type/index.vue b/pages/collections/_type/index.vue
index a95deca2..5b3bf05b 100644
--- a/pages/collections/_type/index.vue
+++ b/pages/collections/_type/index.vue
@@ -43,6 +43,7 @@
   import {
     searchEntities,
     getEntitySlug,
+    getEntityTypeApi,
     getEntityTypeHumanReadable
   } from '../../../plugins/europeana/entity';
 
@@ -67,7 +68,7 @@
       const entityIndexParams = {
         query: '*:*',
         page: currentPage - 1,
-        type: params.type.slice(0, -1),
+        type: getEntityTypeApi(params.type.slice(0, -1)),
         pageSize: PER_PAGE,
         scope: 'europeana',
         fl: 'skos_prefLabel.*,isShownBy,isShownBy.thumbnail'
diff --git a/plugins/europeana/entity.js b/plugins/europeana/entity.js
index dac2fa4c..5ac821f3 100644
--- a/plugins/europeana/entity.js
+++ b/plugins/europeana/entity.js
@@ -58,7 +58,7 @@ export function getEntitySuggestions(text, params = {}) {
  * @param {string} type the type of the entity
  * @return {string} retrieved API name of type
  */
-function getEntityTypeApi(type) {
+export function getEntityTypeApi(type) {
   const names = {
     person: 'agent',
     topic: 'concept'
@@ -215,14 +215,14 @@ export function findEntities(entityUris) {
   if (entityUris.length === 0) return;
   const q = entityUris.join('" OR "');
   const params = {
-    query: `entity_uri:("${q}")`,
-    type: 'topic' /* needs to be dynamic */,
-    scope: 'europeana',
-    fl: 'skos_prefLabel.*,isShownBy,isShownBy.thumbnail'
+    query: `entity_uri:("${q}")`
   };
   return searchEntities(params)
     .then((response) => {
-      return response.entities;
+      return response.entities || [];
+    })
+    .catch((error) => {
+      throw error;
     });
 }
 
@@ -303,7 +303,6 @@ export function searchEntities(params = {}) {
   return axios.get(entityApiUrl('/search'), {
     params: {
       ...params,
-      type: getEntityTypeApi(params.type),
       wskey: config.entity.key
     }
   })
```

alansutherland: thanks for your help with resolving this using getEntityTypeApi

nuxt.config.js
```diff
@@ -166,6 +166,11 @@ const config = {
         path: '/*',
         component: 'pages/index.vue'
       });
+      routes.push({
```
rwd, line 169: Get this working as intended with:
```diff
diff --git a/nuxt.config.js b/nuxt.config.js
index 55a5e9a5..8638b281 100644
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -168,8 +168,8 @@ const config = {
       });
       routes.push({
         name: 'collections',
-        path: '/collections',
-        component: 'pages/collections/_type/_.vue'
+        path: '/(collections)',
+        component: 'pages/index.vue'
       });
     },
     linkExactActiveClass: 'exact-active-link'
```

alansutherland: @rwd thanks, the collections foyer page now works as expected, however, visiting https://localhost:3001/en/collections/topics

I see the error:
Error: Unsupported Entity type: undefined, should be one of: agent, concept, place, organization, timespan!

rwd: Can you double-check this? I see no such error.

alansutherland: I cannot replicate the error today either

PR #706, opened 2020-07-02
Default attribution style

alansutherland:


components/generic/ImageWithAttribution.vue
```diff
@@ -1,60 +1,106 @@
 <template>
-  <div>
-    <figure>
```
rwd, line 3: The full image with attribution comprises a figure with an image and figcaption. Without figure, figcaption has no bounding context.

components/generic/ImageWithAttribution.vue
```diff
@@ -1,60 +1,106 @@
...
-      <OptimisedImage
```
rwd, line 4: ImageWithAttribution needs to keep using OptimisedImage, so that images are lazy-loaded, and optimised via the Contentful Image API.

assets/scss/hero.scss
```diff
@@ -99,9 +100,30 @@
...
   }
-  .hero-wrapper {
+  @media (min-width: $bp-extralarge) {
+    blockquote + p cite {
+      width: 38.3rem;
+    }
+  }
+}
+
+.default-attribution {
```
rwd, line 110: As this is not used on the hero image, move it into its own SCSS file, attributions.scss, and rename the class to "figure-attribution".

alansutherland: sure

PR #707, opened 2020-07-03
UI element for like

annachristaki:


components/generic/ContentCard.vue
```diff
@@ -85,9 +85,17 @@
       class="user-buttons"
       data-qa="user buttons"
     >
-      <b-button>
-        <i class="icon-ic-add" />
-      </b-button>
+      <b-button
+        class="icon-ic-add"
+        data-qa="add to gallery button"
+        :aria-label="$t('addToGallery')"
```
rwd, line 91: Change the i18n key to `actions.addToGallery`.

components/generic/ContentCard.vue
```diff
@@ -85,9 +85,17 @@
...
+      />
+      <b-button
+        :pressed.sync="liked"
+        class="icon-heart"
+        data-qa="like button"
+        :aria-label="$t('like')"
```
rwd, line 97: Change the i18n key to `actions.like`.

PR #710, opened 2020-07-07
Collection page header

alansutherland:


assets/scss/hero.scss
```diff
@@ -195,3 +195,9 @@
   }
 }
 
+.entity-page {
```
rwd, line 198: Shouldn't this be in the entity page view itself, as it won't be used elsewhere?

components/generic/RelatedChip.vue
```diff
@@ -9,7 +9,7 @@
   >
     <b-img
       v-if="img"
-      :src="img"
+      :src="`${img}&size=w200`"
```
rwd, line 12: Don't do this here. Move it up to the parent component's prop provision.

components/generic/RelatedChip.vue
```diff
@@ -9,7 +9,7 @@
...
       :alt="title"
```
rwd, line 13: While not from this branch, let's change this to `alt=""`, because the title text that follows is identical, so this is worse for accessibility.

assets/scss/hero.scss
alansutherland: yes, thats correct, I've now moved it

components/entity/EntityDetails.vue
```diff
@@ -46,6 +26,10 @@
         >
           {{ showAll ? $t('showLess') : $t('showMore') }}
         </b-button>
+        <p class="curated d-flex align-items-center">
```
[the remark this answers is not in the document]
alansutherland: sure

components/generic/RelatedChip.vue
alansutherland: ok

components/generic/RelatedChip.vue
alansutherland: yep that's a good point about accessibility

pages/collections/_type/_.vue
```diff
@@ -261,6 +276,15 @@
           values: [title],
           code: null
         };
+      },
+      suggestionLinkGen(id, prefLabel) {
```
rwd, line 280: Rename this `relatedLinkGen` as it has nothing to do with suggestions.

pages/collections/_type/_.vue
```diff
@@ -1,67 +1,80 @@
...
+                :title="relatedEntity.prefLabel[$i18n.locale]"
+                :img="`${relatedEntity.isShownBy.thumbnail}&size=w200`"
+                data-qa="related entities"
+              />
+            </section>
+            <section
+              v-else-if="relatedCollectionCards"
+              class="mb-2"
+            >
+              <RelatedChip
```
rwd, line 42: These links do not work. See for example the curated card links at http://localhost:3000/en/collections/topic/55-fashion

alansutherland: for some reason on related collection cards the identifier wasn't being sent to linkGen function, I have refactored the linkGen to take in the entire card/item object as an argument and then access the id/identifier from it, links are now working on both related collection cards and related entities.

pages/collections/_type/_.vue
```diff
@@ -261,6 +277,24 @@
           values: [title],
           code: null
         };
+      },
+      relatedLinkGen(item) {
+        let id = '',
```
rwd, line 282: This shorthand syntax is not one we use elsewhere. Replace with:
```js
let id;
let name;
```

pages/collections/_type/_.vue
```diff
@@ -261,6 +277,24 @@
...
+            name = '';
+        if (typeof item.id === 'undefined') {
+          id = item.identifier;
+          name = item.name;
+        } else {
+          id = item.id;
+          name = item.prefLabel[this.$i18n.locale];
```
rwd, line 289: For the URL slugs, we always use English, so change this to `item.prefLabel.en`.

Otherwise, when on http://localhost:3000/de/collections/topic/221-watercolor-painting, following the link to "Zeichnung (Kunst)" takes you first to http://localhost:3000/de/collections/topic/35-zeichnung-(kunst), which then redirects to http://localhost:3000/de/collections/topic/35-drawing.

Use the English slug here and that redirect is avoided.

components/generic/RelatedChip.vue
```diff
@@ -28,8 +28,8 @@
         default: ''
       },
       linkGen: {
```
rwd, line 30: As this is no longer a link _generating_ function, rename it to `linkTo`

pages/collections/_type/_.vue
alansutherland: sure

components/generic/RelatedChip.vue
alansutherland: ok

pages/collections/_type/_.vue
alansutherland: sorry my mistake

assets/scss/variables.scss
```diff
@@ -6,6 +6,7 @@ $white: #fff;
 $grey: #777;
 $offwhite: #f1f1ee;
 $paper: #d9d4ca;
+$lightmagenta: #e7e7e9;
```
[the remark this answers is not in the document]
alansutherland: I wanted to use a more descriptive name than `someothershadeofgrey` since we already have several shades of grey in the variables, I googled the hexcode and #e7e7e9 is a type of magenta https://encycolorpedia.com/e7e7e9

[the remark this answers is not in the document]
lbiedinger: let's go for lightbluemagenta then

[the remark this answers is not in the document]
alansutherland: ok updated

PR #714, opened 2020-07-09
Rights statement button

alansutherland:


components/generic/RightsStatementButton.vue
```diff
@@ -0,0 +1,45 @@
...
+  import RightsStatement from '../../components/generic/RightsStatement';
+  import SmartLink from '../../components/generic/SmartLink';
+
+  export default {
+    components: {
+      RightsStatement,
+      SmartLink
+    },
+    props: {
+      rightsStatementIsUrl: {
```
rwd, line 24: I don't think this really makes sense as a prop as it is conditional for rendering the entire template. This should instead be left to the parent view to determine whether or not to even render the RightsStatementButton component. Remove this prop.

components/generic/RightsStatementButton.stories.js
```diff
@@ -0,0 +1,25 @@
...
+const i18n = new VueI18n();
+
+storiesOf('Generic', module)
+  .add('Rights Statement Button', () => ({
+    components: { RightsStatementButton },
+    i18n,
+    data() {
+      return {
+        rightsStatementIsUrl: true,
+        rightsStatement: 'https://creativecommons.org/licenses/by-sa/4.0/',
```
rwd, line 14: Remove the trailing comma.

components/generic/RightsStatementButton.stories.js
```diff
@@ -0,0 +1,25 @@
...
+      };
+    },
+    template: `
+      <b-container class="mt-3 bg-white p-5">
+        <RightsStatementButton
+          v-if="rightsStatementIsUrl"
+          :rights-statement="rightsStatement"
+        />
+      </b-container>
+    `
```
rwd, line 24: Remove the blank line.

components/generic/RightsStatementButton.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <SmartLink
```
rwd, line 2: This results in the icon for external links showing, but in the design, and as clarified during sprint refinement, in this context it is not wanted. Could you come up with a way to not show it here, but to continue showing it in all other contexts, likely via a prop sent down to SmartLink.

components/generic/RightsStatementButton.vue
alansutherland: true, I have removed it

components/generic/RightsStatementButton.stories.js
alansutherland: removed

components/generic/RightsStatementButton.stories.js
alansutherland: removed

components/generic/RightsStatementButton.vue
alansutherland: I have added the hideExternalIcon prop, other smartLinks will still show the external link icon

PR #715, opened 2020-07-10
Github Actions workflow for CI

rwd:


.cfignore
```diff
@@ -87,4 +87,3 @@ tmp
 
 # Tests
 /tests
-tmp
```
rwd, line 90: Removed because already ignored earlier in the file.

PR #717, opened 2020-07-10
Account dropdown-menu

spyrosbek:


components/PageNavigation.vue
```diff
@@ -26,18 +26,40 @@
...
         >
-          <span>{{ $t('account.linkAccount') }}</span>
-        </b-link>
+          <template
+            slot="button-content"
+          >
+            <span class="label">{{ $t('account.linkAccount') }}</span>
+          </template>
+          <b-dropdown-item
+            :to="localePath({ name: 'account' })"
```
annachristaki, line 42: Link to user account page from XX-24

pages/account/logout.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <div />
+</template>
+
+<script>
+  export default {
```
rwd, line 6: Use the minimal layout, and move the logout to middleware, by applying this patch:
```diff
diff --git a/pages/account/login.vue b/pages/account/login.vue
index 5a3c1e18..47e971f9 100644
--- a/pages/account/login.vue
+++ b/pages/account/login.vue
@@ -6,9 +6,8 @@
   export default {
     auth: false,
     layout: 'minimal',
-    mounted() {
-      this.$auth.loginWith('keycloak');
+    middleware({ $auth }) {
+      $auth.loginWith('keycloak');
     }
   };
 </script>
-
diff --git a/pages/account/logout.vue b/pages/account/logout.vue
index 499a9ba3..ab45ad05 100644
--- a/pages/account/logout.vue
+++ b/pages/account/logout.vue
@@ -5,10 +5,12 @@
 <script>
   export default {
     auth: true,
-    layout: 'default',
-    middleware: 'auth',
-    mounted() {
-      this.$auth.logout();
-    }
+    layout: 'minimal',
+    middleware: [
+      'auth',
+      ({ $auth }) => {
+        $auth.logout();
+      }
+    ]
   };
 </script>
```
(It also applies the same approach to the login.)

components/PageNavigation.vue
```diff
@@ -26,12 +26,35 @@
...
+          :class="isAccountPage && 'exact-active-link'"
         >
-          <span>{{ $t('account.linkAccount') }}</span>
-        </b-link>
+          <template
+            slot="button-content"
+          >
+            <span class="label">{{ $t('account.linkAccount') }}</span>
+          </template>
+          <b-dropdown-item
```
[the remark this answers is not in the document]
spyrosbek: What about the <b-dropdown-divider /> element that goes after the 2nd link?

pages/account/logout.vue
spyrosbek: I noticed that, when I do login & logout this way, when I try to access some pages (account page and logout page) unauthorized, the redirection is not working correctly (user is stuck on blank login and logout pages, without being redirected to keycloak-login and home page, respectively)

spyrosbek: @rwd Should I proceed with this patch anyway?

components/PageNavigation.vue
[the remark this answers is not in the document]
spyrosbek: @rwd Should I omit the dropdown-divider or can you suggest any workarounds?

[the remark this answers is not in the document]
rwd: @spyrosbek how about an additional property of the auth link objects indicating whether a divider should be included after the dropdown item? I have updated the example code above to show what I mean.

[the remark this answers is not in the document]
spyrosbek: If we want to stick to the design, the most elegant solution that I came up with, is this :
`<template v-for="(item, index) in authLinks">
            <b-dropdown-divider
              v-if="item.divider"
              :key="index"
            />
            <b-dropdown-item
              v-else
              :key="index"
              :to="item.to"
              :href="item.href"
            >
              <span class="label">{{ item.text }}</span>
            </b-dropdown-item>
          </template>`

And adding the entry
`{ divider: true },`
at the authLinks array.

Does this look good?

pages/account/logout.vue
rwd: @spyrosbek I see.

How about this instead:
```diff
diff --git a/pages/account/login.vue b/pages/account/login.vue
index 5a3c1e18..80d5eb66 100644
--- a/pages/account/login.vue
+++ b/pages/account/login.vue
@@ -6,9 +6,8 @@
   export default {
     auth: false,
     layout: 'minimal',
-    mounted() {
+    created() {
       this.$auth.loginWith('keycloak');
     }
   };
 </script>
-
diff --git a/pages/account/logout.vue b/pages/account/logout.vue
index 499a9ba3..ab45ad05 100644
--- a/pages/account/logout.vue
+++ b/pages/account/logout.vue
@@ -5,10 +5,12 @@
 <script>
   export default {
     auth: true,
-    layout: 'default',
-    middleware: 'auth',
-    mounted() {
-      this.$auth.logout();
-    }
+    layout: 'minimal',
+    middleware: [
+      'auth',
+      ({ $auth }) => {
+        $auth.logout();
+      }
+    ]
   };
 </script>
```

components/PageNavigation.vue
```diff
@@ -26,12 +26,35 @@
...
+            <span class="label">{{ $t('account.profile') }}</span>
+          </b-dropdown-item>
+          <b-dropdown-item
+            :href="accountSettingsPage"
+          >
+            <span class="label">{{ $t('account.settings') }}</span>
+          </b-dropdown-item>
+          <b-dropdown-divider />
+          <b-dropdown-item
+            :to="localePath({ name: 'account-logout' })"
```
rwd, line 53: I have just realised that the account logout URL should not be localised. Could you make that change with this patch:
```diff
diff --git a/components/PageNavigation.vue b/components/PageNavigation.vue
index 544c6e12..93dfa9cb 100644
--- a/components/PageNavigation.vue
+++ b/components/PageNavigation.vue
@@ -50,7 +50,7 @@
           </b-dropdown-item>
           <b-dropdown-divider />
           <b-dropdown-item
-            :to="localePath({ name: 'account-logout' })"
+            :to="{ name: 'account-logout' }"
           >
             <span class="label">{{ $t('account.linkLogout') }}</span>
           </b-dropdown-item>
diff --git a/middleware/l10n.js b/middleware/l10n.js
index 58c12363..7d0f19af 100644
--- a/middleware/l10n.js
+++ b/middleware/l10n.js
@@ -18,7 +18,11 @@ function appSupportsLocale(locale) {
 
 export default ({ app, route, redirect, req }) => {
   // Exit early if this is an auth callback
-  if (app.$auth && (route.path === app.$auth.options.redirect.callback || route.path === app.$auth.options.redirect.login)) return;
+  if (app.$auth && [
+    app.$auth.options.redirect.callback,
+    app.$auth.options.redirect.login,
+    app.$auth.options.redirect.logout
+  ].includes(route.path)) return;
 
   // Is there a locale in the URL path already?
   const routePathLocaleMatch = route.path.match(/^\/([a-z]{2})(\/.*)?$/);
diff --git a/nuxt.config.js b/nuxt.config.js
index faa82fe0..9f7345cb 100644
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -144,11 +144,12 @@ const config = {
         silentFallbackWarn: true,
         dateTimeFormats: i18nDateTime
       },
-      // Disable redirects to account callback & login pages
+      // Disable redirects to account pages
       parsePages: false,
       pages: {
         'account/callback': false,
-        'account/login': false
+        'account/login': false,
+        'account/logout': false
       },
       // Enable browser language detection to automatically redirect user
       // to their preferred language as they visit your app for the first time
@@ -232,7 +233,7 @@ if (Number(process.env['ENABLE_XX_USER_AUTH'])) {
     // 'home' option for redirection after login
     redirect: {
       login: '/account/login',
-      logout: '/',
+      logout: '/account/logout',
       callback: '/account/callback',
       home: '/account'
     },
```

components/PageNavigation.vue
```diff
@@ -227,4 +251,36 @@
...
+      border-radius: 4px;
+      box-shadow: 2px 2px 4px 0 rgba(0, 0, 0, 0.2);
+      border: solid 1px #dcdcde;
+      li a {
+        padding: 0.4rem 1.2rem 0.4rem 1.2rem;
+        &:hover {
+          background-color: #dcdcde;
+        }
+        .label {
+          height: 2.4rem;
```
alansutherland, line 281: this is an unusual way to style this, maybe it will be fine but there could be cross-browser bugs. Using padding is bulletproof, so remove this style and on `li a` change padding to `0.85rem 1rem` this will make it 48px. Currently height 2.4rem is making it 53px, the design is 48px.

components/PageNavigation.vue
```diff
@@ -62,12 +83,25 @@
         default: () => []
       }
     },
+    data() {
+      return {
+        authLinks: [
+          { to: this.localePath({ name: 'account' }), text: this.$t('account.profile') },
```
rwd, line 89: Instead of using `this.localePath` use `this.$path` (from `modules/http/templates/plugin.path.js`).

PR #719, opened 2020-07-13
Download button

alansutherland:


components/generic/DownloadButton.stories.js
```diff
@@ -0,0 +1,60 @@
...
+        <DownloadButton
+          :url="url"
+          :download-url="downloadUrl"
+          :download-disabled="downloadDisabled"
+          :use-proxy="useProxy"
+        />
+      </b-container>
+    `
+  }))
+  .add('Download Button Disabled', () => ({
```
rwd, line 40: The two stories should be grouped under "Download Button", as "Enabled" and "Disabled". See the other grouped stories under e.g. Generic/Messages and Generic/HTMLEmbed.

components/generic/DownloadButton.vue
```diff
@@ -0,0 +1,54 @@
...
+  >
+    <span class="ic-download d-inline-flex pr-1"></span>
+    {{ $t('actions.download') }}
+  </b-button>
+</template>
+
+<script>
+  export default {
+    props: {
+      downloadUrl: {
```
rwd, line 19: Rename this to `url`. It's implied that it's the download URL given this is the download button component.

components/generic/DownloadButton.vue
```diff
@@ -0,0 +1,54 @@
...
+        type: String,
+        default: ''
+      },
+      downloadDisabled: {
```
rwd, line 23: Rename this to `disabled`. It's implied that it's the download that's disabled given this is the download button component.

components/generic/DownloadButton.vue
```diff
@@ -0,0 +1,54 @@
...
+        type: Boolean,
+        default: false
+      },
+      useProxy: {
```
rwd, line 27: With the download button in its own component and having been supplied the constructed URL -- proxied or otherwise -- it need not be aware of whether there is a proxy.

Therefore rename this to `target`, type `String`, and default "_blank".

components/generic/DownloadButton.stories.js
alansutherland: yes, good point

components/generic/DownloadButton.vue
alansutherland: ok

components/generic/DownloadButton.vue
alansutherland: ok

components/generic/DownloadButton.vue
alansutherland: yes good point

components/generic/DownloadButton.stories.js
```diff
@@ -0,0 +1,64 @@
...
+      InfoMessage
+    },
+    data() {
+      return {
+        url: 'ggpht.com',
+        disabled: true
+      };
+    },
+    template: `
+      <b-container class="mt-3 p-5 bg-white">
```
lbiedinger, line 52: Shouldn't this _not_ have the white background?
The info message will sit directly on the page.

alansutherland: ok I have removed the background

PR #722, opened 2020-07-15
Check for record cards before entity type lookup

lbiedinger:


components/browse/ContentCardSection.vue
```diff
@@ -83,6 +83,7 @@
       isPeopleSection() {
         if (this.cards.length !== 4) return false;
         return this.cards.every((card) => {
+          if (card['__typename'] === 'AutomatedRecordCard') return false;
```
rwd, line 86: Shouldn't we instead be checking for the card types that this _does_ match against? i.e.
```diff
diff --git a/components/browse/ContentCardSection.vue b/components/browse/ContentCardSection.vue
index bd66f56d..5fedf111 100644
--- a/components/browse/ContentCardSection.vue
+++ b/components/browse/ContentCardSection.vue
@@ -83,7 +83,7 @@
       isPeopleSection() {
         if (this.cards.length !== 4) return false;
         return this.cards.every((card) => {
-          if (card['__typename'] === 'AutomatedRecordCard') return false;
+          if (card['__typename'] !== 'AutomatedEntityCard') return false;
           const identifier = card.identifier;
           return identifier ? entityParamsFromUri(identifier).type === 'person' : false;
         });
```

PR #723, opened 2020-07-16
Metadata box refactor

lbiedinger:
Refactor the metadatabox into it's own component in order to use it in storybook and demonstrate how it expands to the width of it's container.

### DIFF
--- a/components/item/MetadataBox.stories.js
+++ b/components/item/MetadataBox.stories.js
@@ -1,6 +1,5 @@
 import { storiesOf } from '@storybook/vue';
 import VueI18n from 'vue-i18n';
-import StoryRouter from 'storybook-vue-router';
 import MetadataBox from './MetadataBox';
 
 const i18n = new VueI18n({
@@ -70,7 +69,7 @@ const i18n = new VueI18n({
 });
 
 const formattedMetadata = {
-  coreFields: {
+  coreMetadata: {
     edmDataProvider: {
       def: ['Provider']
     },
@@ -86,7 +85,7 @@ const formattedMetadata = {
     },
     dcType: { en: ['Format'] }
   },
-  allMetaData: {
+  allMetadata: {
     edmDataProvider: {
       def: ['Provider']
     },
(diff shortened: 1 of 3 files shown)


pages/item/_.vue
```diff
@@ -97,69 +97,11 @@
...
-                      <hr
-                        v-if="index !== (transcribingAnnotations.length - 1)"
-                      >
-                    </div>
-                  </b-card-text>
-                </b-tab>
-              </b-tabs>
-            </b-card>
-          </div>
+          <MetadataBox
```
alansutherland, line 100: The main content & metadata box should take up the full width of the page with a maximum width of 1024 set. The acceptance criteria maybe doesn't explain this well but I believe we need to remove the column containing the relatedEntities and set the main content col to full width on desktop not col-lg-9.

@ashmarriotteuropeana can you confirm this is within scope of this ticket? i.e. removing the relatedEntities column to allow the main content to be full width

components/item/MetadataBox.stories.js
```diff
@@ -0,0 +1,172 @@
...
+  transcribingAnnotations: []
+};
+
+storiesOf('Item page', module)
+  .addDecorator(StoryRouter({}, {
+    routes: [
+      { name: 'item-all', path: '/item/*' }
+    ]
+  }))
+  .add('Metadata Box 12 columns', () => ({
```
rwd, line 132: Group these stories so that they both appear under "Item page/Metadata box"

components/item/MetadataBox.vue
```diff
@@ -0,0 +1,89 @@
...
+
+<script>
+  import MetadataField from './MetadataField';
+  export default {
+    name: 'MetadataBox',
+    components: {
+      MetadataField
+    },
+    props: {
+      coreFields: {
```
rwd, line 75: Rename to `coreMetadata`

components/item/MetadataBox.vue
```diff
@@ -0,0 +1,89 @@
...
+        type: Object,
+        default: null
+      },
+      allMetaData: {
```
rwd, line 79: Rename to `allMetadata`

components/item/MetadataBox.stories.js
```diff
@@ -0,0 +1,172 @@
...
+    dcFormat: {
+      en: ['Photography']
+    },
+    keywords: {}
+  },
+  transcribingAnnotations: []
+};
+
+storiesOf('Item page', module)
+  .addDecorator(StoryRouter({}, {
```
rwd, line 127: This appears to be redundant. Remove it.

PR #724, opened 2020-07-16
Share modal

alansutherland:


assets/scss/modal.scss
```diff
@@ -26,4 +26,67 @@
...
+        padding: 1rem 1.5rem 1.5rem 1.5rem;
+        .icon-wrapper > div {
+          display: flex;
+          justify-content: space-between;
+        }
+        a {
+          border-radius: 0.25rem;
+          width: calc(100% / 3 - 10px);
+          height: 3rem;
+          &.facebook {
```
rwd, line 55: Do these need to be here in the shared SCSS or should they be scoped within the component?

assets/scss/modal.scss
```diff
@@ -26,4 +26,67 @@
...
+        }
+        button {
+          font-weight: 600;
+        }
+      }
+    }
+  }
+}
+
+#shareModal + .modal-backdrop {
```
rwd, line 90: This should be scoped in the component.

components/generic/SocialShareModal.stories.js
```diff
@@ -0,0 +1,42 @@
+import { storiesOf } from '@storybook/vue';
+import Vuex from 'vuex';
+import SocialShareModal from './SocialShareModal.vue';
+import ShareButton from './ShareButton.vue';
+
+const i18n = {
+  locale: 'en',
+  messages: {
+    en: {
+      actions: {
```
rwd, line 10: Add the shareOn key here so that the link titles in the modal are populated correctly.

components/generic/ShareButton.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
rwd, line 1: Let's move the three sharing-related components to components/sharing

components/generic/SocialShareModal.stories.js
```diff
@@ -0,0 +1,42 @@
...
+        close: 'close',
+        share: 'share'
+      }
+    }
+  }
+};
+
+const store = new Vuex.Store({
```
rwd, line 18: This appears not to do anything as removing it has no impact.

Let's try and get an actual URL into the share links instead. Use https://www.example.org/

components/generic/SocialShare.vue
```diff
@@ -4,28 +4,43 @@
       target="_blank"
       :title="$t('actions.shareOn', { social: 'Facebook' })"
       :href="facebookShareUrl"
-      class="social-share"
+      class="social-share facebook"
       data-qa="share facebook button"
     >
       <span class="icon-facebook" />
+      <span
```
rwd, line 11: Let's DRY this component up with:
```diff
diff --git a/components/generic/SocialShare.vue b/components/generic/SocialShare.vue
index bb10f988..75135d90 100644
--- a/components/generic/SocialShare.vue
+++ b/components/generic/SocialShare.vue
@@ -1,45 +1,19 @@
 <template>
   <div>
     <b-link
+      v-for="(network, index) in networks"
+      :key="index"
       target="_blank"
-      :title="$t('actions.shareOn', { social: 'Facebook' })"
-      :href="facebookShareUrl"
-      class="social-share facebook"
-      data-qa="share facebook button"
+      :title="$t('actions.shareOn', { social: network.name })"
+      :href="network.url"
+      :class="`social-share ${network.identifier}`"
+      :data-qa="`share ${network.identifier} button`"
     >
-      <span class="icon-facebook" />
+      <span :class="`icon-${network.identifier}`" />
       <span
         v-if="withText"
         class="text"
-      >Facebook
-      </span>
-    </b-link>
-    <b-link
-      target="_blank"
-      :title="$t('actions.shareOn', { social: 'Twitter' })"
-      :href="twitterShareUrl"
-      class="social-share twitter"
-      data-qa="share twitter button"
-    >
-      <span class="icon-twitter" />
-      <span
-        v-if="withText"
-        class="text"
-      >Twitter
-      </span>
-    </b-link>
-    <b-link
-      target="_blank"
-      :title="$t('actions.shareOn', { social: 'Pinterest' })"
-      :href="pinterestShareUrl"
-      class="social-share pinterest"
-      data-qa="share pinterest button"
-    >
-      <span class="icon-pinterest" />
-      <span
-        v-if="withText"
-        class="text"
-      >Pinterest
+      >{{ network.name }}
       </span>
     </b-link>
   </div>
@@ -64,23 +38,30 @@
 
     data() {
       return {
-        location: null
+        networks: [
+          {
+            identifier: 'facebook',
+            name: 'Facebook',
+            url: `https://www.facebook.com/sharer/sharer.php?display=page&u=${this.shareUrl}`
+          },
+          {
+            identifier: 'pinterest',
+            name: 'Pinterest',
+            url: `https://pinterest.com/pin/create/link/?url=${this.shareUrl}&media=${this.mediaUrl}`
+          },
+          {
+            identifier: 'twitter',
+            name: 'Twitter',
+            url: `https://twitter.com/intent/tweet?text=${this.shareUrl}`
+          }
+        ]
       };
     },
 
     computed: {
       ...mapGetters({
         shareUrl: 'http/canonicalUrl'
-      }),
-      facebookShareUrl() {
-        return `https://www.facebook.com/sharer/sharer.php?display=page&u=${this.shareUrl}`;
-      },
-      twitterShareUrl() {
-        return `https://twitter.com/intent/tweet?text=${this.shareUrl}`;
-      },
-      pinterestShareUrl() {
-        return `https://pinterest.com/pin/create/link/?url=${this.shareUrl}&media=${this.mediaUrl}`;
-      }
+      })
     }
   };
 </script>
```

components/generic/SocialShare.vue
```diff
@@ -40,6 +55,10 @@
       mediaUrl: {
         type: String,
         default: ''
+      },
+      withText: {
```
rwd, line 59: Is this necessary, or could we always render the text, which is better accessibility, and just use CSS in a parent component to not display it if preferred in that context?

components/generic/SocialShareModal.vue
```diff
@@ -0,0 +1,42 @@
...
+
+  export default {
+    name: 'SocialShareModal',
+
+    components: {
+      SocialShare
+    },
+
+    props: {
+      url: {
```
rwd, line 35: Rename this `mediaUrl` to avoid confusion.

components/generic/ShareButton.vue
alansutherland: ok shareButton, socialShare and socialShareModal have been to the new sharing folder

components/generic/SocialShareModal.vue
alansutherland: sure

components/generic/SocialShare.vue
alansutherland: The socialShare icons are already used across the site on the footer, the exhibition and maybe a few other places I'm not aware of, all of which don't show the text, currently the share modal is the only component which renders the text so the withText sense makes more sense than rendering the text everywhere and then hiding it using css on all other components

assets/scss/modal.scss
alansutherland: they can be moved to the component, however, adding the scoped attribute to components scss means this style is not applied to the component. I have moved the style to the socialShareModal component but without the scoped attribute

assets/scss/modal.scss
alansutherland: ok

components/generic/SocialShareModal.stories.js
alansutherland: I don't understand, add the shareOn translation key? 
the links in storybook aren't populated correctly on the socialShare component either, for example the facebook link is https://www.facebook.com/sharer/sharer.php?display=page&u=undefined

components/generic/SocialShareModal.stories.js
alansutherland: Try removing it and then clicking the share button you will the see the error `Cannot read property 'getters' of undefined`

components/generic/SocialShare.vue
alansutherland: thanks, yes this is much better

components/generic/SocialShareModal.stories.js
alansutherland: Checking the SocialShare story it uses the store and canonicalUrl, which I assume worked previously, however, now both this and the modal story are not creating the correct link on storybook.

alansutherland: ok added the shareOn key

components/generic/SocialShareModal.stories.js
alansutherland: the socialShare story uses the canonicalUrl as well. Now socialShare unit tests are failing because example.org is expected but 'undefined' is returned

components/sharing/SocialShare.stories.js
```diff
@@ -11,7 +11,7 @@ const store = new Vuex.Store({
   }
 });
 
-storiesOf('Generic', module)
+storiesOf('Sharing', module)
```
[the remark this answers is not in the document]
mirjamverloop: Forget what I said, you had discussed the current folder/structure with Richard.

components/sharing/ShareButton.vue
```diff
@@ -0,0 +1,31 @@
...
+    text-transform: capitalize;
+    background: $offwhite;
+    color: $mediumgrey;
+    font-size: $font-size-small;
+    border-color: transparent;
+    box-shadow: none;
+    border-radius: 0.25rem;
+    padding: 0.375rem 0.5rem;
+    span {
+      font-size: 1.125rem;
```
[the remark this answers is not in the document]
alansutherland: you're correct they should both be 14px, the download button was 16px, I have reduced it on this PR

components/sharing/SocialShare.vue
```diff
@@ -0,0 +1,97 @@
...
+
+  a {
+    align-items: center;
+    background: transparent;
+    border: 1px solid $black;
+    border-radius: 50%;
+    color: $black;
+    display: inline-flex;
+    height: calc(2.5rem - 1px);
+    justify-content: center;
```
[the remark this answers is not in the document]
alansutherland: good point I had missed that

components/sharing/SocialShare.vue
```diff
@@ -0,0 +1,97 @@
...
+    width: calc(2.5rem - 1px);
+
+    [class^='icon'] {
+      color: $black;
+      font-size: $font-size-medium;
+    }
+
+    &:hover {
+      background: $black;
+
```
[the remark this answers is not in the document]
alansutherland: it's working for me when I test on storybook

components/sharing/SocialShare.vue
```diff
@@ -0,0 +1,97 @@
...
+      },
+      withText: {
+        type: Boolean,
+        default: false
+      }
+    },
+
+    data() {
+      return {
+        networks: [
```
lbiedinger, line 40: This isn't able to get the mapped getter and doesn't work in the data section, if 'networks' is a computed property it'll work properly.
There are also some unit tests failing because of this.

components/sharing/SocialShareModal.stories.js
```diff
@@ -0,0 +1,43 @@
...
+        share: 'share',
+        "shareOn": "Share on {social}"
+      }
+    }
+  }
+};
+
+const store = new Vuex.Store({
+  getters: {
+    'http/canonicalUrl': () => {}
```
lbiedinger, line 21: return a value here 'https://www.example.org/'

components/sharing/SocialShareModal.stories.js
```diff
@@ -0,0 +1,43 @@
...
+storiesOf('Sharing', module)
+  .add('Social share modal', () => ({
+    components: {
+      SocialShareModal,
+      ShareButton
+    },
+    i18n,
+    data() {
+      return {
+        mediaUrl: 'https://www.example.org/'
```
lbiedinger, line 34: Make this:
'https://www.example.org/media'

components/sharing/SocialShareModal.stories.js
```diff
@@ -0,0 +1,43 @@
...
+      };
+    },
+    store,
+    template: `
+      <b-container class="mt-3 p-3 bg-white">
+        <ShareButton />
+        <SocialShareModal media-url="mediaUrl" />
```
lbiedinger, line 41: needs to be:
` <SocialShareModal :media-url="mediaUrl" />`
To resolve mediaUrl, otherwise it comes through as a string literal.

components/sharing/SocialShare.vue
alansutherland: thanks I have moved it to a computed property and unit tests are passing

components/sharing/SocialShareModal.stories.js
alansutherland: as in `'http/canonicalUrl': () => { 'https://www.example.org/' }` ?

components/sharing/SocialShareModal.stories.js
alansutherland: ok

assets/scss/modal.scss
```diff
@@ -27,3 +42,7 @@
     padding: 1rem 0;
```
lbiedinger, line 42: This padding isn't needed anymore now. It's causing extra space in the user form modal `/?path=/story/user-collection--modal`

lbiedinger: maybe the footer only needs padding on the top.

alansutherland: good spot I've removed padding except for top

PR #725, opened 2020-07-16
API request logger & display

rwd:
* Go to /en/debug and you will see a very simple page with a checkbox styled as a toggle and labelled "API requests".
* Switch that on, and you will see a new "Debug" menu appear in the footer underneath the language switcher.
* There's an "API requests" button there that opens a modal showing API requests that have been made for the current page.
* Once you've switched it on, try doing a search, then hit the "API requests" button and you'll see a list of the requests that were made.
* For GET requests, the URL is a link to facilitate inspecting the response; for other method requests, just the URL.
* It's essentially like the browser dev tools network tab, but aimed at our non-developer users still needing to get to upstream API responses.

### DIFF
--- a/assets/scss/modal.scss
+++ b/assets/scss/modal.scss
@@ -26,4 +26,32 @@
     justify-content: space-between;
     padding: 1rem 0;
   }
+  &#api-requests {
+    font-family: $font-family-sans-serif;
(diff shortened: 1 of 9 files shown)


rwd:
Superceded by #744

PR #728, opened 2020-07-20
Low quality warning

alansutherland:


components/generic/LowQualityWarning.stories.js
```diff
@@ -0,0 +1,24 @@
+import { storiesOf } from '@storybook/vue';
+import DownloadButton from './DownloadButton.vue';
```
mirjamverloop, line 2: This is not used in this story, right?

alansutherland: yes, have removed now

PR #729, opened 2020-07-20
Create Likes gallery, add/remove items

annachristaki:


components/generic/ContentCard.vue
```diff
@@ -256,6 +257,33 @@
 
       imageNotFound() {
         this.cardImageUrl = '';
+      },
```
rwd, line 260: This introduces a lot of logic to the component.

Let's instead move the two buttons and all of this accompanying interaction logic into a new component.

plugins/europeana/sets.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+      }
+
+    )
+      .then((response) => {
+        return response.data;
+      });
+  },
+
+  modifyItems(modify, likesId, itemId) {
```
rwd, line 21: Document the parameters in a JSdoc block.

components/account/UserButtons.vue
```diff
@@ -0,0 +1,73 @@
...
+    },
+
+    data() {
+      return {
+        liked: false
+      };
+    },
+    computed: {
+      itemId() {
+        return this.itemUrl.params[0];
```
rwd, line 40: This presumes that the itemUrl is a route object, which is fine in the scenarios we have for the user buttons at present, but means that the prop should only accept object-type values, not strings.

components/account/UserButtons.vue
```diff
@@ -0,0 +1,73 @@
...
+        if (process.browser) {
+          localStorage.setItem('likesId', id);
+        }
+      },
+
+      getLikesId() {
+        return localStorage.getItem('likesId');
+      },
+
+      doLike() {
```
rwd, line 55: Rename this to `toggleLiked`

components/account/UserButtons.vue
```diff
@@ -0,0 +1,73 @@
...
+        this.liked ? this.like(false) : this.like(true);
+      },
+
+      async like(val) {
```
rwd, line 59: Remove the parameter from the function signature and create a second function to `unlike`.

components/account/UserButtons.vue
```diff
@@ -0,0 +1,73 @@
...
+        if (val) {
+          if (!this.getLikesId()) {
+            const response = await this.$galleries.createLikes();
+            this.setLikesId(response.id.split('/').pop());
+          }
+          await this.$galleries.modifyItems('add', this.getLikesId(), this.itemId);
+        } else {
+          await this.$galleries.modifyItems('delete', this.getLikesId(), this.itemId);
+        }
+        this.liked = !this.liked;
```
rwd, line 69: Move this into `toggleLiked` (formerly `doLike`) after it has awaited `like` or `unlike`

plugins/europeana/sets.js
```diff
@@ -0,0 +1,65 @@
+import { config } from './';
```
rwd, line 1: Rename this file to plugins/europeana/set.js to be consistent with the others in the same directory.

store/galleries.js
```diff
@@ -0,0 +1,16 @@
+export const state = () => ({
```
rwd, line 1: Rename this file to store/set.js

store/galleries.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const mutations = {
+  setLikesId(state, value) {
+    state.likesId = value;
+  }
+};
+
+export const getters = {
+  likesId(state) {
+    return state.likesId ? state.likesId : '';
```
rwd, line 13: This getter seems redundant when the state value could be retrieved directly.

PR #731, opened 2020-07-21
Title description box

lbiedinger:
New summary info component for the item page.
With storybook story and some unit tests.

### DIFF
--- a/components/item/SummaryInfo.vue
+++ b/components/item/SummaryInfo.vue
@@ -1,8 +1,7 @@
 <template>
-  <div class="card mb-3  px-4">
+  <div class="card p-4">
     <header
       v-if="titles"
-      class="pt-4"
     >
       <template
         v-for="(heading, index) in titles"
@@ -11,6 +10,7 @@
           v-if="index === 0"
           :key="index"
           :lang="heading.code"
+          class="mb-0"
         >
           {{ heading.value }}
         </h1>
@@ -18,7 +18,7 @@
           v-else
           :key="index"
           :lang="heading.code"
-          class="font-weight-bold"
+          class="font-weight-bold my-3"
         >
           {{ heading.value }}
         </p>
@@ -51,7 +51,7 @@
       <b-button
         v-if="expandableDescription"
         data-qa="description show link"
-        class="btn-link is-size-4 p-0 mb-4"
+        class="btn-link is-size-4 p-0 mt-2"
         variant="link"
         @click="toggleMoreDescription"
       >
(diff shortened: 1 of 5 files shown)


components/item/SummaryInfo.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <div class="card mb-3 ">
+    <header
+      v-if="titles"
+      class="card-heading px-4 pt-4"
```
mirjamverloop, line 5: I think the "card-heading" class is no longer needed in this new structure. The class currently helps to organise title(s), image and description in the card.

components/item/SummaryInfo.vue
```diff
@@ -0,0 +1,122 @@
...
+          :lang="heading.code"
+          class="font-weight-bold"
+        >
+          {{ heading.value }}
+        </p>
+      </template>
+    </header>
+    <div
+      v-if="description"
+      class="description px-4"
```
mirjamverloop, line 29: Perhaps give the card a padding "p-4", instead of spacing classes on this description div and the header.

components/item/SummaryInfo.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <div class="card mb-3 ">
+    <header
+      v-if="titles"
+      class="card-heading px-4 pt-4"
+    >
+      <template
+        v-for="(heading, index) in titles"
+      >
+        <h1
```
mirjamverloop, line 10: The main title in the design has a font-weight of normal, here it is bold (600). I haven't found any styling for subtitles in Zeplin. 

The title is also slightly smaller in the design, 28px.

components/item/SummaryInfo.vue
```diff
@@ -0,0 +1,122 @@
...
+    color: $innovationblue;
+    text-decoration: underline;
+    text-transform: uppercase;
+    font-size: 16px;
+    font-weight: 600;
+
+    &:hover {
+      text-decoration: none;
+    }
+  }
```
mirjamverloop, line 120: In the design, this button is styled slightly different. For example, no uppercase, no underlining.

components/item/SummaryInfo.vue
```diff
@@ -0,0 +1,126 @@
...
+<style lang="scss">
+  @import './assets/scss/variables.scss';
+
+  h1 {
+    font-size: 1.75rem;
+    font-weight: normal;
+  }
+
+  .btn-link {
+    color: $innovationblue;
```
alansutherland, line 116: this can just be $blue

lbiedinger:
> * I can’t see an example on Zeplin where the subtitles are included, has this been added just in case?

Indeed. Using the sub-title styling as is on the current page as there's no other design provided.


I've update the paddings and margins so as to be more consistent when various combinations of elements are present or not.

PR #733, opened 2020-07-21
Client baseURL lookup in i18n baseUrl option function

lbiedinger:


lbiedinger:
This issue has been addressed via #734

PR #735, opened 2020-07-22
collection page header title fix

alansutherland:


pages/collections/_type/_.vue
```diff
@@ -241,6 +241,14 @@
...
           ? this.page.relatedLinksCollection.items : null;
       },
+      relatedCollectionsFound() {
+        if (this.relatedEntities !== null && this.relatedEntities.length > 0) {
+          return true;
+        } else if (this.relatedCollectionCards !== null && this.relatedCollectionCards.length > 0) {
+          return true;
+        }
+        return false;
+      },
```
mirjamverloop, line 251: Perhaps you can use the optional chaining operator? Not sure, if we use that elsewhere already. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

`!this.relatedEntities?.length`

alansutherland: thanks I have implemented this

pages/collections/_type/_.vue
```diff
@@ -241,6 +241,14 @@
           && this.page.relatedLinksCollection.items.length > 0)
           ? this.page.relatedLinksCollection.items : null;
       },
+      relatedCollectionsFound() {
+        if (this.relatedEntities?.length > 0) {
+          return true;
+        } else if (this.relatedCollectionCards?.length > 0) {
+          return true;
+        }
```
mirjamverloop, line 249: I guess you can even combine these. 
`if (this.relatedEntities?.length > 0 || this.relatedCollectionCards?.length > 0) {`

lbiedinger:
@alansutherland 
What about:
```
      relatedCollectionsFound() {
        if (this.relatedEntities && this.relatedEntities.length > 0) {
          return true;
        } else if (this.relatedCollectionCards && this.relatedCollectionCards.length > 0) {
          return true;
        }
        return false;
      },
```
I'd say combine the two checks with `||` as suggested by Mirjam, but without the chaining operator it would make for a very long line. So that's more readable.

PR #742, opened 2020-07-28
Update Media Player

alansutherland:


pages/media/index.vue
```diff
@@ -51,6 +53,7 @@
         script: [
           { src: `https://code.jquery.com/jquery-${this.JQUERY_VERSION}.min.js` },
           { src: `https://code.jquery.com/ui/${this.JQUERY_UI_VERSION}/jquery-ui.min.js` },
+          { src: this.dashRequired ? `https://unpkg.com/dashjs@${this.DASHJS_VERSION}/dist/dash.all.min.js` : '' },
```
lbiedinger, line 56: This results in an empty <src> tag being added to the markup.
Try this instead:
```
        script: [
          { src: `https://code.jquery.com/jquery-${this.JQUERY_VERSION}.min.js` },
          { src: `https://code.jquery.com/ui/${this.JQUERY_UI_VERSION}/jquery-ui.min.js` },
          { src: `https://unpkg.com/@europeana/media-player@${this.MEDIA_PLAYER_VERSION}/dist/europeana-media-player.min.js` }
        ].concat(this.dashRequired ? [
          { src: `https://unpkg.com/dashjs@${this.DASHJS_VERSION}/dist/dash.all.min.js` }
        ] : [])
```

plugins/media.js
```diff
@@ -73,3 +73,7 @@ export function isRichMedia(media) {
     isIIIFMedia(media) ||
     isPlayableMedia(media);
 }
+
+export function requiresDashJS(media) {
```
lbiedinger, line 77: Could you add a unit test?
Should be straight forward, there are already similar ones in `tests/unit/plugins/media.spec.js` for the rest of the methods in here.

pages/media/index.vue
alansutherland: thanks I knew there was better way of doing this, just couldn't think of it

plugins/media.js
alansutherland: sure

components/item/MediaPresentation.vue
```diff
@@ -17,7 +17,7 @@
       <iframe
         data-qa="media player"
         allowfullscreen="true"
-        :src="$path({ name: 'media', query: { id: europeanaIdentifier, mediaUrl: media.about } })"
+        :src="$path({ name: 'media', query: { id: europeanaIdentifier, mediaUrl: media.about, dashRequired: requiresDashJS } })"
```
rwd, line 20: Remove this and replace with `mediaType: media.ebucoreHasMimeType`. Then make the media page responsible for performing the check against `requiresDashJS`.

alansutherland: ok thanks, have updated

PR #743, opened 2020-07-29
Big swipeable images

alansutherland:


components/item/AwesomeSwiper.stories.js
```diff
@@ -0,0 +1,51 @@
...
+    template: `
+      <div class="mt-3">
+        <AwesomeSwiper
+          :europeana-identifier="id"
+          :media="media"
+        />
+      </div>
+    `
+  }))
+  .add('Full height images', () => ({
```
lbiedinger, line 34: let's get rid of this as it's just the same as the other example at the moment.

components/item/AwesomeSwiper.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <swiper
+    ref="awesome"
+    class="swiper"
```
lbiedinger, line 4: we're not using this class, but might be useful to keep around as we start applying more custom styling ourselves.

components/item/AwesomeSwiper.vue
```diff
@@ -0,0 +1,110 @@
...
+    },
+    mounted() {
+      // console.log('Current Swiper instance object', this.swiper)
+      // this.swiper.slideTo(3, 1000, false)
+    }
+  };
+</script>
+
+<style lang="scss" scoped>
+  /* .swiper-slide {
```
lbiedinger, line 86: let's get rid of the unused css.

pages/item/_.vue
```diff
@@ -351,6 +351,8 @@
           this.selectedMedia.about = msg.data.id;
         }
       });
+      console.log(this.media);
```
lbiedinger, line 354: this can go.

components/item/AwesomeSwiper.stories.js
alansutherland: ok

components/item/AwesomeSwiper.vue
alansutherland: sure, this was on the example but I didn't see the purpose of it so had commented out, will delete

pages/item/_.vue
alansutherland: ok

PR #747, opened 2020-07-31
disable inactive cards, click to focus

alansutherland:


components/item/AwesomeSwiper.vue
```diff
@@ -59,6 +60,7 @@
           slidesPerView: 'auto',
           spaceBetween: 40,
           centeredSlides: true,
+          slideToClickedSlide: true,
```
mirjamverloop, line 63: Is this a swiper option? Or is it used?

components/item/AwesomeSwiper.vue
```diff
@@ -74,30 +76,43 @@
...
+      position: absolute;
+    }
     a {
       display: inline-flex;
       height: 100%;
       align-items: center;
     }
+    &:last-of-type {
+      margin-right: 50vw !important;
+    }
```
mirjamverloop, line 111: As discussed, could you remove this for now?

alansutherland: ok

components/item/AwesomeSwiper.vue
alansutherland: yes it's a built-in option which achieves this requirement: “When a resource is not the current focus image, clicking it brings it to the focus”

mirjamverloop: Thanks for clarifying!

PR #748, opened 2020-08-04
Swipeable carousel: media info

alansutherland:


components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+    props: {
+      media: {
+        type: Array,
+        default: () => []
+      }
+    },
+    data() {
+      return {
+        mediaUrl: 'https://www.example.org/media',
+        id: '/2020601/https___1914_1918_europeana_eu_contributions_10265',
```
lbiedinger, line 48: This'll need to be passed as a prop.

components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+    },
+    props: {
+      media: {
+        type: Array,
+        default: () => []
+      }
+    },
+    data() {
+      return {
+        mediaUrl: 'https://www.example.org/media',
```
lbiedinger, line 47: default `mediaUrl` to an empty string.

components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+        id: '/2020601/https___1914_1918_europeana_eu_contributions_10265',
+        rightsStatementIsUrl: true,
```
lbiedinger, line 49: default `rightsStatementIsUrl`  to false

components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+        disabled: false,
```
lbiedinger, line 50: I'd say get rid of the `disabled` property for now. We'll need to add it again later, but there's additional work here for doing this lookup properly and then also adding another notice to the page. For now it doesn't do anything and when it's added with functionality it should be named something more descriptive.

components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+        selectedMediaItem: null,
+        target: '_blank'
```
lbiedinger, line 52: As target defaults to '_blank' on the button anyway and there's no way to change it here, just remove this and don't pass anything to the button component.

components/item/AwesomeSwiper.vue
```diff
@@ -76,6 +77,11 @@
       swiper() {
         return this.$refs.awesome.$swiper;
       }
+    },
+    methods: {
+      onSlideChange() {
```
lbiedinger, line 82: This could do with a unit test.

components/item/ItemHero.vue
```diff
@@ -0,0 +1,77 @@
...
+      },
+      rightsStatement() {
+        return this.selectedMedia.rightsStatement;
+      },
+      downloadUrl() {
+        return this.selectedMedia.downloadUrl;
+      }
+    },
+    methods: {
+      selectMedia(about) {
```
lbiedinger, line 72: This also should have a unit test.
Ensure that when it's called the `selectedMedia` of the component changes.

components/item/AwesomeSwiper.stories.js
```diff
@@ -1,15 +1,61 @@
...
-  { thumbnails: { large: 'https://api.europeana.eu/api/v2/thumbnail-by-url.json?size=w400&type=IMAGE&uri=https%3A%2F%2Feuropeana1914-1918.s3.amazonaws.com%2Fattachments%2F119639%2F10265.119639.original.jpg' } },
-  { thumbnails: { large: 'https://api.europeana.eu/api/v2/thumbnail-by-url.json?size=w400&type=IMAGE&uri=https%3A%2F%2Feuropeana1914-1918.s3.amazonaws.com%2Fattachments%2F119640%2F10265.119640.original.jpg' } }
+  {
+    about: 'https://europeana1914-1918.s3.amazonaws.com/attachments/119112/10265.119112.original.jpg',
+    thumbnails: { large: 'https://api.europeana.eu/api/v2/thumbnail-by-url.json?size=w400&type=IMAGE&uri=https%3A%2F%2Feuropeana1914-1918.s3.amazonaws.com%2Fattachments%2F119112%2F10265.119112.original.jpg' },
+    rightsStatement: 'http://creativecommons.org/licenses/by-sa/3.0/',
+    downloadUrl: 'https://proxy.europeana.eu/2020601/https___1914_1918_europeana_eu_contributions_10265?view=https%3A%2F%2Feuropeana1914-1918.s3.amazonaws.com%2Fattachments%2F119112%2F10265.119112.original.jpg&api_url=https%3A%2F%2Fapi.europeana.eu%2Fapi'
```
lbiedinger, line 33: The downloadUrl should be derived from the `about`, see the method in the MediaActionBar component.

components/item/ItemHero.vue
alansutherland: ok

components/item/ItemHero.vue
alansutherland: ok

components/item/ItemHero.vue
alansutherland: yes good point

components/item/ItemHero.vue
alansutherland: ok removed just passing downloadUrl to the button

components/item/AwesomeSwiper.stories.js
alansutherland: ok copied the method from MediaActionBar

components/item/ItemHero.vue
alansutherland: ok

PR #749, opened 2020-08-04
Swipeable mirador

lbiedinger:


assets/scss/media.scss
```diff
@@ -0,0 +1,34 @@
...
+    @media (min-width: $bp-extralarge) {
+      max-width: 1140px;
+    }
+  }
+  iframe {
+    height: 100%;
+    width: 100%;
+    border: 1px solid #a6a6a6;
+    border-radius: 0.25rem;
+    box-shadow: 0 1px 3px 0 rgba(0, 0, 0, 0.25);
```
alansutherland, line 32: use variable $boxshadow-small

assets/scss/media.scss
```diff
@@ -0,0 +1,34 @@
...
+      max-width: 720px;
+    }
+
+    @media (min-width: $bp-large) {
+      max-width: 960px;
+    }
+
+    @media (min-width: $bp-extralarge) {
+      max-width: 1140px;
+    }
```
mirjamverloop, line 25: Okay, a thought: if there is just one slide, the slide can have a width of 100%. (When smaller, the media will be centred). 

In that case, I think you don't need the iiif-frame-wrapper class. You could make use of the existing "container" class. That is a Bootstrap class and handles the width exactly as you have described here.  And you could use the "h-100" class to set the height.

components/item/AwesomeSwiper.stories.js
```diff
@@ -31,3 +46,46 @@ storiesOf('Item page/Awesome Swiper', module)
       </div>
     `
   }));
+// TODO: This story doesn't work as the IIIF player makes use of an iframed page.
```
mirjamverloop, line 49: Then perhaps just remove?

assets/scss/media.scss
lbiedinger: We don't want the viewer to be so wide. It leads to strange aspect ratios. Mainly because of the max height of 568px on the whole swiper.

lbiedinger: I misunderstood your comment initially. I had another look though and yes, that works and is simpler.

PR #750, opened 2020-08-05
Display user collections

annachristaki:


plugins/europeana/set.js
```diff
@@ -12,15 +14,96 @@ export default ($axios) => ({
...
+ * @return {Object[]} the initial sets with a thumbnail property added for each set that contains at least one item
+ */
+  async getSetImages(sets) {
+    let q = sets.map(e => e.firstItem).filter(Boolean).join('" OR "');
+    if (!q) {
+      return sets;
+    }
+    const origin = config.record.origin;
+    const path = config.record.path;
+    return axios.get(`${origin}${path}/search.json`, {
```
rwd, line 82: This should instead use `search` exported from plugins/europeana/search.js

pages/account/index.vue
```diff
@@ -54,15 +64,39 @@
...
+    },
+    fetchOnServer: false,
     computed: {
       loggedInUser() {
         return this.$store.state.auth.user;
       }
     },
-
+    methods: {
+      refresh() {
```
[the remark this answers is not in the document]
annachristaki: It's not, sorry for insufficient cleanup

plugins/europeana/set.js
```diff
@@ -12,15 +14,96 @@ export default ($axios) => ({
...
+ */
+  async getAllSets(ids) {
+    return axios.all(ids.map(id => $axios.get(setApiUrl('/') + id + '?profile=standard')))
+      .then(responseArray => {
+        return responseArray.map(set => {
+          return {
+            ...{
+              id: set.data.id.split('/').pop(),
+              title: set.data.title ? set.data.title['en'] : '',
+              description: set.data.description ? set.data.description['en'] : 'No description',
```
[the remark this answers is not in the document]
annachristaki: I was following the [design](https://app.zeplin.io/project/5e3c1167d638ca9a6327895b/screen/5ecfb636bf41de48752600a5) comments on this.

pages/account/index.vue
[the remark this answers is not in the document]
rwd: This method is still on the branch.

plugins/europeana/set.js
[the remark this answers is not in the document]
rwd: Understood. I think it better to display nothing, but will take it to the design team to confirm.

plugins/europeana/set.js
```diff
@@ -12,15 +14,81 @@ export default ($axios) => ({
...
+      });
+  },
+
+  /**
+ * Get all set metadata for the provided set ids
+ * @param {Array} ids the ids of sets to retrieve
+ * @return {Object[]} an array of sets
+ */
+  async getAllSets(ids) {
+    return axios.all(ids.map(id => $axios.get(setApiUrl('/') + id + '?profile=standard')))
```
rwd, line 52: This can in fact just use `Promise.all`, removing the need to import axios.

plugins/europeana/set.js
```diff
@@ -12,15 +14,81 @@ export default ($axios) => ({
...
+    if (!q) {
+      return sets;
+    }
+    const results =  await search(
+      {
+        query: `europeana_id:("${q}")`
+      });
+
+    sets.forEach((set, index) => {
+      let result = results.results.find(res => res.europeanaId === set.firstItem);
```
rwd, line 84: Use const instead of let

plugins/europeana/set.js
```diff
@@ -12,15 +14,81 @@ export default ($axios) => ({
...
+      if (result) {
```
rwd, line 85: To avoid needless object creation, replace with:
```js
if (result) sets[index].thumbnail = result.edmPreview;
```

components/account/UserSets.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <b-container data-qa="user sets">
```
lbiedinger, line 2: This data-qa attribute is overwritten when the component is used in the page.
It's also not used for any tests as far as I see, so I don't think it's needed.

pages/account/index.vue
```diff
@@ -30,18 +30,28 @@
               data-qa="public collections"
               :title="$t('account.publicCollections')"
             >
-              <!-- TODO: Update this section to preview the results retrieved with Sets API -->
-              <div class="text-center p-3">
-                Placeholder for {{ $t('account.publicCollections') }} tab.
+              <div v-if="!$fetchState.pending">
```
lbiedinger, line 33: The content in these tabs should have 40px distance from the tabs above.
There's some space in here now, but it's too little.

PR #751, opened 2020-08-05
Searchbar on desktop

mirjamverloop:


tests/features/collections/newspaper.feature
```diff
@@ -45,10 +45,3 @@ Feature: Newspapers collection
     And I go to page number 2
     And I wait for the page to load
     Then I should be on `/en/collections/topic/18-newspapers?page=2&view=grid&api=metadata`
-
-  Scenario: Newspapers collection API toggle is removed by removing search pill
```
lbiedinger, line 49: This should be replaced by a test that checks the removal by using the 'Search for "query" in our entire Collection' option.

components/search/SearchQueryOptions.vue
```diff
@@ -0,0 +1,434 @@
...
+    <b-list-group-item
+      v-if="showLoader && isLoading"
+      class="loading"
+    >
+      {{ $t('loadingResults') }}{{ $t('formatting.ellipsis') }}
+    </b-list-group-item>
+
+    <template v-else>
+      <template
+        v-if="pillLabel"
```
lbiedinger, line 20: `pillLabel` doesn't make any sense as a variable name anymore. It should be renamed to 'entityCollectionLabel'

components/search/SearchQueryOptions.vue
```diff
@@ -0,0 +1,434 @@
...
+    role="listbox"
+    :aria-label="$t('searchSuggestions')"
+    :aria-hidden="!isActive"
+  >
+    <b-list-group-item
+      v-if="showLoader && isLoading"
+      class="loading"
+    >
+      {{ $t('loadingResults') }}{{ $t('formatting.ellipsis') }}
+    </b-list-group-item>
```
lbiedinger, line 16: This doesn't seem to be used anywhere. 'showLoader' defaults to 'false' and isn't  changed at any point.

components/search/SearchQueryOptions.vue
alansutherland: yep good point

components/search/SearchQueryOptions.vue
alansutherland: true, I've removed, good to simplify the component more

tests/features/collections/newspaper.feature
alansutherland: I can't find this scenario

lbiedinger: Something like this I mean:
```
 Scenario: Newspapers collection API toggle is removed by searching in the entire collection
    Given I am on `/en/collections/topic/18-newspapers?api=fulltext&view=grid`	
    When I click the `show search button`	
    And I enter "paris" into the `search box`
    And I click the `search entire collection button`	
    Then I see the `search page`	
    And I should be on `/en/search?page=1&view=grid&query=paris`
```
Might need a bit of tweaking for the order of the search params on the final page, but the important thing is that it doesn't contain 'api=fulltext'.

lang/en.js
```diff
@@ -515,6 +517,7 @@ export default {
   "results": "Results",
   "resultsLimitWarning": "Additional results are not shown as only the first 1000 most relevant results are shown. If you haven't found what you're looking for, please consider refining your search.",
   "search": "Search",
+  "searchResults": "Search results for ",
```
lbiedinger, line 520: use interpolation here like so:
```
"Search results for '{0}'",
```
And pass the search term into the string.

lbiedinger (changes requested):

lang/en.js
alansutherland: ok

tests/features/collections/newspaper.feature
alansutherland: ok

components/search/SearchForm.vue
```diff
@@ -13,29 +13,9 @@
       :aria-expanded="isAutoSuggestActive"
       class="auto-suggest"
     >
-      <template
-        v-if="pillLabel"
-        v-slot:prepend
-      >
-        <SearchBarPill
```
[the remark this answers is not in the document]
alansutherland: yep good point

components/search/SearchForm.vue
```diff
@@ -284,7 +194,7 @@
       suggestionLinkGen(suggestion) {
         const query = {
           view: this.view,
-          query: `"${suggestion}"`
+          query: `${suggestion}`
```
lbiedinger, line 197: This still needs to be escaped, otherwise any of the suggestions with parentheses won't return results.
For example type 'artist' and select one of the options that contain '[NAME] (artist)'

lbiedinger (changes requested):

components/search/SearchForm.vue
alansutherland: I hadn't realised that, I see your point. I have added quotes back in to escape suggestions.

PR #756, opened 2020-08-07
View liked items in search results

annachristaki:


store/set.js
```diff
@@ -1,9 +1,41 @@
...
+  like(state, value) {
+    state.liked.push(value);
+  },
+  unlike(state, value) {
+    state.liked.splice(state.liked.indexOf(value), 1);
+  }
+};
+
+export const actions = {
+  async like({ commit, state }, { action, itemId }) {
```
rwd, line 24: Split this into two actions: `like` and `unlike`

components/account/UserButtons.vue
```diff
@@ -38,37 +38,29 @@
...
-          return this.$store.state.set.likesId;
-        },
-        set(value) {
-          this.$store.commit('set/setLikesId', value);
-        }
+      likesId() {
+        return this.$store.state.set.likesId;
       }
     },
+    created() {
```
rwd, line 45: Move this into fetch and delete the method `setLiked` that's only used once:
```js
    fetch() {
      this.liked = this.$store.state.set.liked.includes(this.itemId);
    },
```

PR #757, opened 2020-08-07
Single item styling & behaviour

lbiedinger:


components/item/AwesomeSwiper.vue
```diff
@@ -76,6 +77,11 @@
       swiper() {
         return this.$refs.awesome.$swiper;
       }
+    },
+    methods: {
+      swipingEnabled() {
```
rwd, line 82: This needn't be a method. Just put it into `data`.

components/item/AwesomeSwiper.vue
```diff
@@ -10,12 +10,23 @@
...
+        />
+      </div>
       <MediaCard
+        v-else
         :europeana-identifier="item.europeanaIdentifier"
         :media="item"
       />
     </swiper-slide>
     <div
+      v-if="swipingEnabled"
```
mirjamverloop, line 29: Is this perhaps a leftover? I don't see any other mention of `swipingEnabled` in this file and there is a "Property or method "swipingEnabled" is not defined on the instance but referenced during render..." warning.

components/item/AwesomeSwiper.vue
```diff
@@ -10,12 +10,23 @@
       v-for="(item, index) in displayableMedia"
       :key="index"
     >
+      <div
+        v-if="singleMediaResource"
+        class="container h-100"
```
mirjamverloop, line 15: I am not sure if it is due to this `h-100`, but on mobile, single images don't keep their aspect ratio. I think they should. Try with item, /2020738/3DICONS_8432, the image is a square, but on mobile not.

PR #760, opened 2020-08-10
New item page: alternative page + grid

mirjamverloop:
- Alternative page, replace `item` in url with `item-new`
- New grid, with some components
- Todos were new components have to be added, or to indicate any other change

### DIFF
--- a/pages/item-new/_.vue
+++ b/pages/item-new/_.vue
@@ -64,6 +64,7 @@
         </b-row>
         <b-row class="mb-3">
           <b-col>
+            <!-- TODO: remove rounded-0 class in MetadataBox component -->
             <MetadataBox
               :all-metadata="allMetaData"
               :core-metadata="coreFields"
@@ -106,8 +107,6 @@
   import { getRecord, similarItemsQuery } from '../../plugins/europeana/record';
   import { search } from '../../plugins/europeana/search';
 
-  // TODO: add isIIIFPresentation if necessary for swiper
-  import { isRichMedia } from '../../plugins/media';
   import { langMapValueForLocale } from  '../../plugins/europeana/utils';
   import { findEntities } from '../../plugins/europeana/entity';
   import { search as searchAnnotations } from '../../plugins/europeana/annotation';
@@ -169,7 +168,6 @@
         isShownAt: null,
         media: [],
         relatedEntities: [],
-        selectedMediaItem: null,
         similarItems: [],
         annotations: [],
         taggingAnnotations: [],
@@ -238,31 +236,6 @@
         if (!this.descriptionInCurrentLanguage) return '';
         return this.descriptionInCurrentLanguage.values[0] ? this.descriptionInCurrentLanguage.values[0] : '';
       },
-      isRichMedia() {
-        return isRichMedia(this.selectedMedia);
-      },
-      selectedMedia: {
-        get() {
-          return this.selectedMediaItem || this.media[0] || {};
-        },
-        set(about) {
-          this.selectedMediaItem = this.media.find((item) => item.about === about) || {};
-        }
-      },
-      selectedMediaImage() {
-        if (!this.selectedMedia.thumbnails) return {};
-        return {
-          src: this.selectedMedia.thumbnails.large,
-          link: this.isShownAt
-        };
-      },
-      edmRights() {
-        return this.selectedMedia.webResourceEdmRights ? this.selectedMedia.webResourceEdmRights : this.fields.edmRights;
-      },
-      rightsStatement() {
-        if (this.edmRights) return langMapValueForLocale(this.edmRights, this.$i18n.locale).values[0];
-        return false;
-      },
       dataProvider() {
         const edmDataProvider = langMapValueForLocale(this.coreFields.edmDataProvider, this.$i18n.locale);
 
@@ -302,10 +275,6 @@
     methods: {
       annotationsByMotivation(motivation) {
         return this.annotations.filter(annotation => annotation.motivation === motivation);
-      },
-
-      selectMedia(about) {
-        this.selectedMedia = about;
       },
 
       getSimilarItems() {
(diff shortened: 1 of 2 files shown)


pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
+<template>
+  <div>
```
lbiedinger, line 2: Would it make sense to add a css class here for the purpose of keeping all the new styling together & potentially away from interfering with the current layout?

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+              Useful data:
+
+              :entities="relatedEntities"
+              data-qa="related entities"
+            -->
+          </b-col>
+        </b-row>
+        <b-row class="mb-3">
+          <b-col>
+            <MetadataBox
```
lbiedinger, line 67: The metadatbox will be getting rounded corners again.

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+  import { mapGetters } from 'vuex';
+
+  import ClientOnly from 'vue-client-only';
+  import MetadataBox from '../../components/item/MetadataBox';
+
+  import { getRecord, similarItemsQuery } from '../../plugins/europeana/record';
+  import { search } from '../../plugins/europeana/search';
+
+  // TODO: add isIIIFPresentation if necessary for swiper
+  import { isRichMedia } from '../../plugins/media';
```
lbiedinger, line 110: isRichMedia can be removed. The heroItem will be handling any of this logic.

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+        return this.titlesInCurrentLanguage[0] ? this.titlesInCurrentLanguage[0].value : this.$t('record.record');
+      },
+      metaDescription() {
+        if (!this.descriptionInCurrentLanguage) return '';
+        return this.descriptionInCurrentLanguage.values[0] ? this.descriptionInCurrentLanguage.values[0] : '';
+      },
+      isRichMedia() {
+        return isRichMedia(this.selectedMedia);
+      },
+      selectedMedia: {
```
lbiedinger, line 244: All this logic for selectedMedia has been moved into the ItemHero component too. It's still needed for some of the metadata here, but I don't believe this to actually be valuable. There's no point in changing the og metadata for the image when a user changes the image as when shared whatever service the item is shared to will simply retrieve the page again, thereby selecting the default media for the og metadata.

We will need to have the page be aware of when download has been disabled for the current media, but that hasn't been implemented on the ItemHero yet either.

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+      if (process.browser) {
+        this.$gtm.push({
+          itemCountry: langMapValueForLocale(this.fields.edmCountry, 'en').values[0],
+          itemDataProvider: langMapValueForLocale(this.coreFields.edmDataProvider, 'en').values[0],
+          itemProvider: langMapValueForLocale(this.fields.edmProvider, 'en').values[0],
+          itemRights: langMapValueForLocale(this.fields.edmRights, 'en').values[0]
+        });
+      }
+
+      // TODO: remove or is this useful for the new swiper/action?
```
lbiedinger, line 293: Indeed, this'll need to be moved to the itemHero component. It wasn't ready to be picked up on that PR as the IIIF viewer hadn't been added to the swiper then.

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+        }
+      });
+    },
+
+    methods: {
+      annotationsByMotivation(motivation) {
+        return this.annotations.filter(annotation => annotation.motivation === motivation);
+      },
+
+      selectMedia(about) {
```
lbiedinger, line 307: won't be needed

pages/item-new/_.vue
```diff
@@ -0,0 +1,369 @@
...
+
+    head() {
+      return {
+        title: this.metaTitle,
+        meta: [
+          { hid: 'title', name: 'title', content: this.metaTitle },
+          { hid: 'description', name: 'description', content: this.metaDescription },
+          { hid: 'og:title', property: 'og:title', content: this.metaTitle },
+          { hid: 'og:description', property: 'og:description', content: this.metaDescription },
+          { hid: 'og:image', property: 'og:image', content: this.selectedMediaImage.src ? this.selectedMediaImage.src : '' },
```
lbiedinger, line 353: See my comment above about setting the selectedMedia for og metadata.

Until we allow users to load the page with a specific media resource selected at page load time(preserved in the URL so it is transparent for bots), this is useless I believe.

pages/item-new/_.vue
mirjamverloop: I am not sure if that is necessary. If we need specific CSS, it can be scoped here without breaking the old item page. Did you have a specific use case in mind?

pages/item-new/_.vue
mirjamverloop: Ah, I have missed that memo. I have added some temp styling and a todo.

pages/item-new/_.vue
```diff
@@ -350,7 +319,8 @@
           { hid: 'description', name: 'description', content: this.metaDescription },
           { hid: 'og:title', property: 'og:title', content: this.metaTitle },
           { hid: 'og:description', property: 'og:description', content: this.metaDescription },
-          { hid: 'og:image', property: 'og:image', content: this.selectedMediaImage.src ? this.selectedMediaImage.src : '' },
+          // TODO: update the og:image
+          { hid: 'og:image', property: 'og:image', content: '' },
```
lbiedinger, line 323: content could be set to `this.media[0].src : ''`

pages/item-new/_.vue
lbiedinger: The rounded corners on the metadata box was one example, but the way you've done it by adding the styling to this page only will work fine too.

PR #768, opened 2020-08-12
Item page: title, description and related collections

alansutherland:


pages/item-new/_.vue
```diff
@@ -35,31 +35,23 @@
...
+            <SummaryInfo
+              :description="descriptionInCurrentLanguage"
+              :titles="titlesInCurrentLanguage"
+            />
           </b-col>
         </b-row>
-        <b-row class="mb-3">
+        <b-row
+          v-if="relatedEntities && relatedEntities.length > 0"
+        >
```
mirjamverloop, line 49: When there are related entities, the whitespace looks okay. But when there aren't any, for example http://localhost:3000/en/item-new/327/_00000070p, it could perhaps use some extra whitespace between title/description-box and metadata. What do you think?

pages/item-new/_.vue
```diff
@@ -35,31 +35,23 @@
       -->
         <client-only />
       </b-container>
-      <b-container>
+      <b-container class="mw-1024">
```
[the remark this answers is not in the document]
mirjamverloop: As discussed, remove this and use Bootstrap grid instead. Something like: 

```
<b-row class="mb-3" align-h="center">
          <b-col cols="12" md="10">
```

Or perhaps, it should be lg="10", see what works best on different screen sizes.

assets/scss/texts.scss
```diff
@@ -16,4 +16,5 @@ header .col {
   font-size: $font-size-small;
   font-weight: 600;
   line-height: 1;
+  margin-top: 1rem;
```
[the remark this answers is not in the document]
alansutherland: I wanted to use bootstrap classes on the heading but the margin is 32px and `.mt-4`  = 24px and `.mt-5` = 48px. I've moved the styling to the `.related-collections` scoped within this page

components/generic/RelatedCollections.vue
```diff
@@ -1,10 +1,10 @@
 <template>
-  <b-container
+  <div
     v-if="relatedCollections.length > 0"
     data-qa="related collections"
     class="related-collections"
   >
-    <h2 class="related-heading text-uppercase mt-4 mb-2">
+    <h2 class="related-heading text-uppercase mb-2">
```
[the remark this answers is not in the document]
alansutherland: reverted

pages/item-new/_.vue
alansutherland: ok, I didn't think it was too bad but I've add in a buffer row if there's no related
```
<b-row
    v-else
    class="mb-3"  
></b-row>
```

pages/item-new/_.vue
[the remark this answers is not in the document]
alansutherland: switched to 10 col grid as discussed, width is down to 950px which is a bit lower than 1024 but design said they preferred it smaller

components/generic/RelatedCollections.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <b-container
+  <div
```
lbiedinger, line 2: Looks like this is squishing the related collections on the search page to the right side.
Example: /en/search?query="Mario Testino"

alansutherland: ok fixed this

components/item/SummaryInfo.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <div class="card rounded-0 border-0 p-4">
     <header
-      v-if="titles"
+      v-if="titles.length > 0"
+      class="mb-3"
```
lbiedinger, line 5: This leaves too large a gap when there is only a title.
Can be seen in the storybook story for 'Only title'

alansutherland: OK I've moved the styling to texts & button-links in assets/scss and added a wrapping class .info-panel, the component isn't jumping around anymore. 

The issue with there being too large a gap on pages with only a title was a little hard since the gap is needed and just adding it to the .description doesn't work on pages with no title. But this css works in all cases, only applying the gap if there's a header
```
header + .description {
    margin-top: 1rem;
 }
```

components/generic/RightsStatementButton.vue
```diff
@@ -36,6 +36,6 @@
     color: $mediumgrey;
     background: $offwhite;
     border-radius: 0.25rem;
-    padding: 0.625rem;
+    padding: 0.38rem 0.625rem;
```
mirjamverloop, line 39: The rights statement button has a slightly larger font-size (16px) than the share and download button (14px). Could you give the rights button the same size?

pages/item-new/_.vue
```diff
@@ -17,7 +17,7 @@
     >
       <b-container
         fluid
-        class="bg-white mb-3"
+        class="bg-white mb-3 px-0"
       >
         <!-- TODO: add swiper here
```
mirjamverloop, line 22: Could you remove the todos as it is now done?

components/item/ItemHero.vue
```diff
@@ -61,7 +61,7 @@
         return this.$proxyMedia(this.selectedMedia.about, this.identifier);
       },
       rightsStatement() {
-        return this.selectedMedia.rightsStatement;
+        return this.selectedMedia.webResourceEdmRights.def[0];
```
mirjamverloop, line 64: There are item pages that don't show, because of an error. I think it has to do with the def here (or the one in downloadEnabled), but haven't looked at it in more detail.

Look at item-new/2048208/c_pca_28759 for example.

components/item/ItemHero.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <div>
+  <div class="item-hero">
     <AwesomeSwiper
       :europeana-identifier="identifier"
       :media="media"
       @select="selectMedia"
     />
-    <b-container class="d-flex justify-content-around mt-5">
+    <b-container class="d-flex justify-content-around media-bar">
```
mirjamverloop, line 8: When you look at an item page with all three buttons: rights, download and share, there is space between them. According to the design, the download and share should be on the right and the rights statement on the left. 

An example: item-new/117/HKMS000005_00000ekd

pages/item-new/_.vue
```diff
@@ -33,46 +33,52 @@
...
-        <b-row class="mb-3">
-          <b-col>
-            <!-- TODO: add new title + description box
-              Useful data:
-
-              titlesInCurrentLanguage
-              descriptionInCurrentLanguage
-            -->
+        <b-row class="mb-3 justify-content-center">
+          <b-col cols="10">
```
mirjamverloop, line 49: Could you change this (and everywhere else on this page) to 

`<b-col cols="12" class="col-lg-10">`

So it takes the full width on smaller screens.

components/generic/RightsStatementButton.vue
alansutherland: ok

pages/item-new/_.vue
alansutherland: ok

pages/item-new/_.vue
alansutherland: oh yes, good point

components/item/ItemHero.vue
alansutherland: yes that's true, have fixed styling

components/item/ItemHero.vue
alansutherland: I've added a ternary operator to fix this

mirjamverloop (approved):

PR #769, opened 2020-08-13
Use Set API search profile "itemDescriptions"

rwd:
Associated refactors:
* In plugins/europeana/search.js search(), do not mutate response from Record API, but return intact to caller
* Move search result mutations into item preview card (group) components
* Replace similar items component with variant of item preview card group


components/account/UserSets.vue
```diff
@@ -2,7 +2,15 @@
   <b-container>
     <b-row class="flex-md-row pb-5">
       <b-col cols="12">
+        <LoadingSpinner
+          v-if="$fetchState.pending"
+        />
+        <AlertMessage
+          v-if="$fetchState.error"
```
rwd, line 9: Should be v-else-if

PR #772, opened 2020-08-14
One playable media resource only

rwd:
Because the media player downloads all at once.


components/item/AwesomeSwiper.vue
```diff
@@ -44,7 +46,7 @@
 <script>
   import { Swiper, SwiperSlide } from 'vue-awesome-swiper';
   import 'swiper/css/swiper.css';
```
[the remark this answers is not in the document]
alansutherland: ok moved to css

components/item/AwesomeSwiper.vue
```diff
@@ -99,6 +104,9 @@
       },
       updateSwiper() {
         this.swiper.update();
+      },
+      isSinglePlayableMedia() {
+        return (this.playableMedia.length === 1) && (this.playableMedia[0].about === this.media[0].about);
```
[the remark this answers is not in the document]
alansutherland: true, I hadn't considered that. surely we don't actually need the second check, just `(this.playableMedia.length === 1);` should suffice right?

components/item/AwesomeSwiper.vue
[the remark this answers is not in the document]
alansutherland: this actually causes an error on the unit tests, it works fine while running the app but on testing it can't locate the css file, tried a few solutions for this but the best solution I found was putting it back as module import within script

components/item/MediaCard.vue
```diff
@@ -59,19 +86,22 @@
...
-        oEmbedData: {},
-        ratio: 56.25
+        oEmbedData: {}
       };
     },
 
     computed: {
       displayImage() {
-        return (this.imageSrc !== '') && !isRichMedia(this.media);
+        return (this.media.thumbnails !== '') && !isRichMedia(this.media);
```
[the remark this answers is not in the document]
alansutherland: we still need to check if the thumbnails attribute exists otherwise plain text would pass as a displayImage if we just checked 
```
displayImage() {
        return (!isRichMedia(this.media);
},
```

so I've switched the first check to confirm the thumbnails attribute exists

```
displayImage() {
        return (has(this.media, 'thumbnails')) && !isRichMedia(this.media);
      },
```

components/item/MediaCard.vue
```diff
@@ -17,6 +16,30 @@
       class="media-player"
     />
   </div>
+  <div
```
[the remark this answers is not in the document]
alansutherland: yes that works nicely, thanks for your help

components/item/AwesomeSwiper.vue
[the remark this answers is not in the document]
rwd: Fair enough.

components/item/AwesomeSwiper.vue
```diff
@@ -15,14 +15,16 @@
         class="container h-100"
       >
         <MediaCard
-          :europeana-identifier="item.europeanaIdentifier"
+          :europeana-identifier="europeanaIdentifier"
           :media="item"
+          :is-single-playable-media="isSinglePlayableMedia(item)"
```
rwd, line 20: `isSinglePlayableMedia` no longer accepts an argument. Remove it from this call.

components/item/AwesomeSwiper.vue
```diff
@@ -15,14 +15,16 @@
...
         />
       </div>
       <MediaCard
         v-else
-        :europeana-identifier="item.europeanaIdentifier"
+        :europeana-identifier="europeanaIdentifier"
         :media="item"
+        :is-single-playable-media="isSinglePlayableMedia(item)"
```
rwd, line 27: `isSinglePlayableMedia` no longer accepts an argument. Remove it from this call.

components/item/AwesomeSwiper.vue
```diff
@@ -91,6 +93,9 @@
     computed: {
       swiper() {
         return this.$refs.awesome.$swiper;
+      },
+      playableMedia() {
```
rwd, line 97: This is only ever used by `isSinglePlayableMedia`, which no longer accepts an argument, so just replace this computed property with one for `isSinglePlayableMedia`.

components/item/AwesomeSwiper.vue
```diff
@@ -99,6 +104,9 @@
       },
       updateSwiper() {
         this.swiper.update();
+      },
+      isSinglePlayableMedia() {
```
rwd, line 108: Move to computed properties, replacing `playableMedia`.

components/item/MediaCard.vue
```diff
@@ -59,19 +86,22 @@
...
-        oEmbedData: {},
-        ratio: 56.25
+        oEmbedData: {}
       };
     },
 
     computed: {
       displayImage() {
-        return (this.imageSrc !== '') && !isRichMedia(this.media);
+        return (has(this.media, 'thumbnails')) && !isRichMedia(this.media);
```
rwd, line 104: I still don't understand what this is aiming to achieve. `this.media.thumbnails` will always be an object here. If this is to serve unit tests, then set it as an object in the unit tests instead.

components/item/AwesomeSwiper.vue
alansutherland: ok

components/item/MediaCard.vue
alansutherland: I added this based on the test ( describe('displayImage') ) its items media props include `ebucoreHasMimeType = 'text/plain'` but I suppose there are no actual items with this type? so isRichMedia will suffice for determining if an item is an image

PR #789, opened 2020-08-25
Extend set store

rwd:


lbiedinger (approved):

store/set.js
```diff
@@ -1,51 +1,131 @@
...
-    state.liked = likesArray;
+  setLikedItems(state, value) {
+    state.likedItems = value;
+    state.likedItemIds = value.map(item => item.id);
+  },
+  like(state, itemId) {
+    state.likedItemIds.push(itemId);
+  },
+  unlike(state, itemId) {
+    state.likedItemIds.splice(state.likedItemIds.indexOf(itemId), 1);
```
annachristaki, line 21: Unlike should also remove the item from likedItems. Currently the grid still shows the item if you unlike it while you are on the likes page.

annachristaki (changes requested):
All looks good apart from minor fix

PR #790, opened 2020-08-25
Display API errors on set pages

annachristaki:


pages/set/_.vue
```diff
@@ -1,5 +1,24 @@
 <template>
+  <b-container v-if="$fetchState.pending">
+    <b-row class="flex-md-row py-4 text-center">
+      <b-col cols="12">
+        <LoadingSpinner
+          v-if="$fetchState.pending"
```
rwd, line 6: This condition is redundant if the container has the same.

pages/set/_.vue
```diff
@@ -128,17 +151,31 @@
...
-      this.description = set.description;
-      this.visibility = set.visibility;
-      this.creator = set.creator;
-      this.total = set.total || 0;
-      this.items = set.items;
+      const token = this.$auth.loggedIn ? this.$auth.getToken('keycloak') : null;
+      const headers = {};
+      if (token) headers['Authorization'] = token;
+      const apiUrl = `${this.apiConfig.set.origin}${this.apiConfig.set.path}/${this.$route.params.pathMatch}?wskey=${this.apiConfig.set.key}&profile=itemDescriptions`;
+      const set = await fetch(apiUrl, { headers })
```
rwd, line 158: Why is all of this here, duplicating the functionality from the set plugin?

If it's just to get the status code, then apiError includes that as a property named `statusCode`, so just catch the error.

rwd (changes requested):

rwd (approved):

lbiedinger (approved):

PR #793, opened 2020-08-26
Clear persistent user credentials on logout

spyrosbek:


store/set.js
```diff
@@ -43,9 +43,5 @@ export const actions = {
   async createLikes({ commit }) {
     const likesId = await this.$sets.createLikes().then(response =>  response.id.split('/').pop());
     commit('setLikesId', likesId);
-  },
-  // TODO: is this used?
-  reset({ commit }) {
```
rwd, line 48: Restore this as it is needed. You could remove the TODO note though.

plugins/authAxios.js
```diff
@@ -2,24 +2,17 @@ import recommendation from './europeana/recommendation';
 import set from './europeana/set';
 
 export default ({ $axios, $auth, store, redirect }, inject) => {
-  const token = $auth.getToken('keycloak');
+  const token = $auth.loggedIn ? $auth.getToken('keycloak') : null;
 
   const headers = {};
   if (token) headers['Authorization'] = token;
+  else delete $axios.defaults.headers.common['Authorization'];
```
rwd, line 9: Why is this needed?

pages/account/logout.vue
```diff
@@ -8,9 +8,9 @@
     layout: 'minimal',
     middleware: [
       'auth',
-      ({ $auth, store }) => {
-        store.dispatch('set/reset');
-        $auth.logout();
+      ({ app }) => {
```
rwd, line 11: Unless there is a particular reason not to, use the $goto helper:
```diff
diff --git a/pages/account/logout.vue b/pages/account/logout.vue
index f107d5aa..5fc6c13a 100644
--- a/pages/account/logout.vue
+++ b/pages/account/logout.vue
@@ -8,9 +8,14 @@
     layout: 'minimal',
     middleware: [
       'auth',
-      ({ app }) => {
-        const path = app.$auth.strategies.keycloak.options.end_session_endpoint;
-        window.location.replace(`${path}?redirect_uri=${encodeURIComponent(window.location.origin)}`);
+      ({ $auth, store, app }) => {
+        store.dispatch('set/reset');
+        app.$goto({
+          path: $auth.strategies.keycloak.options.end_session_endpoint,
+          query: {
+            'redirect_uri': window.location.origin
+          }
+        });
       }
     ]
   };
```

rwd (changes requested):

pages/account/logout.vue
spyrosbek: @rwd The $goto helper uses "window.location.href" for redirection. Like I mentioned in this commit
https://github.com/europeana/portal.js/pull/793/commits/e1b6ea38de382990b7f8d4dbf6ed6b73bb7b2415
there is a compatibility issue with the href property of window.location, that's why I replaced it with the method window.location.replace() which works also with Safari.

Please let me know how I should proceed. Maybe I could adjust the $goto method to use window.location.replace instead?

plugins/authAxios.js
annachristaki: The token is not cleared from the browser when auth.loggedIn is false, it's still there and valid for whatever time it was set to in Keycloak (I think the value Access Token Lifespan in keycloak controls it and defaults to 5 mins) . This leads to errors, if for example you try to like an item after logout it will throw an error that the user already has a likes collection because the token is sent and it is still valid.

store/set.js
annachristaki: This is not needed anymore since the window is reloaded and store starts empty, it was added before to deal with auth.logout() that was not reloading the window.

lbiedinger (approved):

PR #799, opened 2020-08-28
Add & like buttons on item page

alansutherland:


components/item/ItemHero.vue
```diff
@@ -21,7 +21,13 @@
           >
             {{ rightsStatement }}
           </span>
-          <div>
+          <div class="d-flex">
+            <UserButtons
+              v-if="showUserButtons"
+              v-model="identifier"
+              @like="$emit('like', identifier)"
```
rwd, line 28: These two event emissions are redundant, I think. Remove them if you concur.

rwd (changes requested):

components/item/ItemHero.vue
alansutherland: I agree, removed

rwd (approved):

components/item/ItemHero.vue
```diff
@@ -95,6 +100,9 @@
       },
       downloadEnabled() {
         return this.rightsStatement && !this.rightsStatement.includes('/InC/');
+      },
+      showUserButtons() {
+        return Boolean(Number(process.env.ENABLE_XX_USER_AUTH)) && Boolean(Number(process.env.ENABLE_UNAUTHENTICATED_USER_BUTTONS));
```
lbiedinger, line 105: This won't show the button for logged in users now, add:
`|| (this.$store.state.auth && this.$store.state.auth.loggedIn));`
To bring it in line with this https://github.com/europeana/portal.js/blob/master/components/item/ItemPreviewCard.vue#L78

components/account/UserButtons.vue
```diff
@@ -82,7 +82,11 @@
         this.$bvModal.show(this.addItemToSetModalId);
       },
       async toggleLiked() {
-        await (this.liked ? this.unlike() : this.like());
+        if (this.$auth.loggedIn) {
+          await (this.liked ? this.unlike() : this.like());
+        } else {
+          this.$auth.loginWith('keycloak');
+        }
```
lbiedinger, line 89: I don't think this change is needed.
Attempting to like an item while not logged in will trigger the authentication in the store.
Revert this and the changes to the specs, unless there's a case I'm missing?

lbiedinger (changes requested):

components/item/ItemHero.vue
alansutherland: oops, ok added

components/account/UserButtons.vue
alansutherland: it was showing me an error when I was logged out, but now I'm unable to replicate. Reverting

lbiedinger (approved):

PR #809, opened 2020-09-03
show recommendations on sets

annachristaki:


lbiedinger (changes requested):
The "Items you might like" heading should have a font weight of 600.
To achieve this add `class="related-heading"` on the h2

pages/set/_.vue
```diff
@@ -154,7 +154,12 @@
         }
         throw apiError;
       }
+      this.getRecommendations();
+      this.$watch('set.total', function() {
```
lbiedinger, line 158: This should go in the watch block below with 'set.id'

pages/set/_.vue
```diff
@@ -154,7 +154,12 @@
...
+        this.getRecommendations();
+      });
     },
+    fetchOnServer: false,
```
lbiedinger, line 162: This doesn't make sense as it also prevents the error code from being set which only happens ` if (process.server)`.

The reason recommendations aren't being shown on SSR seems to me to be that this.$auth.loggedIn is always false while on SSR.

lbiedinger (changes requested):

lbiedinger (approved):

PR #811, opened 2020-09-03
Fix style of edit-collection modal

spyrosbek:


components/set/SetFormModal.vue
```diff
@@ -50,20 +50,22 @@
...
-          >
-            {{ $t('set.actions.delete') }}
-          </b-button>
-          <b-button
-            variant="primary"
-            type="submit"
-          >
-            {{ isNew ? $t('set.actions.create') : $t('set.actions.update') }}
-          </b-button>
+          <div class="right">
```
mirjamverloop, line 53: Adding the div is enough. The class "right" currently doesn't do anything, could you remove it? 
The space between the red and blue buttons is slightly larger than the design, this can be fixed by adding a d-flex class on this div. 

In other words, can you replace `<div class="right">` with `<div class="d-flex">`?

mirjamverloop (changes requested):

mirjamverloop (approved):

lbiedinger (approved):

PR #818, opened 2020-09-08
Update blog style

alansutherland:


assets/scss/hero.scss
```diff
@@ -90,22 +90,74 @@
...
+      background-color: $blue;
+      border: none;
+    }
+    cite {
+      margin-bottom: 1.5rem;
+    }
+  }
+  .credits {
+    max-width: 720px;
+  }
```
lbiedinger, line 160: where is this used?

assets/scss/hero.scss
```diff
@@ -90,22 +90,74 @@
...
+  article {
+    .author:last-of-type .comma {
+      display: none;
+    }
+  }
+  .article {
+    iframe {
+      border: none;
+      width: 100%;
+      min-height: 400px;
```
lbiedinger, line 130: this makes iframes like the audio player much too large.

If people are going to use iframes like this it's up to the editor to make sure they're styled correctly.

Just remove the height attribute.

lbiedinger (changes requested):
The multiple author prop isn't needed, this can be achieved via css. I think this also gets rid of the linting error:
(maybe move the css to a scss file though to prevent the ',' from appearing after the component loads)
```
diff --git a/components/blog/BlogAuthor.vue b/components/blog/BlogAuthor.vue
index 1efa6cf0..8abeb44e 100644
--- a/components/blog/BlogAuthor.vue
+++ b/components/blog/BlogAuthor.vue
@@ -1,22 +1,17 @@
 <template>
   <div class="author">
-    <SmartLink
+    <span
       v-if="url"
-      :destination="url"
-      class="font-weight-bold"
     >
-      {{ name }}<!-- remove space
+      <SmartLink
+        :destination="url"
+        class="font-weight-bold"
+      >
+        {{ name }}<!-- remove space
     --></SmartLink>
-    <span v-else>{{ name }}</span><!-- remove space
-    --><span
-      v-if="multipleAuthors && !organisation"
-      class="comma"
-    >,</span>
-    <span v-if="organisation">({{ organisation }})</span><!--
- --><span
-      v-if="multipleAuthors && organisation"
-      class="comma"
-    >,</span>
+    </span><!-- remove space
+    --><span v-else>{{ name }}</span><!-- remove space
+    --><span v-if="organisation">, ({{ organisation }})</span>
   </div>
 </template>
 
@@ -53,3 +48,18 @@
     }
   };
 </script>
+
+<style lang="scss" scoped>
+  @import './assets/scss/variables.scss';
+
+  .author {
+    &::after {
+      content: ', ';
+    }
+    &:last-child {
+      &::after {
+        content: '';
+      }
+    }
+  }
+</style>
```

I also wonder though if multiple authors shouldn't be seperated via a ';' instead? If so, then it should use the formatting.listSeperator value from the translations file(s). Was there something for this in the design?

assets/scss/hero.scss
alansutherland: .credits is used in the exhibitions.

good point about the authors, I keep doing everything purely the JS way and not going for the easier/better CSS solution, have changed. Still need to keep the comment spacing between {{ name }} and the closing </SmartLink> though because this was causing the underline text-decoration to continue after the author name in the space before organisation

assets/scss/hero.scss
alansutherland: sure

lbiedinger (approved):

PR #823, opened 2020-09-10
Show message when user has no likes collection

annachristaki:


pages/account/index.vue
```diff
@@ -26,7 +26,7 @@
                 <b-row class="flex-md-row">
                   <b-col cols="12">
                     <div
-                      v-if="likedItems && likedItems.length === 0"
+                      v-if="$fetchState.pending"
```
lbiedinger, line 29: This is going to re-introduce the wait on the private & public set calls being fully loaded before showing the likes tab.

Instead maybe check that there is a likesId available. _or_ whenever a user visits their account page ensure a likes set is created if there isn't one already.

lbiedinger (changes requested):

pages/account/index.vue
```diff
@@ -39,7 +39,7 @@
                       v-else
                     >
                       <div
-                        v-if="!likedItems"
+                        v-if="!likedItems || likedItems && likedItems.length === 0"
```
lbiedinger, line 42: `likedItems &&` is redundant after the `!likedItems` condition.

pages/account/index.vue
```diff
@@ -116,8 +116,7 @@
     },
 
     async fetch() {
-      this.fetchLikes();
-      await this.$store.dispatch('set/fetchCreations');
```
lbiedinger, line 120: by removing this here I believe there's the possibility that the 'public' & 'private' tabs will show "You haven't created any x collections yet" while they have not yet been retrieved, as these are conditional based on this fetch state.

I can't actually see this because you still have to select them, but I imagine if we were to allow a user to load the account page by use of an anchor tag like /account#private they may see the message show briefly.

This is what @spyrosbek was attempting to solve previously. Spyros are you able to reproduce the issue on this branch?

pages/account/index.vue
```diff
@@ -134,17 +133,17 @@
...
       likesId: 'fetchLikes'
     },
-
+    mounted() {
+      this.$store.dispatch('set/fetchCreations');
+    },
     methods: {
       fetchLikes() {
-        this.$store.dispatch('set/fetchLikes');
+        return this.$store.dispatch('set/fetchLikes');
```
lbiedinger, line 144: I don't think this needs to ever return anything. The return value isn't ever assigned, all the uses are watching the store value directly.

pages/account/index.vue
spyrosbek: @lbiedinger Yes, at this state of the branch, I'm seeing again the "You haven't created any x collections yet" message briefly, until creations are finished fetching.

PR #827, opened 2020-09-11
Refactor like items empty modelling

lbiedinger:
Use empty array for likedItems only after API call or when no Likes set exists.
Otherwise default to undefined.


pages/account/index.vue
```diff
@@ -25,8 +25,23 @@
               <b-container>
                 <b-row class="flex-md-row">
                   <b-col cols="12">
+                    <template
+                      v-if="likedItems"
```
spyrosbek, line 29: Since there will always be a likedItems array (after the changes in set store), this condition will always be satisfied. As a result, the likes-spinner will never appear.

I tested it and, indeed, the empty-likes message appears while the likes are still loading, instead of the spinner

rwd (approved):

PR #828, opened 2020-09-14
Use Elastic APM Vue plugin

rwd:
Locked at version 0.4.1 because @elastic/apm-rum v5 introduces breaking changes for distributed tracing requiring support on APIs etc for `traceparent`.


rwd:
Superceded by #944

PR #834, opened 2020-09-17
Refactor search form component

lbiedinger:


components/entity/EntityDetails.vue
```diff
@@ -56,6 +56,10 @@
         type: Object,
         required: true
       },
+      query: {
+        type: String,
+        default: null
+      },
```
mirjamverloop, line 62: Is this used in this component?

components/search/SearchQueryOptions.vue
```diff
@@ -0,0 +1,426 @@
...
+    class="auto-suggest-dropdown"
+    data-qa="search suggestions"
+    role="listbox"
+    :aria-label="$t('searchSuggestions')"
+    :aria-hidden="!isActive"
+  >
+    <template
+      v-if="onCollectionPage"
+    >
+      <b-list-group-item
```
mirjamverloop, line 13: These list-group-items are not the same, but look similar. I am wondering if the code can be more DRY?

components/search/SearchQueryOptions.vue
```diff
@@ -0,0 +1,426 @@
...
+        :to="inCollectionLinkGen(query)"
+        class="search"
+        data-qa="search in collection button"
+        role="option"
+        :aria-label="$t('search')"
+        :aria-selected="focus === 0"
+        :class="{ 'hover': focus === 0 }"
+        :data-index="0"
```
mirjamverloop, line 21: I am not sure if the data-index still has a purpose, especially because not all list-group-items have them.

pages/search/index.vue
```diff
@@ -63,15 +68,29 @@
...
+<style lang="scss" scoped>
+  h1 {
+    font-size: 1.875rem;
+    font-weight: 300;
+    line-height: 1.375;
+    margin-bottom: 1rem;
+    span {
+      font-weight: 600;
+    }
+  }
```
mirjamverloop, line 95: This might have been discussed somewhere, but I think we should keep the same h1 styling on every page. That is better for accessibility too.

components/entity/EntityDetails.vue
lbiedinger: no, left over from previous iteration when the query was shown alongside the entity title instead of above the results as  is the case now.

components/search/SearchQueryOptions.vue
lbiedinger: each item is already using a component, and there are subtle differences between the various elements. So the I'm not sure how this could be refactored.

pages/search/index.vue
lbiedinger: yeah, this is to align the styling with that of the heading on collection pages where there is already a h1 for the collection title. We could  make it a h2 on the search results page, but then we'd be missing a h1 for the title here altogether.

rwd (approved):

components/PageHeader.vue
```diff
@@ -34,62 +47,64 @@
...
-      <PageNavigation
-        v-if="mainNavigation"
-        :links="mainNavigation.links"
-      />
-    </b-navbar>
-    <transition name="slide">
       <b-navbar
-        v-if="showSidebar"
-        class="p-lg-0 align-items-start justify-content-lg-end flex-column flex-lg-row d-lg-none"
+        class="align-items-center flex-row d-flex p-0 pr-3"
```
alansutherland: I spotted a minor regression caused by pr-3, this caused the hamburger icon to be half width on Safari/iOS devices. I've pushed a fix for this.
Screenshot 2020-10-09 at 19 34 30

alansutherland (approved):

PR #836, opened 2020-09-18
Exhibition load order; Remove include & consistent formatting in blog style

lbiedinger:


components/sharing/ShareButton.vue
```diff
@@ -12,8 +12,7 @@
 </template>
 
 <style lang="scss" scoped>
-  @import './assets/scss/variables.scss';
-  @import './assets/scss/icons.scss';
```
rwd, line 16: There is a span using `icon-share` in this template. Isn't this SCSS file therefore needed?

lbiedinger: all the icons.scss styling is loaded already in assets/scss/style.scss the reason it needs to be loaded in some other components/pages is that those modify/overwrite or inherit/use the default icon styles.

rwd (approved):

PR #845, opened 2020-09-24
fix multiple IIIF render bug

alansutherland:


rwd:
The issue this seeks to fix is caused by incorrect modelling at the data level and will not be fixed on the front-end.

PR #847, opened 2020-09-25
Run entity search from page

rwd:


rwd:
Would need to be rewritten given subsequent mainline changes.

PR #859, opened 2020-10-07
Disable form on submit to avoid duplicate collections creation

alansutherland:


components/set/SetFormModal.vue
```diff
@@ -135,6 +135,7 @@
         titleValue: '',
         descriptionValue: '',
         isPrivate: false,
+        disableForm: false,
```
rwd, line 138: Rename this to `submissionPending`

components/set/SetFormModal.vue
```diff
@@ -179,13 +180,18 @@
...
       // TODO: error handling
       submitForm() {
-        const handler = this.isNew ?
-          this.$store.dispatch('set/createSet', this.setBody) :
-          this.$store.dispatch('set/updateSet', { id: this.setId, body: this.setBody });
-
-        return handler.then(() => {
-          this.hide(this.isNew ? 'create' : 'update');
-        });
+        if (!this.disableForm) {
```
rwd, line 183: Change this to `if (this.submissionPending) return;`

rwd (changes requested):

rwd (approved):

PR #863, opened 2020-10-08
Remove labels from foyer page cards

alansutherland:


components/generic/ContentCard.vue
```diff
@@ -141,6 +141,10 @@
         type: Boolean,
         default: true
       },
+      noSubtitle: {
```
rwd, line 144: Let's name this showSubtitle instead, defaulting to true.

rwd (changes requested):

components/generic/ContentCard.vue
alansutherland: good idea

rwd (approved):

rwd (approved):

PR #869, opened 2020-10-14
Provider name links to source

alansutherland:


lbiedinger (changes requested):

plugins/europeana/record.js
```diff
@@ -73,7 +73,7 @@ function extraFields(proxyData, edm, entities) {
   const providerAggregation = edm.aggregations[0];
   const europeanaAggregation = edm.europeanaAggregation;
   return lookupEntities(omitBy({
-    edmProvider: providerAggregation.edmProvider,
+    edmProvider: { link: providerAggregation.edmIsShownAt, value: providerAggregation.edmProvider },
```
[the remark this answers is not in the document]
alansutherland: ok thanks for your help

lbiedinger (approved):

components/item/MetadataField.vue
```diff
@@ -33,13 +33,29 @@
             />
           </li>
         </template>
+        <template v-else-if="typeof fieldData.url !== 'undefined'">
```
mirjamverloop, line 36: Would it work to just say `<template v-else-if="fieldData.url">`?

components/item/MetadataField.vue
```diff
@@ -33,13 +33,29 @@
...
+          <li
+            :key="index"
+            :lang="langMappedValues.code"
+            data-qa="literal value"
```
mirjamverloop, line 40: Is this correct? As in this case, it is not the literal value, right?

components/item/MetadataField.vue
```diff
@@ -33,13 +33,29 @@
...
+          >
+            <SmartLink
+              :destination="fieldData.url"
+              :link-class="name === 'edmDataProvider' ? 'view-at' : null"
```
mirjamverloop, line 44: Should this link show an icon as it is/could be an external link?

mirjamverloop (changes requested):

components/item/MetadataField.vue
alansutherland: yes that does appear to suffice

components/item/MetadataField.vue
alansutherland: as discussed we'll change to 'entity link'

components/item/MetadataField.vue
alansutherland: ok I've added the icon

mirjamverloop (approved):

mirjamverloop (approved):

PR #874, opened 2020-10-19
Pre-filter unusued Record API data

rwd:
From the Record API response, before being passed to the Vue components:
* Reduce lang maps to the locale needed to be displayed to the user
* Remove web resource attributions
* Reduce entities to just ID, prefLabel and geo co-ordinates
* Freeze objects to prevent them being made reactive


plugins/europeana/record.js
```diff
@@ -336,6 +346,41 @@ export default (axiosOverrides) => {
...
+    }
+  } else {
+    return value;
+  }
+};
+
+const isLangMap = (value) => {
+  return (typeof value === 'object') && Object.keys(value).every(key => {
+    // TODO: is this good enough to determine lang map or not?
+    return key.length === 2 || key.length === 3;
```
lbiedinger, line 373: probably we'll need to account for 'en-GB' style codes too.

plugins/europeana/record.js
```diff
@@ -336,6 +346,41 @@ export default (axiosOverrides) => {
...
+  });
+};
+
+const reduceEntity = (entity) => {
+  return Object.freeze({
+    about: entity.about,
+    prefLabel: entity.prefLabel
```
lbiedinger, line 380: Probably needs to include altLabel, maybe even hiddenLabel?

plugins/europeana/record.js
```diff
@@ -257,6 +261,10 @@ export default (axiosOverrides) => {
 
         // Inject service definitions, e.g. for IIIF
         webResource.services = services.filter((service) => (webResource.svcsHasService || []).includes(service.about));
+
+        // Remove unused data
+        delete webResource.htmlAttributionSnippet;
+        delete webResource.textAttributionSnippet;
```
lbiedinger, line 267: I think it's going to be easier long term if we declare the fields we do want, then only select those. Instead of removing specific ones.

plugins/europeana/record.js
rwd: Not at present.

plugins/europeana/record.js
rwd: Indeed, e.g. http://localhost:3000/en/item/2021672/resource_document_mauritshuis_L60. Updated accordingly.

plugins/europeana/record.js
rwd: Indeed, and done.

lbiedinger (approved):

mirjamverloop (approved):

PR #880, opened 2020-10-22
Mobile slider position

alansutherland:


pages/iiif/index.vue
```diff
@@ -46,7 +46,7 @@
               attribution: false,
               canvas: false,
               annotations: true,
-              search: false
+              search: true
```
rwd, line 49: This should not be here.

components/item/AwesomeSwiper.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <swiper
+    v-show="swiperLoaded"
```
rwd, line 3: What does this achieve?

rwd (changes requested):

pages/iiif/index.vue
alansutherland: sorry, removed

components/item/AwesomeSwiper.vue
alansutherland: I added a v-show conditional since the first image was appearing unstyled before the swiper had finished loading.

rwd: I think then that it may be worth using client-only instead in the parent component as that keeps the swiper out of the server-side render altogether.

So you'd remove all the swiperLoaded-related changes in this component and instead do this in the parent:
```
diff --git a/components/item/ItemHero.vue b/components/item/ItemHero.vue
index 9f4bc21c..88df6425 100644
--- a/components/item/ItemHero.vue
+++ b/components/item/ItemHero.vue
@@ -1,10 +1,12 @@
 <template>
   <div class="item-hero">
-    <AwesomeSwiper
-      :europeana-identifier="identifier"
-      :displayable-media="displayableMedia"
-      @select="selectMedia"
-    />
+    <client-only>
+      <AwesomeSwiper
+        :europeana-identifier="identifier"
+        :displayable-media="displayableMedia"
+        @select="selectMedia"
+      />
+    </client-only>
     <b-container>
       <b-row>
         <b-col
@@ -45,17 +47,18 @@
 </template>
 
 <script>
-  import AwesomeSwiper from './AwesomeSwiper';
   import DownloadButton from '../generic/DownloadButton.vue';
   import RightsStatementButton from '../generic/RightsStatementButton.vue';
   import SocialShareModal from '../sharing/SocialShareModal.vue';
   import ShareButton from '../sharing/ShareButton.vue';
   import has from 'lodash/has';
+  import ClientOnly from 'vue-client-only';
   import { isIIIFPresentation } from '../../plugins/media';
 
   export default {
     components: {
-      AwesomeSwiper,
+      AwesomeSwiper: () => import('./AwesomeSwiper'),
+      ClientOnly,
       DownloadButton,
       RightsStatementButton,
       SocialShareModal,
```

What do you think?

alansutherland: This is a better solution

lbiedinger (approved):

mirjamverloop (approved):

rwd (approved):

PR #881, opened 2020-10-23
Scale blog foyer images to 510px

alansutherland:


pages/blog/index.vue
```diff
@@ -81,6 +83,21 @@
       };
     },
 
+    methods: {
+      imageUrl(post) {
+        if (post.primaryImageOfPage) {
+          return post.primaryImageOfPage.image.url;
+        }
+        return '';
```
rwd, line 91: Return null here instead of an empty string.

pages/blog/index.vue
```diff
@@ -81,6 +83,21 @@
...
+      },
+      imageContentType(post) {
+        if (post.primaryImageOfPage) {
+          return post.primaryImageOfPage.image.contentType;
+        }
+        return '';
```
rwd, line 97: Return null here instead of an empty string.

rwd (changes requested):

rwd (approved):

PR #884, opened 2020-10-27
Exhibition chapter preview

mirjamverloop:


lbiedinger (changes requested):

lbiedinger:
this functionality can be achieved via the contentful preview settings directly using `entry.linkedBy` as a selector to get the landing page identifer.

PR #886, opened 2020-10-30
Show 100 of total if > 100

alansutherland:


pages/set/_.vue
```diff
@@ -97,7 +97,11 @@
       <b-row>
         <b-col>
           <h2 class="related-heading text-uppercase">
-            {{ $tc('items.itemCount', itemCount, { count: itemCount }) }}
+            {{
+              itemCount > 100 ?
+              $tc('items.itemOf', itemCount, { count: itemCount }) :
+              $tc('items.itemCount', itemCount, { count: itemCount })
```
lbiedinger, line 103: these two lines need to be indented one more level as they're par t of the previous statement.
Although I think having to split ternary operators onto multiple lines like this is also not great. Let's move the whole thing into a computed property and have the logic there as follows:

```
 computed: {
   ...
   displayItemCount() {
      const label = this.set.total > 100 ? 'items.itemOf' : 'items.itemCount';
      return this.$tc(label, this.this.set.total) :
   }
```
Which can replace the 'itemCount' property as that was only used here anyway.

Also note the change of what's passed to the $tc function, I don't know why count was being passed twice, but I suspect the intention was once for pluralization and once for the value, however once should suffice. I tried it and it seems to work for 0, 1, 2...100 & 101 items, but worth double checking.

lbiedinger (changes requested):

pages/set/_.vue
alansutherland: Yes, this approach makes more sense than the ternary operator. I'm not sure why the count was being passed twice before but as you say it's not needed. I've tested multiple different numbers of items in a set and found no issues.

lbiedinger (approved):

lang/en.js
```diff
@@ -475,6 +475,7 @@ export default {
   },
   "items": {
     "itemCount": "1 item |{count} items",
+    "itemOf": "100 of {count} items",
```
rwd, line 478: 100 should be a variable inserted by interpolation, not hard-coded here.

rwd (changes requested):

lang/en.js
alansutherland: ok 100 has been moved to a variable

rwd (approved):

PR #890, opened 2020-11-03
Recommendation toggle

mirjamverloop:


pages/set/_.vue
```diff
@@ -184,6 +184,9 @@
       displayDescription() {
         return langMapValueForLocale(this.set.description, this.$i18n.locale);
       },
+      enableRecommendations() {
```
lbiedinger, line 187: Add a check for the status of this in the mounted hook and the set.total watcher so as to prevent retrieval of recommendations.

pages/item/_.vue
```diff
@@ -274,7 +274,7 @@
         const noSimilarItems = { results: [] };
         if (this.error) return noSimilarItems;
 
-        if (this.$auth.loggedIn) {
+        if (this.$auth.loggedIn && Boolean(Number(process.env.ENABLE_RECOMMENDATIONS))) {
```
lbiedinger, line 277: I think we may benefit ever so slightly from switching these around as I'd imagine the loggedIn check is a bit more expensive since it needs to check the token and might result in a redirect to log in again if the token has expired.

lbiedinger (changes requested):

pages/set/_.vue
mirjamverloop: I have moved it to the getRecommendations method itself, where it checks for both this toggle and the logged in status.

pages/item/_.vue
mirjamverloop: Yes, good suggestion!

lbiedinger (approved):

pages/set/_.vue
```diff
@@ -115,7 +115,7 @@
         </b-col>
       </b-row>
       <b-row
-        v-if="recommendations.length > 0"
+        v-if="enableRecommendations && recommendations.length > 0"
```
rwd, line 118: Is it necessary to check whether recommendations are enabled or not here? I think it should suffice to only check for that when about to retrieve them, and here to just look for whether there are any set.

mirjamverloop: Yes, you are right. I was thinking about removing it here this morning but kept it in for no particular reason. I have removed the check here.

rwd (approved):

PR #894, opened 2020-11-04
IIIF newspaper provider metadata

alansutherland:


components/item/MetadataField.vue
```diff
@@ -129,6 +129,8 @@
           return { values: this.fieldData, code: '' };
         } else if (typeof this.fieldData.url !== 'undefined') {
           return langMapValueForLocale(this.fieldData.value, this.$i18n.locale);
+        } else if (this.name === 'edmDataProvider') {
```
lbiedinger, line 132: instead of relying on the specific field name here, let's check for presence of the 'url' key in the datastrucure, so that this would work for any other fields where we may want to take the same approach and have them link somewhere.

`else if (Object.prototype.hasOwnProperty.call(this.fieldData, 'url')` instead of `else if (typeof this.fieldData.url !== 'undefined')` above.

lbiedinger (changes requested):

lbiedinger (approved):

PR #901, opened 2020-11-09
HTML output share

LeoniePeters:


alansutherland (approved):

lbiedinger (changes requested):

components/sharing/SocialShareModal.vue
```diff
@@ -36,6 +68,45 @@
       mediaUrl: {
         type: String,
         default: null
+      },
+      identifier: {
+        type: String,
+        default: null
+      },
+      onItemPage: {
```
lbiedinger, line 76: Couldn't this be derived from whether or not the identifier is present?
```
methods: {
  onItemPage() {
    return !this.identifer
  }
  ...
}
```

components/sharing/SocialShareModal.vue
```diff
@@ -5,16 +5,46 @@
     hide-header-close
     hide-footer
     data-qa="share modal"
+    v-on="onItemPage ?
+      { show: requestEmbed,
```
lbiedinger, line 9: This results in an attempt to make a request every time the share modal is opened. I say attempt because actually the browser appears to cache it after the first attempt.

One way around this would be to check if `this.oEmbedDataHtml` is present, and if so return early in `requestEmbed`

lbiedinger (approved):

components/sharing/SocialShareModal.vue
```diff
@@ -5,16 +5,46 @@
...
+      :onItemPage ? { hide: resetEmbedCopied }
+        :{}"
   >
     <div class="icon-wrapper">
       <SocialShare
         :media-url="mediaUrl"
         with-text
       />
     </div>
+    <b-form
```
rwd, line 19: This should go into its own component, e.g. /components/item/ItemEmbedCode.vue.

SocialShareModal should then have an unnamed [slot](https://vuejs.org/v2/guide/components-slots.html) at this position, which the ItemHero would use to insert ItemEmbedCode, thus:
```
<SocialShareModal
  :media-url="selectedMedia.about"
>
  <ItemEmbedCode
    :identifier="identifier"
  />
</SocialShareModal>
```

components/sharing/SocialShareModal.vue
```diff
@@ -5,16 +5,46 @@
...
+      v-if="oEmbedDataHtml"
+      data-qa="share modal embed"
+      class="mt-3"
+      @submit.stop.prevent="submitForm"
+    >
+      <label> {{ $t('record.copyEmbedLabel') }}</label>
```
rwd, line 25: The label element needs to be linked to the textarea element it labels, with the attribute `for="shareEmbed"`.

Also remove the extraneous space after the opening label tag.

components/sharing/SocialShareModal.vue
```diff
@@ -5,16 +5,46 @@
...
+      <b-form-textarea
+        id="shareEmbed"
+        ref="shareEmbed"
+        v-model="oEmbedDataHtml"
+        readonly
+        tabindex="0"
```
rwd, line 31: Why is this needed?

components/sharing/SocialShareModal.vue
```diff
@@ -36,6 +68,41 @@
...
+                          `${EUROPEANA_DATA_URL}/item${this.identifier}`)
+          .then((response) => {
+            if (response.data.html) {
+              this.oEmbedDataHtml = response.data.html;
+            }
+          });
+      },
+      copyEmbedCode(event) {
+        if (event.type === 'click' || event.keyCode === 13) {
+          let textarea = this.$refs.shareEmbed;
```
rwd, line 98: This doesn't warrant a variable assignment. Just do `this.$refs.shareEmbed.select();`

components/sharing/SocialShareModal.vue
```diff
@@ -36,6 +68,41 @@
...
+        oEmbedForEndpoint(process.env.OEMBED_ENDPOINT || 'https://oembedjs.europeana.eu/',
+                          `${EUROPEANA_DATA_URL}/item${this.identifier}`)
+          .then((response) => {
+            if (response.data.html) {
+              this.oEmbedDataHtml = response.data.html;
+            }
+          });
+      },
+      copyEmbedCode(event) {
+        if (event.type === 'click' || event.keyCode === 13) {
```
rwd, line 97: [keyCode is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), so if possible we should not introduce further uses of it.

Can this instead use [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)?

rwd (changes requested):

components/sharing/SocialShareModal.vue
LeoniePeters: For accessibility we would want the textarea to be accessible through keyboard, since it's also clickable. But on reviewing, I guess we don't need to set tabindex, since textarea is an interactive element and already focusable. I believe that at some point I was testing and couldn't reach the textarea via keyboard so added the tabindex. Will take another look at this.

components/sharing/SocialShareModal.vue
LeoniePeters: Then the embed code will be copied on any key. This is for keyboard users. So if the user tabs through this module, focuses on the textarea with the embed code and hits tab again, the code will be copied. I think that would be annoying/undesired behavior.

LeoniePeters: [Vue key modifiers](https://vuejs.org/v2/guide/events.html#Key-Modifiers) solves this and makes that conditional unnecessary :+1:

components/sharing/SocialShareModal.vue
LeoniePeters: Tested keyboard accessibility and it is indeed not needed to use tabindex.

rwd (approved):

PR #920, opened 2020-11-26
Announcements to state the page loaded

LeoniePeters:
Vue Announcer library:
- Using the Vue Announcer plugin 
- Adds an aria-live region to the top of the default template
- This makes screen readers announce a message: Title (page title) has loaded

Furthermore changes:
- After a route change focus is reset to an element just before the skip link
- Cookie Disclaimer is disabled instead of visually hidden after accepting cookies


layouts/default.vue
```diff
@@ -99,7 +106,12 @@
     },
 
     watch: {
-      '$i18n.locale': '$fetch'
+      '$i18n.locale': '$fetch',
+      $route() {
+        this.$nextTick(() => {
+          this.$refs.toppage.focus();
```
rwd, line 112: Shouldn't we be reusing the existing `#main` element here?

.env.example
```diff
@@ -85,5 +85,8 @@ ENABLE_ITEM_EMBED_CODE=1
 # Enable recommendations
 # ENABLE_RECOMMENDATIONS=0
 
+# Enable announcer on route changes
+ENABLE_ANNOUNCER=1
```
rwd, line 89: Rename this to `ENABLE_VUE_ANNOUNCER`.

Also, while this is a work-in-progress, in the example file it should disabled by default, i.e.:
```
# ENABLE_VUE_ANNOUNCER=0
```

rwd (changes requested):

layouts/default.vue
LeoniePeters: After a route change we should reset the focus to the container div so that the skip link is the first focused on element after a tab. That's how I wrote it now. There are also arguments to put it directly on the skip link, but in that case the skip link would be visible after each route change and that's undesirable for our visual users.

lbiedinger (changes requested):

layouts/default.vue
LeoniePeters: In the end Mirjam and I concluded there is not an ideal solution. We decided that adding an empty element right before the skip link would work best. Setting focus on a wrapper causes unwanted behavior for screen readers (read to much).

components/generic/CookieDisclaimer.vue
```diff
@@ -61,6 +63,10 @@
 
       if (status !== 'accepted') {
         this.isOpen = true;
+        this.accepted = false;
+      }
+      if (status === 'accepted') {
```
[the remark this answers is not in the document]
LeoniePeters: Turned it into the following to avoid negated condition.
```javascript
if (status === 'accepted') { 
    this.accepted = true; 
} else { 
    this.isOpen = true; 
    this.accepted = false; 
}
```

layouts/default.vue
```diff
@@ -1,8 +1,16 @@
 <template>
   <div>
     <client-only>
+      <VueAnnouncer
+        v-if="enableAnnouncer"
```
lbiedinger, line 5: since the announcer is always on this conditional and the data property can be removed.

layouts/default.vue
```diff
@@ -99,7 +108,19 @@
...
+      '$i18n.locale': '$fetch',
+      $route(to, from) {
+        this.$nextTick(() => {
+          if (to.path === from.path) {
+            // this.enableAnnouncer = false;
+          } else {
+            this.$refs.resetfocus.setAttribute('tabindex', '0');
+            this.$refs.resetfocus.focus();
+            this.enableAnnouncer = true;
+          }
```
lbiedinger, line 120: if enableAnnouncer isn't needed this can simply be:
```
          if (to.path !== from.path) {
            this.$refs.resetfocus.setAttribute('tabindex', '0');
            this.$refs.resetfocus.focus();
          }
```

tests/unit/layouts/default.spec.js
```diff
@@ -0,0 +1,41 @@
...
+
+const store = new Vuex.Store({
+  state: { breadcrumb: {} }
+});
+
+const factory = () => shallowMount(layout, { localVue,
+  store,
+  data() {
+    return {
+      enableAnnouncer: true
```
lbiedinger, line 20: this also won't be needed

lbiedinger (changes requested):

layouts/default.vue
LeoniePeters: This was a mistake. We do need to disable the announcer when the path stays the same (to prevent an announcement after each application of a search filter for example.
I commented it out to test something and then forgot to uncomment it. In the last commit I brought it back in.

lbiedinger (approved):

layouts/default.vue
```diff
@@ -99,7 +108,19 @@
     },
 
     watch: {
-      '$i18n.locale': '$fetch'
+      '$i18n.locale': '$fetch',
```
rwd, line 111: At present the announcement is always in English.

Let's make use of the fact that we are already here watching for locale changes, and in the `fetch` function, set the announcement text based on the user's locale, using [`$announcer.set`](https://vue-announcer-v2.surge.sh/guide/announcer.html#methods) on the VM, with messages from `$i18n`.

rwd (changes requested):

layouts/default.vue
LeoniePeters: The Vue Announcer also offers the [$announcer.setComplementRoute](https://vue-announcer-v2.surge.sh/guide/announcer-router.html#announcer-setcomplementroute-complementroute) method specifically for this case so I'm using that. 
complementRoute refers to the string that is appended to the page title in the announcements after route changes.

LeoniePeters: And I also call it on the mount method to have the translated value always. Otherwise it would only check for translations after the user changes de language.

layouts/default.vue
```diff
@@ -99,7 +109,23 @@
...
+            this.$refs.resetfocus.focus();
+            this.enableAnnouncer = true;
+          }
+        });
+      }
+
+    },
+
+    mounted() {
+      this.$announcer.setComplementRoute(this.$t('pageHasLoaded'));
```
rwd, line 128: This should not be necessary, because `fetch` is a Nuxt lifecycle function called when the page is first rendered, as well as by the watching of `$i18n.locale`

rwd (approved):

PR #926, opened 2020-12-02
Homepage banner updated

alansutherland:
⚠️ Don't merge until after release v1.23.0


assets/scss/hero.scss
```diff
@@ -1,3 +1,27 @@
+.item-hero .user-buttons .btn {
```
alansutherland, line 1: I had to move this selector since stylelint was complaining about descending selectors
i.e. .item-hero .user-buttons .btn should come before .jumbotron.browse-jumbotron .container .card .btn

assets/scss/hero.scss
```diff
@@ -29,61 +53,85 @@
   }
 
   &.browse-jumbotron {
-    .container {
-      padding-bottom: 3rem;
+    margin-top: 2rem;
+    border-radius: 0.5rem;
+    overflow: hidden;
```
mirjamverloop, line 58: Is the overflow hidden neccassary? What is overflowing?

assets/scss/hero.scss
```diff
@@ -29,61 +53,85 @@
...
+    &:before {
+      display: none;
```
mirjamverloop, line 60: Is there a before?

assets/scss/hero.scss
```diff
@@ -29,61 +53,85 @@
...
-      display: block;
-      height: 0;
-      position: absolute;
-      right: 0;
-      width: 0;
-      z-index: 3;
+    .container {
+      .card {
+        margin: 1rem;
+        max-width: 37rem;
```
mirjamverloop, line 65: The card looks too large, it takes over 50% of the banner. Is that correct?

assets/scss/hero.scss
```diff
@@ -29,61 +53,85 @@
...
+        border: none;
+        color: $black;
+        .btn {
+          text-decoration: none;
+        }
```
mirjamverloop, line 70: The button has too much whitespace (top/bottom) compared to the design.

assets/scss/hero.scss
```diff
@@ -255,3 +232,33 @@
...
+      border-radius: 0;
+      .card {
+        margin: 0;
+        border-radius: 0;
+        h1 {
+          font-size: 1rem;
+        }
+        p, .btn {
+          font-size: 0.875rem;
+        }
```
mirjamverloop, line 260: There is a comment in Zeplin that says "On mobile the card should expand vertically from the top.
•	The image is always the same size / ratio." 

When I now take a look at the current implementation on mobile, an image would hardly be visible if there is a lot of text.

assets/scss/hero.scss
```diff
@@ -29,61 +53,85 @@
   }
 
   &.browse-jumbotron {
-    .container {
-      padding-bottom: 3rem;
+    margin-top: 2rem;
```
mirjamverloop, line 56: On smaller screens (anything between the mobile view of the header and desktop, like iPad portrait) the header is "clued" to the notification.

components/browse/HeroHeader.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <div class="container hero-wrapper">
+    <b-jumbotron
+      :style="jumbotronStyle"
+      fluid
+      class="browse-jumbotron"
+      text-variant="white"
```
mirjamverloop, line 7: This seems kind of contradictory, as the text in the banner is black/grey. This only affects the i icon for attribution. So perhaps remove this here, and add the white color there. Seems more clear.

components/browse/HeroHeader.vue
```diff
@@ -0,0 +1,98 @@
...
+      data-qa="hero banner"
+      @click="toggleCite"
```
mirjamverloop, line 9: Is this here with a reason? 
As it opens the attribution no matter where you click on the hero banner.

pages/index.vue
```diff
@@ -95,7 +112,7 @@
 
     head() {
       return {
-        title: this.$pageHeadTitle(this.name),
+        title: this.$t('homepageTitle') + this.$pageHeadTitle(),
```
mirjamverloop, line 115: The $t('homepageTitle') seems missing in the language file.

mirjamverloop (changes requested):

pages/index.vue
lbiedinger: it shouldn't be coming from the translation file. It should be the one from 'this.name'

assets/scss/hero.scss
alansutherland: I've removed this

assets/scss/hero.scss
alansutherland: yes, the before is still used on jumbotron images on the exhibition, it adds a semi-transparent dark cover over images

assets/scss/hero.scss
alansutherland: ok I was using rem based on the design px but now I've limited to 50%

assets/scss/hero.scss
alansutherland: fixed

assets/scss/hero.scss
alansutherland: well the mobile image needed to be set at a fixed height since the image is the same one used on desktop, if it was larger/used % based width then we would see tiling effects as the image grew. If the h1/p text can be limited then we can avoid having the image hidden

assets/scss/hero.scss
alansutherland: ok fixed

components/browse/HeroHeader.vue
alansutherland: good point

components/browse/HeroHeader.vue
alansutherland: this was added back in summer as one of the acceptance criteria for the new citations, clicking outside of the citation should close it. It does of course have the reverse effect too of opening the citation if it's closed

pages/index.vue
alansutherland: ok fixed

components/browse/HeroHeader.vue
mirjamverloop: Ah yes, now I see that it is working similarly on exhibitions/blogs. 
What I find slightly annoying is that if you click the CTA button, you briefly see the attribution before going to the link.

alansutherland: yeah that is annoying, we could change it to something like `toggleCite ? !toggleCite : null`

mirjamverloop (approved):

PR #935, opened 2020-12-08
Adding nuxt-google-optimize and first experiment

mirjamverloop:


layouts/default.vue
```diff
@@ -145,6 +145,9 @@
           { hreflang: 'x-default', rel: 'alternate', href: this.canonicalUrlWithoutLocale },
           ...i18nSeo.link
         ],
+        script: [
+          { src: `https://www.googleoptimize.com/optimize.js?id=${this.$config.googleOptimize.id}` }
+        ],
```
lbiedinger, line 150: This section should be optional depending on:
a) A google Optimize ID being present.
b) An experiment being active/available through `this.$exp`.
Otherwise we don't need to include the optimize code at all.

lbiedinger (changes requested):

PR #939, opened 2020-12-11
Timespan collection pages

alansutherland:


middleware/legacy/rules/explore.js
```diff
@@ -1,7 +1,7 @@
 // Redirect legacy gallery page URLs
 
 export default (route) => {
-  const pattern = /^\/portal(\/[a-z]{2})?\/explore\/(newcontent|colours|sources|topics|people|periods)(\.html)?$/;
+  const pattern = /^\/portal(\/[a-z]{2})?\/explore\/(newcontent|colours|sources|topics|times|people|periods)(\.html)?$/;
```
lbiedinger, line 4: this won't be needed as there were never 'times' pages on the classic portal

middleware/legacy/rules/entity.js
```diff
@@ -1,13 +1,13 @@
 // Redirect legacy entity page URLs
 
 export default (route) => {
-  const pattern = /^\/portal(\/[a-z]{2})?(\/explore\/(people|topics))(\/[0-9]+)/;
+  const pattern = /^\/portal(\/[a-z]{2})?(\/explore\/(people|topics|times))(\/[0-9]+)/;
```
rwd, line 4: This should not add "times" but "periods" which is what the classic portal used.

middleware/legacy/rules/entity.js
```diff
@@ -1,13 +1,13 @@
...
   const match = route.path.match(pattern);
 
   return match ? {
     path: [
       match[1],
-      match[2].replace('/explore', '/collections').replace('/people', '/person').replace('/topics', '/topic'),
+      match[2].replace('/explore', '/collections').replace('/people', '/person').replace('/topics', '/topic').replace('/times', '/timespan'),
```
rwd, line 10: This should replace "periods" with "time" (not "times" with "timespan")

plugins/europeana/record.js
```diff
@@ -202,9 +202,9 @@ export default (axiosOverrides) => {
       const concepts = (edm.concepts || []).map(reduceEntity).map(Object.freeze);
       const places = (edm.places || []).map(reduceEntity).map(Object.freeze);
       const agents = (edm.agents || []).map(reduceEntity).map(Object.freeze);
-      const timespans = (edm.timespans || []).map(reduceEntity).map(Object.freeze);
+      const times = (edm.timespans || []).map(reduceEntity).map(Object.freeze);
```
rwd, line 205: Revert this change, as here we use agents and concepts (as opposed to people and topics) so should also use timespans.

plugins/europeana/record.js
```diff
@@ -202,9 +202,9 @@ export default (axiosOverrides) => {
...
 
-      const entities = [].concat(concepts, places, agents, timespans)
+      const entities = [].concat(concepts, places, agents, times)
```
rwd, line 207: Revert this change, as here we use agents and concepts (as opposed to people and topics) so should also use timespans.

plugins/europeana/record.js
```diff
@@ -223,6 +223,7 @@ export default (axiosOverrides) => {
         media: this.aggregationMedia(providerAggregation, edm.type, edm.services),
         agents,
         concepts,
+        times,
```
rwd, line 226: Here we return agents and concepts (as opposed to people and topics) so should also return timespans (as opposed to times).

rwd (changes requested):

rwd (approved):

PR #940, opened 2020-12-11
Bug: content cards without images

LeoniePeters:


components/generic/ContentCard.vue
```diff
@@ -248,7 +247,7 @@
       },
 
       imageNotFound() {
-        this.cardImageUrl = '';
+        this.imageUrl = '';
```
rwd, line 250: Props can not be mutated by the component itself, hence the use of the data variable.

rwd (approved):

PR #953, opened 2020-12-22
New static page

LeoniePeters:


migrations/0037-create-static-page.js
```diff
@@ -0,0 +1,155 @@
...
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false)
+    .items({
+      type: 'Link',
+
+      validations: [
+        {
+          linkContentType: ['richText']
```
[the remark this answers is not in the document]
LeoniePeters: @mirjamverloop I followed what's available on the current browse page content type's section field: rich text, card group and latest card group. Leaving out the last two as I see them specific for browse pages.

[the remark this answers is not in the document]
mirjamverloop: I agree, those card groups are not applicable to the static page. I came up with my question after looking at the content model for blog posts. That model looks in some way similar to the static page and made me wonder if an editor should be able to add an image or link. Let's ask PO/design?

migrations/0038-update-browsePage-slug.js
```diff
@@ -0,0 +1,19 @@
+require('dotenv').config();
+
+module.exports = function(migration) {
+  const browsePage = migration.editContentType('browsePage');
+  browsePage
+    .editField('identifier')
+    .required(true);
+
+  if (!process.env.SLUG_VALIDATION_APP_ID) {
```
lbiedinger, line 9: this could even go up before anything else is even initialized. No point in going on if the APP isn't set.

migrations/0037-create-static-page.js
```diff
@@ -34,17 +36,31 @@ module.exports = function(migration) {
...
     ])
     .disabled(false)
     .omitted(false);
 
-  staticPage.changeFieldControl('identifier', 'builtin', 'slugEditor', {
-    helpText: 'Will always be prefixed with "/page/"'
+  if (!process.env.SLUG_VALIDATION_APP_ID) {
+    console.log('No app ID specified in SLUG_VALIDATION_APP_ID; aborting.');
+    process.exit(1);
+  }
```
lbiedinger, line 58: Also could go to the top of the file.

components/browse/BrowsePage.vue
```diff
@@ -0,0 +1,113 @@
...
+          this.$store.state.i18n.locale
+        }?utm_source=new-website&utm_medium=button`;
+      },
+      optimisedImageUrl() {
+        // use social media image if set in Contentful, otherwise use hero image
+        let img = this.image === null ? this.heroImage : this.image;
+        return this.$options.filters.optimisedImageUrl(img.url, img.contentType, {
+          width: 800,
+          height: 800
+        });
```
mirjamverloop, line 88: This isn't used in this component, but optimisedImageUrl is called in pages/index.vue. In pages/index.vue there is no computed property or method with that name. So I assume the social media images aren't coming through correctly.

components/static/StaticPage.vue
```diff
@@ -0,0 +1,106 @@
...
+      optimisedImageUrl() {
+        // use social media image if set in Contentful, otherwise use hero image
+        let img = this.image === null ? this.hero : this.image;
+        return this.$options.filters.optimisedImageUrl(
+          img.url,
+          img.contentType,
+          { width: 800, height: 800 }
+        );
+      }
+    }
```
mirjamverloop, line 104: Similar to what I noted for the BrowsePage, this is not being used in this component.

mirjamverloop (changes requested):

components/static/StaticPage.vue
```diff
@@ -0,0 +1,96 @@
...
+        default: null
+      },
+      primaryImageOfPage: {
+        type: Object,
+        default: null
+      },
+      image: {
+        type: Object,
+        default: null
+      },
```
mirjamverloop, line 81: These two props aren't used in this component.

components/browse/BrowsePage.vue
```diff
@@ -0,0 +1,101 @@
...
+        type: String,
+        default: null
+      },
+      primaryImageOfPage: {
+        type: Object,
+        default: null
+      },
+      image: {
+        type: Object,
+        default: null
```
mirjamverloop, line 64: These two props are no longer used in this component.

pages/index.vue
```diff
@@ -1,46 +1,39 @@
...
-      v-if="heroImage"
+  <div>
+    <BrowsePage
+      v-if="browsePage"
+      :identifier="identifier"
+      :name="name"
+      :headline="headline"
+      :description="description"
+      :primary-image-of-page="primaryImageOfPage"
+      :image="image"
```
mirjamverloop, line 10: As far as I can tell, these props are not used in the component itself.

pages/index.vue
```diff
@@ -1,46 +1,39 @@
...
-        :sections="hasPartCollection.items"
-      />
-    </b-container>
+    <StaticPage
+      v-if="staticPage"
+      :identifier="identifier"
+      :name="name"
+      :description="description"
+      :primary-image-of-page="primaryImageOfPage"
+      :image="image"
```
mirjamverloop, line 21: Similar to the BrowsePage, these props are not used in the component itself.

pages/index.vue
```diff
@@ -50,34 +43,51 @@
...
+
     computed: {
-      showNotificationBanner() {
-        return this.$config.app.features.linksToClassic && (this.identifier === 'home');
-      },
-      notificationUrl() {
-        return `https://classic.europeana.eu/portal/${this.$store.state.i18n.locale}?utm_source=new-website&utm_medium=button`;
-      },
       optimisedImageUrl() {
+        console.log(this.image, this.hero.image.url);
```
mirjamverloop, line 84: Remove the console.log

mirjamverloop (approved):

modules/contentful-graphql/queries/static-page.graphql
```diff
@@ -0,0 +1,73 @@
+query StaticPage(
```
lbiedinger, line 1: This whole query is moved to be part of BrowseStaticPage, and actually never used right? 
Then let's get rid of it.

pages/index.vue
```diff
@@ -50,34 +39,50 @@
         preview: query.mode === 'preview'
       };
 
-      return app.$contentful.query('browsePage', variables)
```
lbiedinger, line 53: Since this looks like it was the only place the 'browsePage' query was called, I think that query can be removed from the codebase.

pages/index.vue
```diff
@@ -50,34 +39,50 @@
...
-          if (data.browsePageCollection.items.length === 0) {
+          if (data.staticPageCollection.items.length > 0) {
+            let itemData = data.staticPageCollection.items[0];
+            itemData.staticPage = true;
+            return itemData;
+          } else if (data.browsePageCollection.items.length > 0) {
+            let itemData = data.browsePageCollection.items[0];
+            itemData.browsePage = true;
+            return itemData;
+          } else if (data.browsePageCollection.items.length === 0) {
```
lbiedinger, line 53: no need for the if  condition, just have it be `else {` or use a default return without any condition as was happening before, just with a 404 as the default.

mixins/exhibitionChapters.js
```diff
@@ -0,0 +1,49 @@
+export default {
+  methods: {
```
lbiedinger, line 2: Did you look into adding some unit tests for this mixin?
I think it would be useful to add tests as this code is going to be used in multiple locations.

lbiedinger (changes requested):

components/browse/BrowsePage.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+  <div data-qa="browse page">
+    <NotificationBanner
```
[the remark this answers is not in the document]
LeoniePeters: Well spotted. Fixed the html with a slot.

mixins/exhibitionChapters.js
```diff
@@ -0,0 +1,49 @@
...
+    manipulateChapters(chapters, identifier) {
```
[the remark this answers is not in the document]
LeoniePeters: As discussed renamed to 'chapterPagesToLinkListItems'

lbiedinger (approved):

PR #963, opened 2021-01-13
Custom dimensions GA bug

LeoniePeters:


tests/unit/pages/item/_.spec.js
```diff
@@ -78,4 +82,14 @@ describe('pages/item/_.vue', () => {
       headMeta.find(meta => meta.property === 'og:image').content.should.eq(thumbnailUrl);
     });
   });
+
+  describe('gtm()', () => {
```
lbiedinger, line 86: This test only asserts that the test itself called the method. Since the gtm code only executes in the browser environment it's also not straight forward to test that though.
Instead let's refactor the options for gtm into their own method and add a unit test for that, like so:
```
  describe('gtmOptions()', () => {
    it('uses first edmDataProvider from corefields', () => {
      const wrapper = factory();
      wrapper.vm.gtmOptions().should.eq(optionsVar);
    });
  });
```

Where optionsVar is set at the top here instead of in another fixture file to make it easier to see the input and expected values within this test alone.

lbiedinger (changes requested):

lbiedinger (approved):

PR #967, opened 2021-01-21
Update nav hover states

alansutherland:


components/PageNavigation.vue
```diff
@@ -194,6 +194,15 @@
     .nav-link {
       color: $mediumgrey;
       text-decoration: none;
+      transition: $standard-transition;
+
+      &:before {
+        display: none;
+      }
```
LeoniePeters, line 201: Reconsider using .nav-link:before in the first place, as it's now set more often to display:none than actually used. The .nav-link:before selector is set to “display:none” on both PageNavigation and tabs.scss. It’s only used on the ViewsToggle I believe. So I think we can make that all more specific.
Remove the .nav-link selector here https://github.com/europeana/portal.js/blob/master/assets/scss/buttons-links.scss#L142 
And edit the mentioned components and tabs.scss

LeoniePeters (changes requested):

components/PageNavigation.vue
mirjamverloop: That is a good suggestion, thanks!

components/PageNavigation.vue
```diff
@@ -194,11 +194,6 @@
     .nav-link {
       color: $mediumgrey;
       text-decoration: none;
-      transition: $standard-transition;
```
LeoniePeters: Did you remove this line by mistake? This removed the smooth transition of the color, which I liked.

LeoniePeters (approved):

PR #968, opened 2021-01-21
Add close button to nav

alansutherland:


components/PageHeader.vue
```diff
@@ -167,7 +175,6 @@
     left: 0;
     z-index: 1030;
     padding: 0;
-    border-bottom: 1px solid $whitegrey;
```
LeoniePeters, line 170: Not sure what this border used to do, but just checking if we should really get rid of it?

mirjamverloop: Yes, it was recommended by Romain. There will be a box-shadow instead when merged in with master.

components/PageHeader.vue
```diff
@@ -303,13 +303,21 @@
       width: 1.5rem;
       height: 1.5rem;
       box-shadow: none;
+      transition: $standard-transition;
```
LeoniePeters, line 306: The transition should be on the span to work properly

LeoniePeters (changes requested):

LeoniePeters (approved):

PR #970, opened 2021-01-29
Hardcode topnav links

lbiedinger:


components/PageHeader.vue
```diff
@@ -122,19 +120,20 @@
...
-    },
-
     data() {
       return {
+        mainNavigation: [
+          { url: '/collections', text: this.$t('header.navigation.collections') },
+          { url: '/europeana-classroom', text: this.$t('header.navigation.europeanaClassroom') },
+          { url: '/about-us', text: this.$t('header.navigation.about') }
+        ],
+        mobileNavigation: [
```
mirjamverloop, line 130: Perhaps slightly out of scope, but this navigation will not be limited to mobile as we are planning to introduce the hamburger on desktop too. So perhaps call this something like `sidebarNavigation` instead of `mobileNavigation`?

mirjamverloop (approved):

rwd (approved):

PR #973, opened 2021-02-02
Enable YouTube embedding

mirjamverloop:


components/generic/HTMLEmbed.vue
```diff
@@ -18,18 +30,42 @@
...
+        default: null
+      },
       error: {
         type: [String, Error],
         default: ''
       }
+    },
+
+    computed: {
+      aspectRatio() {
```
rwd, line 62: This does not return the aspect ratio, so should be renamed.

Perhaps `heightAsPercentOfWidth` returning `(this.height * 100) / this.width`, with the `padding-bottom` moved into the style attribute on the youtube embed div.

components/generic/HTMLEmbed.vue
```diff
@@ -1,7 +1,19 @@
 <template>
   <!-- eslint-disable vue/no-v-html -->
   <div
-    v-if="html"
+    v-if="html && provider === 'YouTube'"
```
rwd, line 4: Is this really only useful / needed in the case of the youtube embed, or can it be applied more generally to other embeds?

rwd (approved):

alansutherland (approved):

PR #979, opened 2021-02-08
Refresh likes set after unliking

mirjamverloop:


store/set.js
```diff
@@ -54,22 +54,37 @@ export const actions = {
...
+      })
+      .catch(() => {
+        return dispatch('fetchLikes');
       });
   },
-  unlike({ commit, state }, itemId) {
+  unlike({ dispatch, commit, state }, itemId) {
     return this.$apis.set.modifyItems('delete', state.likesId, itemId)
-      .then(commit('unlike', itemId));
+      .then(commit('unlike', itemId))
```
rwd, line 64: I think this can be reduced to:
```js
      .then(dispatch('fetchLikes'))
      .catch(dispatch('fetchLikes'));
```
... which is more readable. Likewise the other changes to this file.

rwd (changes requested):

store/set.js
LeoniePeters: That breaks the error handling again. 'then' and 'catch' expect callbacks. I propose to leave as is.

rwd: Indeed. Thanks for the sanity check @LeoniePeters !

rwd (approved):

PR #988, opened 2021-02-15
Add alt description for Contentful images

mirjamverloop:


pages/blog/index.vue
```diff
@@ -95,6 +96,12 @@
           return post.primaryImageOfPage.image.contentType;
         }
         return null;
+      },
+      imageAlt(post) {
+        if (post.primaryImageOfPage && post.primaryImageOfPage.image.description) {
```
alansutherland, line 101: wouldn't `if (post.primaryImageOfPage.image.description)` suffice?

LeoniePeters: If there is no primaryImageOfPage you'd get an error, something like "Can't read property image of undefined".

pages/exhibitions/index.vue
```diff
@@ -91,6 +92,9 @@
       },
       imageContentType(image) {
         if (image && image.image) return image.image.contentType;
+      },
+      imageAlt(image) {
+        if (image && image.image && image.image.description) return image.image.description;
```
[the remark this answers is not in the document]
LeoniePeters: The same counts here as stated in the comment above.

rwd (approved):

PR #995, opened 2021-02-18
Prevent download of isShownAt

alansutherland:


rwd (changes requested):
This does not work as required. It prevents downloading of any media from items having an isShownAt instead of just preventing downloading the isShownAt.

See for example item /400/0264_001_FP where this prevents downloading its isShownBy.

rwd (approved):

PR #1004, opened 2021-02-24
Disable download and link when item is isShownAt

LeoniePeters:


pages/item/_.vue
```diff
@@ -141,7 +140,12 @@
     asyncData({ params, res, app, $apis }) {
       return $apis.record
         .getRecord(`/${params.pathMatch}`, { locale: app.i18n.locale })
-        .then(result => result.record)
+        .then(result => {
+          result.record.media && result.record.media.map(media => {
+            if (media.about === result.record.isShownAt) media.isShownAt = true;
+          });
+          return result.record;
+        })
```
lbiedinger, line 148: instead of doing this on the page here, it would make more sense in the plugins/europeana/record.js file.
Specifically in the aggreagtionMedia method.

lbiedinger (changes requested):

rwd (approved):

lbiedinger (approved):

PR #1015, opened 2021-03-08
Re-implement grid using vue-masonry

mirjamverloop:


components/generic/ContentCard.vue
```diff
@@ -143,7 +143,7 @@
       },
       lazy: {
         type: Boolean,
-        default: true
+        default: false
```
mirjamverloop, line 146: This value should be true by default, false should be passed in.

nuxt.config.js
```diff
@@ -205,7 +205,8 @@ module.exports = {
     '~/plugins/page',
     '~/plugins/vue-filters',
     '~/plugins/vue-directives',
-    { src: '~/plugins/vue-announcer', mode: 'client' }
+    { src: '~/plugins/vue-announcer', mode: 'client' },
+    { src: '~/plugins/vue/vue-masonry', ssr: false }
```
LeoniePeters, line 209: We are now using three different syntaxes in the plugin definitions for specifying if a plugin should be loaded only on client side. Let's choose one!
```javascript
'~/plugins/hotjar.client'
{ src: '~/plugins/vue-announcer', mode: 'client' },
{ src: '~/plugins/vue/vue-masonry', ssr: false }
```
See https://nuxtjs.org/docs/2.x/directory-structure/plugins/#name-conventional-plugin
and
https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-plugins/

LeoniePeters (changes requested):

nuxt.config.js
mirjamverloop: Good point, thanks!

LeoniePeters (approved):

lbiedinger (approved):

PR #1018, opened 2021-03-11
Truncate menu items

alansutherland:


components/PageNavigation.vue
```diff
@@ -272,6 +272,7 @@
       margin: 0 0 0.25rem 0;
       position: relative;
       margin-right: 0;
+      margin-bottom: 0;
```
LeoniePeters, line 275: I don't see why this is needed.
If it is, in that case you could just set margin: 0 and remove margin-right and margin-bottom.

components/PageNavigation.vue
```diff
@@ -295,6 +296,16 @@
             color: $white;
           }
         }
+        span {
+          overflow: hidden;
+          white-space: nowrap;
+          display: block;
```
LeoniePeters, line 302: Remove this display: block. It overwrites the above (line 222) set display: flex which keeps the icon and text properly aligned and the element height correct.

components/PageNavigation.vue
```diff
@@ -295,6 +296,16 @@
...
+          text-overflow: ellipsis;
+          height: 1.5rem;
```
LeoniePeters, line 304: Remove this height as it's unnecessary when display is set to flex as mentioned above.

components/PageNavigation.vue
```diff
@@ -295,6 +296,16 @@
...
+          i {
+            margin-right: 0.5rem;
```
LeoniePeters, line 306: Is this needed?

LeoniePeters (changes requested):

mirjamverloop (approved):

PR #1023, opened 2021-03-22
Keyboard accessibility hamburger menu

LeoniePeters:


components/PageHeader.vue
```diff
@@ -27,57 +27,35 @@
...
-      <b-navbar
-        class="align-items-center flex-row d-flex p-0 mr-3"
-        role="navigation"
+      <b-sidebar
+        id="sidebar"
+        title="Side navigation"
+        bg-variant="white"
+        no-header
+        backdrop
+        width="16rem"
```
lbiedinger, line 46: could this be applied as styling?

components/PageHeader.vue
```diff
@@ -27,57 +27,35 @@
...
-          src="../assets/img/logo.svg"
-          :alt="$t('homeLinkAlt')"
-          class="mw-100"
-          data-qa="logo"
-        >
-      </SmartLink>
-      <b-navbar
-        class="align-items-center flex-row d-flex p-0 mr-3"
-        role="navigation"
+      <b-sidebar
```
lbiedinger, line 40: running AXE with the sidebar open highlights a newly introduced issue.
Specifically no aria-label attribute.

components/PageHeader.vue
LeoniePeters: Yes, width can be set as styling. The bg-variant is better set here as it would turn to its default otherwise and override custom styling. no-header and backdrop also makes more sense to make use of here.

lbiedinger (approved):

assets/scss/sidebar.scss
```diff
@@ -0,0 +1,19 @@
+.b-sidebar-backdrop.bg-dark {
+  background-color: rgb(0, 0, 0) !important;
```
mirjamverloop, line 2: To avoid the !important, it is also possible to add our own backdrop variant, for example

```
<b-sidebar
        id="sidebar"
        bg-variant="white"
        no-header
        backdrop
        backdrop-variant="black"
        aria-label="Side navigation"
      >
```
The css could then change to:

```
b-sidebar-backdrop.bg-black {
  background-color: rgb(0, 0, 0);
  ...
```

assets/scss/sidebar.scss
```diff
@@ -0,0 +1,19 @@
...
+  max-height: 100vh;
+  transition: $standard-transition; // fixes header appear/disappear
+}
+// fixes header appear/disappear
+.container-fluid:not(.show) {
+  #sidebar, .b-sidebar-backdrop {
+    transform: translateY(3.5rem);
+    transition: $standard-transition;
+  }
+}
```
mirjamverloop, line 18: Not sure if it is because of any of the styling here, but the sidebar has a horizontal scrollbar on larger screens. I have tested this in Chrome and Firefox. Mobile is fine, but desktop sizes shows a scrollbar.

LeoniePeters: Well spotted !

mirjamverloop (approved):

PR #1052, opened 2021-04-01
Authors separated by commas

alansutherland:


rwd:
I’ve made changes so that:

1. BlogAuthor does not need to know how many authors there are in total, as it should not be concerned with that
2. BlogAuthor uses the address element instead of div
3. BlogPost adds the commas where needed
4. BlogPost includes BlogAuthor as inline so multiple authors are presented on a single line

mirjamverloop (approved):

PR #1064, opened 2021-04-22
Experiment item swiper height

LeoniePeters:


mirjamverloop:
We decided to go for the other option, see #1065,  so I have closed this one.

PR #1067, opened 2021-04-26
Entity description edit

annachristaki:


src/pages/collections/_type/_.vue
```diff
@@ -14,6 +14,23 @@
               :is-editorial-description="hasEditorialDescription"
               :title="title"
             />
+            <template
+              v-if="isEditable && userIsEditor"
+            >
+              <div class="d-inline-flex">
+                <b-button
+                  variant="outline-primary"
+                  class="text-decoration-none mr-2"
```
spyrosbek, line 23: I think div doesn't really need a display class here.
The mr-2 class in button is also unnecessary.

src/components/entity/EntityUpdateModal.vue
```diff
@@ -0,0 +1,116 @@
...
+        />
+      </b-form-group>
+      <div class="modal-footer">
+        <b-button
+          variant="outline-primary"
+          @click="$bvModal.hide('entityUpdateModal')"
+        >
+          {{ $t('actions.close') }}
+        </b-button>
+        <div class="d-flex">
```
spyrosbek, line 29: This whole div is not needed.
Both buttons can be at the same level, under the 
> <div class="modal-footer">

src/components/entity/EntityUpdateModal.vue
```diff
@@ -0,0 +1,116 @@
...
+        label-for="entity-description"
+      >
+        <b-form-textarea
+          id="entity-description"
+          v-model="descriptionValue"
+          maxlength="240"
+          rows="4"
+        />
+      </b-form-group>
+      <div class="modal-footer">
```
spyrosbek, line 22: You can add `pt-2` in the classes, to get the correct spacing.

spyrosbek (changes requested):
Just a couple of changes in the html

spyrosbek (approved):

src/components/entity/EntityUpdateModal.vue
```diff
@@ -0,0 +1,114 @@
...
+          toaster: 'b-toaster-bottom-left',
+          autoHideDelay: 5000,
+          isStatus: true,
+          noCloseButton: true,
+          solid: true
+        });
+      },
+      submitForm() {
+        const handler = this.$store.dispatch('entity/updateEntity', { id: this.body.id, body: this.entityBody });
+        return handler.then(() => {
```
rwd, line 106: This needs to handle errors gracefully and in a user-friendly manner. To force an error to occur, change in `entityBody`: `type: 'this.body.type'`

src/pages/collections/_type/_.vue
```diff
@@ -78,27 +94,29 @@
   import { BASE_URL as EUROPEANA_DATA_URL } from '../../../plugins/europeana/data';
   import { getEntityTypeHumanReadable, getEntitySlug, getEntityUri } from '../../../plugins/europeana/entity';
   import { langMapValueForLocale } from  '../../../plugins/europeana/utils';
+  import EntityUpdateModal from '../../../components/entity/EntityUpdateModal';
```
rwd, line 97: As this will not be visible to most users, lazy-load it, as with BrowseSections and RelatedCollections

rwd (changes requested):

src/components/entity/EntityUpdateModal.vue
annachristaki: What is a user friendly manner to handle errors when the user interacts with the modal? I'm not sure if/how the error should be displayed.

rwd: As there is no design for it, I have pushed a change to just throw the error when the promise is rejected, which results in the error being displayed in the main body of the page, and is what we do elsewhere with modals encountering API errors.

rwd (approved):

PR #1071, opened 2021-05-04
Referrer in Keycloak account URL

LeoniePeters:


src/mixins/keycloak.js
```diff
@@ -0,0 +1,34 @@
...
+      return redirect;
+    },
+
+    keycloakAccountUrl() {
+      const keycloakAccountUrl = new URL(
+        `/auth/realms/${this.$auth.strategy.options.realm}/account`, this.$auth.strategy.options.origin
+      );
+      keycloakAccountUrl.search = new URLSearchParams({
+        referrer: this.$auth.strategy.options.client_id,
+        'referrer_uri': this.$config.app.baseUrl
```
LeoniePeters, line 21: The baseUrl is passed in here as the referrer uri, which I believe is always set to production, but shouldn't we want this to be more dynamic and it linking back to test or localhost when navigating from there?

mirjamverloop (approved):

src/mixins/keycloak.js
rwd: It's set by env var `PORTAL_BASE_URL`, so if you want it to be your localhost URL, set it to be so. :smiley:

rwd (approved):

PR #1082, opened 2021-05-12
chore: remove old cookie disclaimer

rwd:


rwd:
Made a mess of the mainline merge, will start anew.

PR #1083, opened 2021-05-12
JIRA service desk api endpoint

lbiedinger:


src/server-middleware/jira-service-desk-api.js
```diff
@@ -0,0 +1,51 @@
...
+          requestTypeId: '81', // Type 81 is 'suggestion'
+          requestFieldValues: {
+            summary: parsedData.summary
+          }
+        };
+        if (parsedData.email) {
+          createRequestData.raiseOnBehalfOf = parsedData.email;
+        }
+
+        axios.post('https://europeana.atlassian.net/rest/servicedeskapi/request', createRequestData, {
```
lbiedinger, line 25: Should the endpoint could be passed via env var?

rwd: Yes, and is in my refactor.

src/server-middleware/api/index.js
```diff
@@ -0,0 +1,25 @@
...
+    runtimeConfig = defu(nuxtConfig.privateRuntimeConfig, nuxtConfig.publicRuntimeConfig);
+  }
+  next();
+});
+
+const debugMemoryUsage = require('./debug/memory-usage');
+app.get('/debug/memory-usage', debugMemoryUsage);
+
+const jiraServiceDesk = require('./jira/service-desk');
+app.post('/jira/service-desk', (res, req, next) => jiraServiceDesk(runtimeConfig.jira)(res, req, next));
```
lbiedinger, line 21: SonarCloud is raising a warning on the param order here.

Additionally, in fact I believe that we don't need to pass 'next' at all either.

rwd: I have addressed the security hotspot, which is about the use of `x-powered-by` header, and also removed `next` from the arguments.

mirjamverloop (approved):

PR #1086, opened 2021-05-13
Add create gallery card to public/private user galleries

alansutherland:


src/components/account/CreateSetButton.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <b-card
+    class="text-center default-card content-card create-card"
```
mirjamverloop, line 3: The class "default-card" can be removed here.

src/components/account/CreateSetButton.vue
```diff
@@ -0,0 +1,60 @@
...
+    no-body
+    data-qa="create new gallery card"
+    @click="clickCreateSet"
+  >
+    <div class="card-inner d-flex justify-content-center align-items-center">
+      <span class="icon-ic-add" />
+      <b-card-body data-qa="card body">
+        <b-card-title title-tag="div">
+          <span>
```
mirjamverloop, line 12: What is the purpose of the span here? 
Could the span be removed or could the title-tag be set to a span instead of a div directly?

src/components/account/UserButtons.vue
```diff
@@ -89,13 +89,20 @@
...
+        this.clickCreateSet();
+      });
+    },
     methods: {
       clickCreateSet() {
-        this.showFormModal = true;
-        this.newSetCreated = false;
-        this.$bvModal.hide(this.addItemToSetModalId);
-        this.$bvModal.show(this.setFormModalId);
+        console.log('clicked');
```
mirjamverloop, line 99: Remove the console.log.

mirjamverloop (changes requested):

src/components/account/CreateSetButton.vue
LeoniePeters: The span is also (unnecessarily?) there on the ContenCard titles, so we just copied it from there. I removed the span for the CreateSet and that seems fine indeed.

mirjamverloop (approved):

lbiedinger (approved):

PR #1091, opened 2021-05-24
pin/unpin entity items

annachristaki:


src/components/account/UserButtons.vue
```diff
@@ -42,6 +56,29 @@
...
+              {{ $t('actions.close') }}
+            </b-button>
+            <b-button
+              variant="primary"
+              @click="viewPins()"
+            >
+              {{ $t('entity.actions.viewPinned') }}
+            </b-button>
+          </div>
+        </b-modal>
```
spyrosbek, line 81: This could possibly be removed and be merged with the new PinModal component, by changing PinModal' description and the labels of its buttons.
I think this code could be avoided here (since there is a modal for the workflow), unlike the one above about likes, where clicking the like button doesn't open a modal normally.

spyrosbek (changes requested):

src/components/account/UserButtons.vue
```diff
@@ -68,6 +106,10 @@
       value: {
         type: String,
         required: true
+      },
+      showPins: {
```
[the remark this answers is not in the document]
annachristaki: I think the obvious downside makes this a worse solution

src/pages/collections/_type/_.vue
```diff
@@ -164,7 +167,9 @@
         entity: state => state.entity.entity,
         page: state => state.entity.page,
         relatedEntities: state => state.entity.relatedEntities,
-        recordsPerPage: state => state.entity.recordsPerPage
+        recordsPerPage: state => state.entity.recordsPerPage,
+        featuredSetId: state => state.entity.featuredSetId,
+        pinned: state => state.entity.pinned
```
LeoniePeters, line 172: I don't see `featuredSetId` and `pinned` being used in this component. Are they necessary?

src/components/entity/PinModal.vue
```diff
@@ -0,0 +1,104 @@
...
+          await this.$store.dispatch('entity/pin', this.itemId);
+          this.hide();
+          const msg = this.$store.state.sanitised.page === 1 ?  this.$t('entity.notifications.pinnedFirstPage') : this.$t('entity.notifications.pinned');
+          this.makeToast(msg);
+        } catch (e) {
+          if (e.message === 'too many pins') {
+            this.hide();
+            this.$bvModal.show(`pinned-limit-modal-${this.itemId}`);
+          } else {
+            throw e;
```
LeoniePeters, line 85: Let's add an error message / toast for the user if the pinning fails. Something like "The item was not pinned. Something went wrong. Please try again."

LeoniePeters (changes requested):

src/pages/collections/_type/_.vue
```diff
@@ -207,6 +212,9 @@
         }
         return false;
       },
+      userIsEditor() {
+        return this.$store.state.auth.user && this.$store.state.auth.user.resource_access.entities && this.$store.state.auth.user.resource_access.entities.roles.includes('editor');
```
LeoniePeters, line 216: extend this with a check to see if a user has the editor role for both entities and usersets

LeoniePeters (changes requested):

src/pages/collections/_type/_.vue
annachristaki: You are right, I  removed them

src/components/entity/PinModal.vue
annachristaki: This is not included in the designs. I made an update so that if an error occurs it will be displayed to the user, same way you handle api errors in other modals.

src/pages/collections/_type/_.vue
annachristaki: Done

lbiedinger (changes requested):

lbiedinger (approved):

PR #1092, opened 2021-05-25
Add a 'Curated Collections' tab in account page for editors

spyrosbek:
Add a section to My Profile page for entity-related galleries


src/pages/account/index.vue
```diff
@@ -152,6 +174,15 @@
     },
 
     computed: {
+      keycloakAccountUrl() {
```
annachristaki, line 177: This is already defined elsewhere, remove it

annachristaki (changes requested):
Looks good, only one correction.

src/components/account/UserSets.vue
```diff
@@ -13,6 +13,17 @@
           class="card-deck-4-cols pb-5"
           deck
         >
+          <b-row
+            v-if="visibility === 'curated'"
+            class="w-100 px-3"
+          >
+            <b-col>
+              <h2 class="related-heading text-uppercase">
+                <span class="icon-info-outline" />
```
LeoniePeters, line 22: There is an 'i' icon in front of this text (see the design). I asked the designers and it would be best to use the already existing info icon from our icon font. You can use the `.icon-info` class and or `content: '\e91f'`

LeoniePeters (changes requested):

LeoniePeters (approved):

annachristaki (approved):

PR #1104, opened 2021-06-01
Add context labels

mirjamverloop:


src/components/authored/AuthoredHead.vue
```diff
@@ -7,6 +7,12 @@
       >
         <article>
           <div class="title">
+            <div
+              v-if="contextLabel"
```
LeoniePeters, line 11: Do we need these `v-if`'s? As far as I can see the context-label prop is set by translation variables and are always set. Same for EntityDetails.vue and ContentHeader.vue

LeoniePeters (changes requested):

src/assets/scss/texts.scss
```diff
@@ -149,3 +149,15 @@ header .col {
...
+.context-label {
+  font-size: $font-size-extrasmall;
+  text-transform: uppercase;
+  font-weight: 600;
+  color: $mediumgrey;
+  line-height: initial;
+}
+
+.bg-white .context-label {
+  margin-bottom: -1rem;
```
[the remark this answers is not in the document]
mirjamverloop: Yes, good idea, that would work.

src/components/authored/AuthoredHead.vue
mirjamverloop: For example, the ContentHeader.vue is also used on the exhibition landing page (en/exhibitions) where there is no label. If I would remove the v-if it shows an empty div in the code. Similar to browse pages, they use the ContentHeader.vue, and don't have a label. And the static pages use AuthoredHead.vue and don't have a label. So, it feels "safer" to keep the if in the ContentHeader.vue and AuthoredHead.vue. 

I have removed the if from the EntityDetails component.

rwd (approved):

PR #1117, opened 2021-06-10
Only show feedback widget after cookie disclaimer accepted

rwd:
As otherwise widget may obscure disclaimer dismissal button.


lbiedinger (approved):

rwd:
This fails to account for scenarios in which there has been no interaction with the cookie disclaimer but it's not visible, e.g. with the "I don't care about cookies" browser extension.

PR #1119, opened 2021-06-10
feedback button and cookie disclaimer positioning

LeoniePeters:


rwd:
This fails to account for scenarios in which there has been no interaction with the cookie disclaimer but it's not visible, e.g. with the "I don't care about cookies" browser extension.

PR #1120, opened 2021-06-10
Script to pull Lokalise translations

rwd:
And output in the src/lang/*.js format we use


bin/lokalise.js
```diff
@@ -0,0 +1,89 @@
...
+
+  return str;
+};
+
+const run = async() => {
+  const i18n = {};
+  let items;
+  let page = 0;
+
+  while (!Array.isArray(items) || items.length > 0) {
```
lbiedinger, line 54: is items not an array only on the first run, or can it come back like that from the fetchPage request?

rwd: It's not an array on the first run because it's just been initialised on line 51, but not set to anything yet.

rwd:
Superseded by #1669

PR #1125, opened 2021-06-15
Organisation detail page

mirjamverloop:
- Collection detail pages now also show organisation
- Including depiction, description and external link (all optional)


src/store/search.js
```diff
@@ -22,6 +22,10 @@ export const defaultFacetNames = [
 const filtersFromQf = (qfs) => {
   const filters = {};
 
+  if (!qfs || !qfs[0]) {
```
rwd, line 25: Why is this needed?

mirjamverloop: It breaks on organisation pages without a check.
qfs could be null on these pages.

rwd: I have pushed a change to address (https://github.com/europeana/portal.js/pull/1125/commits/4e640813aa1d1c08c7f775a0cc59fb1504e701bc) this in a different way, returning the fielded filter needed (I think) for the organization URIs.

lbiedinger (approved):

PR #1129, opened 2021-06-18
Harvest organisations to Redis

rwd:
1. Adds Redis as a dependency, configured with `REDIS_URL` and `REDIS_TLS_CA` (optional) env vars
2. Includes Docker Compose file to run Redis in dev environments, with `docker-compose up` from the project root. Redis will then be at `REDIS_URL=redis://localhost:16379`
1. Introduces a directory `src/cachers` which contains scripts for collecting data, writing it to a Redis cache, and getting it back out again.
2. Includes a CLI for running these scripts e.g.:
    ```
    npm run cache entities:organisations:get
    npm run cache entities:organisations:set
    ```
3. `entities:organisations:set` paginates over the Europeana Entity API for organization-type entities, reduces these to an object keyed by identifier with values containing the prefLabel, and writes to the cache
4. `entities:organisations:get` reads from the cache and outputs to the console
4. Adds a server middleware at `/_api/entities/organisations` to output in JSON the harvested organisations, reusing the retrieval code from the `entities:organisations:get` cacher
4. Adds a server middleware at `/_api/version` to output the portal.js app version This is intended for future use in scheduled jobs that need to check the production website for its version before running the CLI against that version to update the Redis cache.


src/cachers/entities/organisations/set.js
```diff
@@ -0,0 +1,84 @@
...
+
+const persistableFields = ({ identifier, prefLabel }) => {
+  return {
+    identifier,
+    prefLabel
+  };
+};
+
+const writeToRedis = (organisations) => {
+  return redisClient.setAsync(utils.CACHE_KEY, JSON.stringify(organisations, null, 2))
```
lbiedinger, line 51: the extra spacing passed as an option to stringify here results in considerable increase in the amount of data stored.
Without it, the whole organisation data string is 393.46kB while with it I get 516.5kB
It shouldn't really be necessary to read what's in this redis cache manually.

src/cachers/index.js
```diff
@@ -0,0 +1,32 @@
...
+  return cli[process.argv[2]].cli();
+};
+
+main()
+  .then(({ body }) => {
+    if (typeof body === 'string') {
+      console.log(`SUCCESS: ${body}`);
+    } else {
+      console.log('SUCCESS:');
+      console.log(JSON.stringify(body, null, 2));
```
lbiedinger, line 25: other than debugging the functionality introduced with this PR, is there actually a purpose to logging the `get` response to the console?

lbiedinger (changes requested):

src/cachers/index.js
```diff
@@ -0,0 +1,32 @@
+if (process.env.NODE_ENV !== 'production') {
+  require('dotenv').config();
+}
```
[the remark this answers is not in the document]
rwd: The expectation is that the cachers are run using the npm script, so this is the only place it's needed.
```
npm run cache entities:organisations:get
npm run cache entities:organisations:set
```

src/cachers/index.js
rwd: No.

src/server-middleware/api/entities/organisations.js
```diff
@@ -1,8 +1,12 @@
 const getOrganisations = require('../../../cachers/entities/organisations/get');
-const { errorHandler } = import('../');
+import { errorHandler } from '../';
 
 export default (options = {}) => (req, res) => {
-  getOrganisations({
+  if (!options.redis.url) {
```
lbiedinger, line 5: this breaks the unit test. Just needs to be stubbed there.

rwd: Actually, it was the removal of prettification of the JSON that broke it. Fixed, and apologies for not testing first.

src/cachers/index.js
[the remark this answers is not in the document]
lbiedinger: A bit strange, I was getting errors before due to what I thought was this also when running the npm script as in your comment, but can no longer reproduce.
Seems to have been unrelated to that actual change.

lbiedinger (approved):

LeoniePeters (approved):

PR #1131, opened 2021-06-25
Update nightwatch-accessibility for axe 4

rwd:
Also start validating against WCAG 2.1 Level A & AA Rules

### DIFF
--- a/tests/e2e/docker/nightwatch/support/step-runners.js
+++ b/tests/e2e/docker/nightwatch/support/step-runners.js
@@ -35,14 +35,11 @@ module.exports = {
     await client.waitForElementVisible(activeLinkSelector);
   },
   async checkPageAccesibility() {
-    let axeOptions = {
+    const axeOptions = {
       reporter: 'v2',
       runOnly: {
         type: 'tags',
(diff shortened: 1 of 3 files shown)


tests/e2e/docker/nightwatch/support/step-runners.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
       reporter: 'v2',
       runOnly: {
         type: 'tags',
-        values: ['wcag2a', 'wcag2aa']
+        values: ['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa']
       },
       rules: {
         'aria-roles': { enabled: false } // https://github.com/bootstrap-vue/bootstrap-vue/issues/2921 + https://github.com/dequelabs/axe-core/issues/1462
```
LeoniePeters, line 45: I believe we can get rid of this rule as the issue seems to have been resolved.

LeoniePeters (approved):

tests/e2e/docker/nightwatch/support/step-runners.js
rwd: Yes, indeed, and that's done. Thanks!

lbiedinger (approved):

PR #1135, opened 2021-06-28
Update feedback button

LeoniePeters:


src/components/feedback/FeedbackWidget.vue
```diff
@@ -204,6 +205,17 @@
...
+      },
+      buttonClasses() {
+        return {
+          'hide-button': this.showWidget,
+          big: this.beforeScroll
+        };
+      }
+    },
+    mounted() {
+      if (process.browser) {
```
rwd, line 217: This check should not be necessary as mounted is only ever run client-side.

rwd (changes requested):

rwd (approved):

PR #1146, opened 2021-07-05
Translate organization entity URIs

rwd:


src/plugins/europeana/record.js
```diff
@@ -20,15 +20,20 @@ export const BASE_URL = process.env.EUROPEANA_RECORD_API_URL || 'https://api.eur
...
     dcPublisher: proxyData.dcPublisher,
     dcSubject: proxyData.dcSubject,
     dcType: proxyData.dcType,
     dctermsMedium: proxyData.dctermsMedium
-  }, isUndefined), entities));
+  }, isUndefined), entities);
+
+  fields.edmDataProvider = {
+    url: providerAggregation.edmIsShownAt, value: fields.edmDataProvider
+  };
```
mirjamverloop, line 35: Wouldn't this always override the `edmDataProvider` field above?

rwd: It does the entity lookup first, above, then preserves the result of that here in `value`, adding isShownAt to `url`.

mirjamverloop (approved):

PR #1168, opened 2021-07-21
Account for creator object in attribution

rwd:


src/components/generic/DownloadModal.vue
```diff
@@ -83,7 +83,9 @@
 
     data() {
       return {
-        snippetCopied: false
+        snippetCopied: false,
+        providerString: this.stringify(this.provider),
+        creatorString: this.stringify(this.creator)
```
LeoniePeters, line 88: Shouldn't these be computed properties?

rwd: They do not need to be reactive, so I don't particularly think so.

lbiedinger (approved):

PR #1174, opened 2021-07-26
Translations on item pages

lbiedinger:


src/plugins/europeana/record.js
```diff
@@ -232,7 +232,7 @@ export default (context = {}) => {
         // refactor to maintain the source info without having to set this.
         const europeanaProxy = edm.proxies.find(proxy => proxy.europeanaProxy);
         ['dcTitle', 'dcDescription'].forEach((field) => {
-          if (europeanaProxy[field]) {
+          if (europeanaProxy && europeanaProxy[field]) {
```
LeoniePeters, line 235: This could be `europeanaProxy?.[field]`

tests/unit/pages/item/_.spec.js
```diff
@@ -64,19 +64,34 @@ const factory = () => shallowMountNuxt(page, {
...
+
+        $apis.record.getRecord.should.have.been.calledWith('/123/abc', { locale: 'en', metadataLang: undefined });
+        response.should.eql(record);
+      });
+    });
+    context('when the page is loaded with a metadataLang', () => {
+      const route = { query: { metadataLang: 'fr' } };
 
-      const response = await wrapper.vm.asyncData({ params, app, route, $apis });
+      it('gets a record from the API for the ID in the params pathMatch, for the current locale', async() => {
```
LeoniePeters, line 87: Rewrite this scenario to specify it more properly. Something like 'gets a record from the API for the ID in the params pathMatch, for the current locale **and its requested translation**'

LeoniePeters (changes requested):

LeoniePeters (approved):

PR #1176, opened 2021-07-27
UI multilingual item page

LeoniePeters:
- Adds new component ItemLanguageSelector
- refactored reused locales function into locales mixin
- Adds buttons with tooltips for translated title and description fields


src/components/item/ItemLanguageSelector.vue
```diff
@@ -0,0 +1,96 @@
...
+      }
+    },
+    data() {
+      return {
+        translated: !!this.$route.query.metadataLang || false
+      };
+    },
+    computed: {
+      languageToggle() {
+        return this.translated ? 'original language' : this.selectedLocale.name;
```
lbiedinger, line 71: 'original language' needs to go into the locale file.

src/components/item/ItemLanguageSelector.vue
```diff
@@ -0,0 +1,96 @@
...
+    ],
+    props: {
+      itemLanguage: {
+        type: String,
+        default: null
+      }
+    },
+    data() {
+      return {
+        translated: !!this.$route.query.metadataLang || false
```
lbiedinger, line 66: The translate profile is always going to be on, I think it's still useful to track translation status, but then let that signify that the item is currently _not_ shown in it's original metadata (edm) language. 
In which case this should be a computed property.

src/components/item/ItemLanguageSelector.vue
```diff
@@ -0,0 +1,96 @@
...
+            <i18n
+              path="multilingual.currentLanguage"
+              tag="span"
+              class="pr-1"
+              data-qa="translate item suggestion"
+            >
+              <span v-if="!translated && itemLanguage === selectedLocale.code">
+                a different language
+              </span>
+              <b-link
```
lbiedinger, line 20: It probably makes more sense to have this and the links in the dropdown to be Nuxt-links with a :to property that can call a method 'translateParams' where it can return:
```
translateParams(language) {
  const queryParams = {};
  if (language) queryParams.metadataLang = language;
  return { path: this.$route.path, query: queryParams };
}
```
That way 'metadataLang' is removed also when  'resetting' to the edmLanguage.

src/pages/item/_.vue
```diff
@@ -285,7 +291,10 @@
...
       ...mapGetters({
         shareUrl: 'http/canonicalUrl'
-      })
+      }),
+      multilingualItemEnabled() {
+        return this.$config.app.features.multilingualItem;
+      }
     },
 
     mounted() {
```
lbiedinger, line 300: bellow in this file 
`watchQuery: ['metadataLang'],`
will need to be replaced with:
```
watch: {
  '$route.query.metadataLang': '$fetch'
},
```

To trigger the re-fetching of new metadata when the toggle is used..

That still doesn't quite update the whole UI properly though, I'm not sure but probably because metadataLang isn't a computed property, but just read from the URL query.

lbiedinger (changes requested):

src/assets/scss/tooltip.scss
```diff
@@ -0,0 +1,12 @@
+.tooltip.b-tooltip {
+  margin-left: 1rem;
+  opacity: 1;
+  .tooltip-inner {
+    font-size: 1rem;
+    font-weight: 400;
+    padding: 10px;
```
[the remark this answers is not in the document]
LeoniePeters: The padding is specific to these tooltips, but I'll change it to use rem. We are not using any variables for font-weight, which would be a bit overkill as we are just using 400, 600, 700 (or normal, semibold, bold)

src/components/item/SummaryInfo.vue
```diff
@@ -100,6 +122,17 @@
           return this.$options.filters.truncate(this.description.values[0], this.limitCharacters, this.$t('formatting.ellipsis'));
         }
         return false;
+      },
+      translationSourceTitle() {
```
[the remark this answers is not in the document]
LeoniePeters: Added 'original' as fallback if x.translationSource undefined

src/pages/item/_.vue
LeoniePeters: As we also need to call asyncData I'm leaving this as is.

PR #1187, opened 2021-08-17
Hard coded footer links

LeoniePeters:


src/components/PageFooter.vue
```diff
@@ -102,29 +102,8 @@
       LinkGroup
     },
 
-    fetch() {
-      const contentfulVariables = {
-        locale: this.$i18n.isoLocale(),
-        preview: this.$route.query.mode === 'preview'
-      };
-
-      return this.$contentful.query('linkGroups', contentfulVariables)
```
lbiedinger, line 111: this was the only place from where `src/modules/contentful-graphql/queries/link-groups.graphql` was called, that can be removed as part of this PR.

The other place link groups are used is on static pages, but they are included in the specific graphql query for those.

lbiedinger (approved):

lbiedinger (changes requested):
I mistakenly approved this.
The graphql still needs to be removed on this branch. Also there are questions about the loss of some of the translations if this were to go in as is.

lbiedinger (approved):

PR #1194, opened 2021-08-24
A/B testing with Matomo

mirjamverloop:
- Re-use experiments folder
- Plugin for basic AB test setup


src/plugins/ab-testing.js
```diff
@@ -0,0 +1,59 @@
...
+  // is there an experiment?
+  if (experiments.length === 0) {
+    inject('experiments', () => {
+      return false;
+    });
+    return false;
+  }
+  // check consent
+  const consentCookie = ctx.$cookies.get('klaro');
+  if (!consentCookie?.matomo) {
```
LeoniePeters, line 13: The matomo AB testing cookie might/should be separate from the matomo analytics cookie. So this check should change once we do the set up with the plugin

src/plugins/ab-testing.js
```diff
@@ -0,0 +1,59 @@
...
+  experiments.forEach((experiment) => {
+    if (!experiment.name || !experiment.variants || experiment.variants.length <= 1) {
+      return; // experiment isn't configured with required settings
+    }
+    // experiment already in cookie?
+    let variant;
+
+    if (ctx.$cookies.get(`eu-ab-${experiment.name}`)) {
+      variant = ctx.$cookies.get(`eu-ab-${experiment.name}`);
+    } else { // else, not in cookie
```
LeoniePeters, line 31: Define `variant` and then just check if undefined (i.e. no cookie yet)
```
let variant = ctx.$cookies.get(`eu-ab-${experiment.name}`);
if (!variant) {
```

src/plugins/ab-testing.js
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+  // check consent
+  const consentCookie = ctx.$cookies.get('klaro');
+  if (!consentCookie?.matomo) {
+    inject('experiments', () => {
+      return false;
+    });
+    return false;
+  }
```
LeoniePeters, line 18: Why inject useless functions? 
```
  if (experiments.length === 0 || !ctx.$cookies.get('klaro')?.matomo) {
    return;
  }
```
In layout/default.vue we can then use `this.$experiments?.()` to first check if the function exists (is injected) and then call it.

LeoniePeters (changes requested):

lbiedinger (approved):

PR #1200, opened 2021-09-01
Prevent feedback spam

mirjamverloop:


LeoniePeters (approved):

src/components/feedback/FeedbackWidget.vue
```diff
@@ -273,6 +287,17 @@
         this.feedbackInputState = false;
       },
 
+      wordCount() {
+        if (!this.feedback) {
+          return false;
+        }
+
+        if (this.feedback?.trim()?.match(/\w+/g)?.length < 5) {
```
rwd, line 295: This does not work as intended:
```js
const req = { body: {} };
validateReq(req) // => true
```

Set a fallback for the chain: `(req.body?.feedback?.trim()?.match(/\w+/g)?.length || 0) < 5`

Same in the server middleware, and ideally we wouldn't be duplicating the validation code but have it in just one place.

src/components/feedback/FeedbackWidget.vue
```diff
@@ -273,6 +287,17 @@
         this.feedbackInputState = false;
       },
 
+      wordCount() {
+        if (!this.feedback) {
```
rwd, line 291: We should be able to remove this and rely solely on the 5 word validation below.

src/server-middleware/api/jira/service-desk.js
```diff
@@ -45,10 +52,14 @@ const jiraOptions = options => ({
...
-    .catch(error => errorHandler(res, error));
+  if (validateReq(req)) {
+    // Docs: https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/#api-rest-servicedeskapi-request-post
+    return axios
+      .create({ baseURL: options.origin })
+      .post(JIRA_SERVICE_DESK_API_PATH, jiraData(options, req), jiraOptions(options))
+      .then(jiraRes => res.sendStatus(jiraRes.status))
+      .catch(error => errorHandler(res, error));
+  } else {
+    return errorHandler(res, { message: 'Invalid feedback.' });
```
rwd, line 63: This will result in a 500 status error but it should be 400.

src/server-middleware/api/jira/service-desk.js
```diff
@@ -45,10 +52,14 @@ const jiraOptions = options => ({
 });
 
 export default (options = {}) => (req, res) => {
-  // Docs: https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/#api-rest-servicedeskapi-request-post
-  return axios
-    .create({ baseURL: options.origin })
-    .post(JIRA_SERVICE_DESK_API_PATH, jiraData(options, req), jiraOptions(options))
-    .then(jiraRes => res.sendStatus(jiraRes.status))
-    .catch(error => errorHandler(res, error));
+  if (validateReq(req)) {
```
rwd, line 55: Let's make the validation function return a promise, so that we can write something like:
```js
  return validateReq(req)
    .then(() => (axios
      .create({ baseURL: options.origin })
      .post(JIRA_SERVICE_DESK_API_PATH, jiraData(options, req), jiraOptions(options))
      .then(jiraRes => res.sendStatus(jiraRes.status))
    ))
    .catch(() => errorHandler(res, error));
```

rwd (changes requested):

rwd (approved):

PR #1224, opened 2021-09-22
Translated metadata labels

lbiedinger:


src/components/item/MetadataOriginLabel.vue
```diff
@@ -0,0 +1,62 @@
...
+
+    props: {
+      translationSource: {
+        type: String,
+        default: null
+      }
+    },
+
+    computed: {
+      hasValuesForLocale() {
```
rwd, line 26: This is not used. Remove it.

src/components/item/MetadataOriginLabel.vue
```diff
@@ -0,0 +1,62 @@
...
+    :class="translationSource"
+    data-qa="translation tooltip"
+  />
+</template>
+
+<script>
+  export default {
+    name: 'MetadataOriginLabel',
+
+    components: {},
```
rwd, line 16: This is empty. Remove it.

src/components/item/MetadataField.vue
```diff
@@ -85,6 +88,10 @@
         type: String,
         default: ''
       },
+      metadataLanguage: {
```
rwd, line 91: Given the component is named MetadataField, this prop could just be named "language".

src/plugins/europeana/record.js
```diff
@@ -301,8 +312,11 @@ export default (context = {}) => {
     * @return {Boolean} true if enriched data will be shown
     */
     localeSpecificFieldValueIsFromEnrichment(field, providerProxy, proxies, predictedUiLang) {
-      if (providerProxy[field][predictedUiLang] ||
-        (!proxies[2][field][predictedUiLang] && providerProxy[field]['en'])) {
+      if (isLangMap(providerProxy[field]) &&
```
rwd, line 315: Let's move this check to its own function to reduce cognitive complexity.

src/plugins/europeana/record.js
```diff
@@ -301,8 +312,11 @@ export default (context = {}) => {
...
+           (providerProxy[field]?.[predictedUiLang] ||
+             (!proxies[2][field]?.[predictedUiLang] && providerProxy[field]?.['en'])
```
rwd, line 317: Should this still be checking `'en'`?

src/plugins/europeana/record.js
```diff
@@ -273,7 +284,7 @@ export default (context = {}) => {
         altTitle: proxyData.dctermsAlternative,
         description: proxyData.dcDescription,
         identifier: edm.about,
-        type: edm.type,
+        type: edm.type, // What's this used for?
```
rwd, line 287: Prefix the comment with "TODO:"

src/pages/item/_.vue
```diff
@@ -340,7 +341,7 @@
 
         const dataSimilarItems = {
           dcSubject: this.getSimilarItemsData(this.coreFields.dcSubject),
-          dcType: this.getSimilarItemsData(this.title),
+          dcType: this.getSimilarItemsData(this.title), // ???
```
rwd, line 344: Prefix the comment with "TODO:"

rwd (changes requested):

src/components/item/MetadataField.vue
lbiedinger: I think metadataLanguage is clearer since it's more specific and for trying to trace this back to it's origin. Since there are still other languages(UI, browser, fallback) that just 'language' could refer to even within this component and it's metadataLanguage elsewhere like in the record plugin and item template.

src/plugins/europeana/record.js
lbiedinger: I've pushed some refactor, hopefully making this a bit clearer.

src/plugins/europeana/record.js
lbiedinger: yes.

This catches the case where:

- The record does not have a recognized/supported EDM language, meaning it will use the 'old' language prioritization
- There is no exact match on the enriched (provider) proxy
- The main (last index) proxy contains no exact language match
- Meaning the fallback logic will kick in.
- The enriched (provider) proxy contains an 'en' value that will be displayed due to fallback being active.

rwd (approved):

PR #1226, opened 2021-09-28
Automated media type count module

LeoniePeters:


src/server-middleware/api/dailyEntries.js
```diff
@@ -23,23 +24,25 @@ export const entriesOfTheDay = (type, config = {}) => {
...
+    key = ITEM_COUNTS_BY_MEDIA_TYPE_CACHE_KEY;
   }
 
   return redisClient.getAsync(key)
     .then(value => JSON.parse(value))
     .then(parsed => {
       redisClient.quitAsync();
-      const offset = offsetOfTheDay(parsed.length);
-      return parsed.slice(offset, offset + subsetSize);
+      const offset = offsetOfTheDay(parsed.length, options);
```
LeoniePeters, line 35: We do not need the 'random' offset for `itemCountsMediaType`. The offset for type `itemCountsMediaType` is always 0 and the size always 5. 
Could we define `subsetSize` and `offset` conditionally if `type === itemCountsMediaType`? Then we don't have to pass the `options` and remove the condition in the return below.

tests/unit/components/generic/InfoCard.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    url: { name: 'fakeURL' },
+    info: '12,000,000',
+    label: 'IMAGE',
+    image: 'ic-image',
+    variant: 'default'
+  }
+});
+
+describe('components/generic/InfoCard', () => {
+  it('shows a cards', () => {
```
LeoniePeters, line 20: 'shows a card'

PR #1227, opened 2021-09-29
chore: optimize Nuxt build config and Bootstrap Vue plugin registration

rwd:
Optimisations to Nuxt config:
1. Enable building source maps in webpack to facilitate debugging in production builds, for instance identifying the source line at which errors occurred, or showing the source module responsible for unused JS identified by Lighthouse performance reports.
2. ~~Enable extractCSS to prevent inlining CSS in HTML, reducing the size of the initial document transfer (by 57 kB/45% on the homepage), but increasing the number of network requests required. Lighthouse performance reports favour extracted CSS.~~ Disabled as it messes up some CSS. See: https://github.com/nuxt/nuxt.js/issues/4219
3. Enable "modern" builds to have Nuxt generate an additional client-side bundle using ECMAscript which will be served to modern browsers detected as supporting it. The modern bundle is 34 kB/5% smaller than the transpiled one for older browsers.
4. Reduce the number of globally registered Bootstrap Vue plugins, instead importing those only used occasionally in the views that need them.

This combination of optimisations has been tested with Lighthouse on the homepage, an item page, the collections hub page and the blog hub page, and result in performance score increases on all of 10-20 points (out of 100).

~~**FIXME:** there is one issue arising out of these changes, with the tooltips indicating metadata origin on the item page. When running in dev mode, these do not render properly at first, but making a change, any change, to `src/components/items/SummaryInfo.vue`, causing the hot reload, gets them rendering properly. This is not however an issue on production builds.~~

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -1,5 +1,10 @@
 /* eslint-disable camelcase */
 
+/*
+** Nuxt config
+** Docs: https://nuxtjs.org/docs/configuration-glossary/
+*/
+
 const APP_SITE_NAME = 'Europeana';
 
 import pkg from './package.json';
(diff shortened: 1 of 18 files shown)


nuxt.config.js
```diff
@@ -227,16 +238,10 @@ export default {
       'LayoutPlugin',
       'LinkPlugin',
       'ListGroupPlugin',
-      'MediaPlugin',
       'ModalPlugin',
       'NavbarPlugin',
-      'NavPlugin',
-      'PaginationNavPlugin',
       'SidebarPlugin',
```
LeoniePeters, line 243: The SidebarPlugin is only used in the PageHeader component so we could also only import it there.

rwd: We could, but it's rendered on every page so it doesn't make a difference to the overall payload.

LeoniePeters (approved):

PR #1228, opened 2021-09-29
refactor: return single metadata fields object from record plugin

rwd:
**What?**

In the record plugin's parsing of Record API responses, stop differentiating and separating fields as core & extra. Instead, return all together as a single metadata object. Then, make the distinction only where it is needed, i.e. in the MetadataBox component.

**Why?**

* Separation of concerns: intermediate views such as the item page should not need to know whether title, description, etc will be in core or extra fields as that distinction is only relevant in the metadata box.
* Moves the field names for core and extra into the metadata box component where it is easier to locate and modify them (as it is the only component doing so would affect).
* Reduction in duplication of logic for entity lookups in the record plugin as it's only done once.
* Reduction in size of the record plugin, which is injected into the context of every request by the `apis` plugin, hence included in the app bundle loaded with any SSR.

### DIFF
--- a/src/pages/item/_.vue
+++ b/src/pages/item/_.vue
@@ -193,7 +193,7 @@
         }, {});
       },
       fieldsAndKeywords() {
-        return { ...this.metadata, ...{ keywords: this.keywords } };
+        return { ...this.metadata, keywords: this.keywords };
       },
       locationData() {
         return this.metadata.dctermsSpatial;
--- a/src/pages/item/_.vue
+++ b/src/pages/item/_.vue
@@ -68,7 +68,7 @@
             class="col-lg-10"
           >
             <MetadataBox
-              v-model="fieldsAndKeywords"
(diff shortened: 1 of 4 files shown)


src/pages/item/_.vue
```diff
@@ -68,8 +68,7 @@
             class="col-lg-10"
           >
             <MetadataBox
-              :all-metadata="allMetaData"
-              :core-metadata="coreFields"
+              v-model="fieldsAndKeywords"
```
lbiedinger, line 71: What's the reason for using v-model here instead of a named prop? Having the metadataBox component refer to `value[name]` seems a bit odd when `value` is in fact multiple fields, so would make more sense to have `values[name]` or `metadata[name]`

rwd: Indeed, thank you. That's been changed to `metadata`.

lbiedinger (approved):

PR #1275, opened 2021-11-02
Client only hit highlighting

lbiedinger:


lbiedinger:
Prefer solution from #1289

PR #1288, opened 2021-11-10
refactor: read thumbnail API URL from context

rwd:
Also:
* Make the thumbnail plugin the SSOT for generating thumbnail URLs
* Make the entity plugin the SSOT for generating image URLs for entities, delegating to the thumbnail plugin where appropriate

Re: EC-5757


src/components/item/ItemPreviewCard.vue
```diff
@@ -112,9 +110,12 @@
       imageUrl() {
         const size = 'w400';
 
-        return this.value.edmPreview ?
-          `${this.value.edmPreview[0]}&size=${size}` :
-          genericThumbnail(this.value.id, { type: this.value.type, size });
+        if (this.value.edmPreview) {
+          const url = new URL(this.value.edmPreview[0]);
+          return this.$apis.thumbnail.url(url.searchParams.get('uri'), url.searchParams.entries());
```
LeoniePeters, line 115: I don't see the search parameters being set in the returned url. I think because `.entries()` returns an iterator, not an object as the `url` function expects. https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/entries
And shouldn't we add the size parameter here too?

src/pages/set/_.vue
```diff
@@ -258,9 +257,14 @@
...
           return null;
         } else {
-          return this.set.items[0].edmPreview ?
-            `${this.set.items[0].edmPreview[0]}&size=w400` :
-            genericThumbnail(this.set.items[0].id, { type: this.set.items[0].type, size: 'w400' });
+          const size = 'w400';
+
+          if (this.set.items[0].edmPreview) {
+            const url = new URL(this.set.items[0].edmPreview[0]);
+            return this.$apis.thumbnail.url(url.searchParams.get('uri'), url.searchParams.entries());
```
LeoniePeters, line 264: Same concern as commented on src/components/item/ItemPreviewCard.vue

src/plugins/apis.js
```diff
@@ -28,7 +30,7 @@ const storeModule = {
       for (const api in state.urls) {
         const apiBaseURL = apiUrlFromRequestHeaders(api, req.headers);
 
-        if (apiBaseURL && this.$apis) {
+        if (apiBaseURL && this.$apis && this.$apis[api].$axios) {
```
LeoniePeters, line 33: Let's use the optional chaining operator `apiBaseURL && this.$apis?.[api].$axios`

LeoniePeters (changes requested):

src/plugins/apis.js
rwd: Done.

src/components/item/ItemPreviewCard.vue
rwd: I have refactored further to make a new `edmPreview` function on the thumbnail plugin, which sets all of type, size and uri, and extended the unit tests to cover.

src/pages/set/_.vue
rwd: Same response as commented on src/components/item/ItemPreviewCard.vue :slightly_smiling_face:

lbiedinger (approved):

PR #1290, opened 2021-11-11
No search translation params on fulltext searches

lbiedinger:


rwd:
Fixed on the API side instead, in https://github.com/europeana/api2/pull/211

PR #1335, opened 2021-12-06
refactor: pre-populate only technical metadata facets

rwd:


rwd:
Redundant given #1473

PR #1337, opened 2021-12-06
selected side filters more visible

lbiedinger:
experimental blue selected dropdown buttons.

### DIFF
--- a/src/assets/scss/dropdown.scss
+++ b/src/assets/scss/dropdown.scss
@@ -9,9 +9,9 @@
   }
 
   .btn-selected {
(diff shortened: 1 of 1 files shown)


src/assets/scss/dropdown.scss
```diff
@@ -9,9 +9,9 @@
   }
 
   .btn-selected {
-    background: $white;
+    background: $blue;
```
LeoniePeters, line 12: This would be applied to the current dropdowns too. Let's add a conditional class for if side filters enabled.

PR #1351, opened 2021-12-20
feat: add Matomo events to content warning modal

LeoniePeters:


src/components/generic/ContentWarningModal.vue
```diff
@@ -41,6 +44,10 @@
       pageSlug: {
         type: String,
         default: null
+      },
+      modalStatic: {
+        type: Boolean,
+        default: false
       }
```
mirjamverloop, line 51: This is a prop, but as far as I can tell, it is never passed to the component. 
This is for example the exhibition index page, similar for exhibition chapters and blog posts.

```
<ContentWarningModal
      v-if="contentWarning"
      :title="contentWarning.name"
      :description="contentWarning.description"
      :page-slug="`exhibition/${identifier}`"
    />
```

Could the prop be removed from this component?

LeoniePeters: The `static` prop is used in the unit test, just as is done for our other uses of the b-modal component. This prop enables us to choose if the modal should be statically loaded and if true enables to do unit testing on the elements in the modal component. Also see: https://bootstrap-vue.org/docs/components/modal#lazy-loading-and-static-modals

mirjamverloop: Okay, thanks!

mirjamverloop (approved):

PR #1360, opened 2022-01-04
test: use Jest for unit tests

rwd:
Re: EC-5594
* Replace use of Vue CLI + Mocha + Chai with Jest for unit tests
* Remove dependency on Vue CLI altogether, just running Jest and eslint directly
* Rewrite all unit tests to use Jest syntax
* Adapt GitHub Actions CI workflow to report unit test code coverage to SonarCloud


rwd:
EC-5594

LeoniePeters (changes requested):

sonar-project.properties
```diff
@@ -1,5 +1,8 @@
+sonar.coverage.exclusions=bin/**,*.js,tests/**
+sonar.dynamicAnalysis=reuseReports
+sonar.exclusions=**/*.stories.js,src/plugins/i18n/locales.js,src/lang/*.js,src/migrations/*.js
```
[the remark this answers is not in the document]
rwd: Indeed, and have done so.

LeoniePeters (approved):

tests/unit/components/item/ItemHero.spec.js
```diff
@@ -83,87 +83,87 @@ const identifier = '/2020601/https___1914_1918_europeana_eu_contributions_10265'
...
-        wrapper.vm.downloadEnabled.should.eq(false);
+        expect(wrapper.vm.downloadEnabled).toBe(false);
       });
     });
-    context('when the rightsstatement is not in copyright and the selected media is not the isShownAt', () => {
+    describe('when the rightsstatement is not in copyright and the selected media is not the isShownAt', () => {
       it('is true', () => {
         const wrapper = factory({ media: [media[0]], identifier });
-        wrapper.vm.downloadEnabled.should.eq(true);
+        expect(wrapper.vm.downloadEnabled);
```
lbiedinger, line 122: This won't fail for me even when downloadEnabled returns false.

It needs to be `expect(wrapper.vm.downloadEnabled).toBe(true);`

lbiedinger: The same applies elsewhere where there is no additional criteria after the `expect(x)` part.

A notable example that's quite common are a lot of the `nock.isDone()` calls. Although it's likely that those still fail if actuall unexpect requests are made which are otherwise not registered in Nock.

lbiedinger (changes requested):

tests/unit/components/item/ItemEmbedCode.spec.js
```diff
@@ -42,50 +42,50 @@ describe('components/item/ItemEmbedCode', () => {
...
 
-        document.execCommand.should.be.calledWith('copy');
+        expect(document.execCommand.calledWith('copy')).toBe(true);
       });
 
       it('shows a notification message', async() => {
         await wrapper.find('#share-embed').trigger('click');
 
-        wrapper.find('[data-qa="share embed copied notice"]').isVisible().should.equal(true);
+        wrapper.find('[data-qa="share embed copied notice"]').isVisible().should;
```
lbiedinger, line 79: This needs to be updated.

lbiedinger (changes requested):

lbiedinger (approved):

PR #1367, opened 2022-01-10
refactor: icon buttons

LeoniePeters:
Re EC-5598


mirjamverloop (approved):

src/assets/scss/icons.scss
```diff
@@ -121,6 +120,18 @@
   content: '\e93c';
 }
 
+.icon-back::before {
```
rwd, line 123: Let's add these icons to the styleguide in docs/style/FontIcons.md

rwd (changes requested):

PR #1387, opened 2022-01-17
feat: content tiers 2-4 in collection search

rwd:
Re EC-5603


src/components/search/SearchInterface.vue
```diff
@@ -299,7 +299,13 @@
         return this.orderedFacets.filter(facet => this.coreFacetNames.includes(facet.name));
       },
       moreFacetNames() {
-        return this.facetNames.filter(facetName => !this.coreFacetNames.includes(facetName));
+        return this.facetNames.filter(facetName => {
```
lbiedinger, line 302: This seems to be causing a regression where the tier-0 filter isn't removed after selecting a thematic-collection while the side-facets are switched off.

rwd: Not from this change, but from a change to `queryUpdatesForFacetChanges` in the search store, but yes.

lbiedinger (approved):

LeoniePeters (approved):

PR #1402, opened 2022-01-28
organisation tier0 label

lbiedinger:


rwd:
Closing in favour of using the incoming SideSwitchFilter component in the context where tier 0 is offered as an on/off filter.

PR #1416, opened 2022-02-08
feat: move type label to align with view buttons

mirjamverloop:


src/components/search/SearchInterface.vue
```diff
@@ -71,10 +71,21 @@
...
-      class="mb-3"
+      class="mb-3 "
+      :class="{ 'd-flex align-items-center': entityHeaderCardsEnabled }"
     >
       <b-col>
-        <p data-qa="total results">
+        <div
+          v-if="entityHeaderCardsEnabled"
+          class="context-label"
+          data-qa="entity label"
```
rwd, line 81: Let's change this to "context label" to be agnostic as to the context, i.e. potentially not to do with entities.

src/components/search/SideFilters.vue
```diff
@@ -127,7 +137,8 @@
         facets: state => state.search.facets,
         resettableFilters: state => state.search.resettableFilters,
         showFiltersSheet: state => state.search.showFiltersSheet,
-        userParams: state => state.search.userParams
+        userParams: state => state.search.userParams,
+        totalResults: state => state.search.totalResults
```
rwd, line 141: Move this up to preserve the alphabetised sorting of the properties.

src/components/search/SearchInterface.vue
```diff
@@ -71,10 +71,21 @@
...
     <b-row
       v-if="hasAnyResults"
-      class="mb-3"
+      class="mb-3 "
+      :class="{ 'd-flex align-items-center': entityHeaderCardsEnabled }"
     >
       <b-col>
-        <p data-qa="total results">
+        <div
+          v-if="entityHeaderCardsEnabled"
```
rwd, line 79: Let's change this condition to `v-if="contextLabel"`, and have the check for whether the feature is enabled go into the parent view instead, i.e. into the entity page.

src/components/search/SideFilters.vue
```diff
@@ -13,7 +13,17 @@
       <b-row
         class="border-bottom border-top d-flex justify-content-between align-items-center"
       >
-        <h2 class="filters-title">
+        <h2
```
rwd, line 16: This is not a section heading so should not use the h2 element.

rwd (changes requested):

rwd (approved):

PR #1425, opened 2022-02-11
feat: permissive CORS on API server middleware

rwd:


rwd:
Not needed for the issue at hand. See #1426 instead.

PR #1434, opened 2022-02-15
feat: announce changed search results

mirjamverloop:


LeoniePeters:
This doesn't work as expected as the enabling of the announcer activates the automated route change announcement which takes preference over setting a new message for the total results. New PR with different approach here: https://github.com/europeana/portal.js/pull/1446

mirjamverloop:
Closing because PR #1446 has a better approach.

PR #1435, opened 2022-02-16
feat: auto-suggest side filters

mirjamverloop:


src/components/search/SideFacetDropdown.vue
```diff
@@ -233,6 +269,17 @@
...
+      criteria() {
+        return this.searchFacet.trim().toLowerCase();
+      },
+
+      searchOptions() {
+        if (this.criteria && this.sortedOptions.length === 0) {
+          return 'There are no tags matching your search criteria';
+        }
+        return '';
       }
```
LeoniePeters, line 311: Why do we need this for the dropdown form description and don't we use the translated value ( `sideFilters.noOptions`) as is done for the actual shown text?

mirjamverloop: To be honest, I'm not sure if we need this description at all. What would be the use case? 

It's a help text ((https://bootstrap-vue.org/docs/components/form-group#comp-ref-b-form-group-props)) and when it will be displayed currently, it will be somewhere next to the search icon and wraps the search field. If we decide to keep it, we should style it.

src/components/search/SideFacetDropdown.vue
```diff
@@ -261,6 +297,17 @@
           profile: 'facets',
           facet: this.name
         };
+      },
+
+      criteria() {
+        return this.searchFacet.trim().toLowerCase();
+      },
+
+      searchOptions() {
```
lbiedinger, line 306: This never returns actual options.

`optionsDescription` would be a better name.

lbiedinger (changes requested):

src/components/search/SideFacetDropdown.vue
LeoniePeters: I'm removing it.

src/components/search/SideFacetDropdown.vue
LeoniePeters: Indeed, naming is incorrect. As the description is not even really used I've removed this prop altogether.

src/components/search/SideFacetDropdown.vue
```diff
@@ -1,84 +1,87 @@
...
-              :name="name"
-              :disabled="filterSelectionDisabled"
-              :data-qa="`${option} ${name} ${RADIO}`"
-              @change="$emit('changed', name, preSelected)"
+          <li
+            v-for="tag in tags"
+            :key="tag"
+            class="list-inline-item mw-100"
+          >
+            <b-form-tag
```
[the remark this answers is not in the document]
LeoniePeters: I don't think using the badge component would work here as we really need the form tag with it's built-in remove event.

lbiedinger (approved):

rwd (approved):

PR #1440, opened 2022-02-21
chore: stop using experimental JSON imports

rwd:
Instead, create a file ./pkg-versions.js with hard-coded values for the package versions of the app itself, @nuxt/core, bootstrap and bootstrap-vue.

The version numbers in this file can be automatically updated using the new script `node bin/pkg-versions.js`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,6 @@ ENV CHROMEDRIVER_SKIP_DOWNLOAD=true \
 
 WORKDIR /app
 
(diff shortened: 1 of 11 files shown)


lbiedinger (approved):

lbiedinger (changes requested):
There's an issue with the e2e tests not starting or being able to connect to the app.

lbiedinger (approved):

PR #1448, opened 2022-02-25
chore: split facet pagination e2e test

rwd:


rwd:
#1449 is a better solution.

PR #1469, opened 2022-03-18
refactor: fetch/pre-fetch facet options as needed

rwd:


rwd:
Incorporated into #1435

PR #1505, opened 2022-04-14
feat: pinning from item pages

lbiedinger:


src/components/item/PinModal.vue
```diff
@@ -0,0 +1,302 @@
...
+        type: Array,
+        required: true
+      },
+      /**
+       * Used for testing, in order to render the modal.
+       */
+      modalStatic: {
+        type: Boolean,
+        default: false
+      }
```
mirjamverloop, line 99: If this is used for testing only, could we remove the prop now?

src/components/item/PinModal.vue
```diff
@@ -0,0 +1,302 @@
...
+    data() {
+      return {
+        fetched: false,
+        selected: null
+      };
+    },
+
+    computed: {
+      infoText() {
+        if (this.selected) {
```
mirjamverloop, line 110: I'm wondering if this conditional could be rewritten. `this.selected` is also been checked in selectedIsPinned and selectedIsFull.

src/pages/item/_.vue
```diff
@@ -243,6 +245,10 @@
           .slice(0, 5)
           .map(entity => entity.about);
       },
+      allEuropeanEntityUris() {
```
mirjamverloop, line 248: Should it be called `allEuropeanaEntityUris`?

src/store/entity.js
```diff
@@ -112,7 +116,8 @@ export default {
     },
     getPins({ state, commit }) {
       return this.$apis.set.get(state.featuredSetId, {
-        profile: 'itemDescriptions'
+        profile: 'itemDescriptions',
+        pageSize: 100 // TODO: should this limit to 24?
```
mirjamverloop, line 120: It makes sense to me to limit this to 24, what's your concern?

src/components/item/PinModal.vue
```diff
@@ -0,0 +1,302 @@
...
+        featuredSetIds: state => state.item.featuredSetIds,
+        allRelatedEntities: state => state.item.allRelatedEntities
+      })
+    },
+
+    methods: {
+      fetchPinningData() {
+        // TODO: Don't fetch all entities if related entities are already present and less than 5. Set all entities to related entities instead for that scenario.
+        return this.$apis.entity.find(this.entities)
+          .then(entities => entities.map(entity => pick(entity, ['id', 'prefLabel', 'isShownBy', 'logo'])))
```
mirjamverloop, line 148: If this is only used for pinning, does it need the logo for example?

src/store/entity.js
lbiedinger: if you accidentally pinned more items by using multiple tabs for example, then you won't be able to see all of the pinned items. So you can't unpin from the end.

tests/unit/components/item/PinModal.spec.js
```diff
@@ -0,0 +1,722 @@
...
+        create: setApiCreateStub,
+        modifyItems: setApiModifyItemsStub
+      },
+      ...apiOverrides
+    }
+  }
+});
+
+describe('components/item/PinModal', () => {
+  afterEach(async() => {
```
lbiedinger, line 154: use: `sinon.resetHistory`

src/components/item/PinModal.vue
```diff
@@ -0,0 +1,302 @@
...
+      :disabled="!fetched"
+      :pressed="selected === entity.id"
+      :data-qa="`pin item to entity choice`"
+      class="btn-collection w-100 text-left d-flex"
+      @click="selectEntity(entity.id)"
+    >
+      <span
+        class="mr-auto"
+      >
+        {{ entity.prefLabel.en }}
```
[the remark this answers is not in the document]
lbiedinger: Yeah, it's like this because it's the name of the best-bets set.
They always get the english pref-label from the entity as a title.

src/components/item/PinModal.vue
lbiedinger: it's used to setup the tests so that when the component is loaded it's got the modal open. The other modal tests also use this same approach.

src/store/entity.js
lbiedinger: I'm going to remove the todo here, because I think there's no harm getting the extra pins _if_ they exist.

src/components/item/PinModal.vue
lbiedinger: well the original design was also including background images and logo/depiction icons.
So that's why it's being retrieved here, but yes it's not being used at the moment.

mirjamverloop (approved):

src/components/item/ItemPinModal.vue
```diff
@@ -0,0 +1,296 @@
...
+      setFactory() {
+        return {
+          id: null,
+          pinned: []
+        };
+      },
+
+      async fetchEntityBestItemsSets() {
+        if (this.fetched) {
+          return;
```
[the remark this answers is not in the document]
lbiedinger: Ignoring the fetched state, or setting fetched to 'false' when the modal is closed wouldn't necessarily solve the underlying issue, but perhaps make it less likely to occur.

The Unpin method should catch that error from the API though.

mirjamverloop (approved):

PR #1511, opened 2022-04-22
feat: quick search component with links to themes

LeoniePeters:


lbiedinger:
prefer https://github.com/europeana/portal.js/pull/1516

PR #1524, opened 2022-05-06
fix: only render Masonry item card groups client-side

rwd:
When using the v3 Thumbnail API, and some item previews have no thumbnail, once the Masonry view initiates client-side it results in duplicate requests for all of the images without a thumbnail, due to vue-masonry's use of imagesloaded: https://github.com/desandro/imagesloaded/blob/v4.1.4/imagesloaded.js#L281

This resolves that by only loading the item card previews client-side if the view is a Masonry one... but at what cost?


rwd:
Not convinced this is a good idea, preferring #1525

PR #1527, opened 2022-05-09
refactor: SearchForm, SearchQueryOptions and PageHeader components

LeoniePeters:
Refactor to improve and clean up the search form and related components. This includes:

- Move back button from PageHeader to SearchForm component
- Move clear and filter buttons outside b-form. The HTML form should only include the search input
- Logic to close the search options dropdown when clicking outside the SearchForm (or its children, i.e. SearchQueryOptions)
- logic to handle keyboard navigation with arrow keys to select a search option
- Clean up deprecated logic for tracking matomo event on search option selection with enter key. This now happens outside the form element and is tracked on click anyway
- Clean up deprecated logic for manually setting focus on search options
- Add PageHeader and SearchQueryOptions components to the style guide

### DIFF
--- a/tests/unit/components/search/SearchForm.spec.js
+++ b/tests/unit/components/search/SearchForm.spec.js
@@ -1,12 +1,15 @@
-import { createLocalVue, shallowMount } from '@vue/test-utils';
+import { createLocalVue, shallowMount, mount } from '@vue/test-utils';
 import SearchForm from '@/components/search/SearchForm.vue';
+import BootstrapVue from 'bootstrap-vue';
 import Vuex from 'vuex';
 import sinon from 'sinon';
 
 const localVue = createLocalVue();
 localVue.use(Vuex);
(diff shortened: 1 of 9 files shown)


tests/unit/components/search/SearchForm.spec.js
```diff
@@ -299,4 +303,120 @@ describe('components/search/SearchForm', () => {
...
+    const wrapper = factory();
+
+    await wrapper.setData({ showSearchOptions: true });
+
+    wrapper.vm.handleKeyDown(escapeEvent);
+
+    expect(wrapper.vm.showSearchOptions).toBe(false);
+  });
+
+  // TODO Fails but why?
```
lbiedinger, line 396: TODO:

LeoniePeters: Instead I just fixed the test.

lbiedinger (approved):

PR #1532, opened 2022-05-11
chore: allow override of iiif presentation endpoint for media player

lbiedinger:


src/pages/media/index.vue
```diff
@@ -8,6 +8,7 @@
 
 <script>
   import { requiresDashJS } from '../../plugins/media';
+  import { apiConfig } from '@/plugins/europeana/utils.js';
```
LeoniePeters, line 11: Is this needed? It's not being used.

LeoniePeters (changes requested):

PR #1540, opened 2022-05-19
feat: search bar on new homepage

LeoniePeters:


src/components/search/SearchForm.vue
```diff
@@ -76,7 +82,14 @@
...
     components: {
       SearchQueryOptions,
-      FilterToggleButton
+      FilterToggleButton: () => import('./FilterToggleButton')
+    },
+
+    props: {
+      absoluteTopPositioned: {
+        type: Boolean,
+        default: false
```
lbiedinger, line 91: wouldn't it make more sense to have the 'default' be true for this?

src/components/home/HomePage.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <div
```
lbiedinger, line 2: While we won't be using it elsewhere, I think this could already be it's own component.
Not for re-usability, but to keep this homepage template from getting too big down the line.

src/components/search/SearchForm.vue
LeoniePeters: My idea was that the absolute positioned variant would definitely only be used once. If reused it would just be placed over the other. The relative positioned search bar could theoretically be reused.

lbiedinger: Righ, that's fine. I was thinking that the one in the top-nav would be used more frequently since it's on every page, therefore should be the default.

lbiedinger (approved):

PR #1564, opened 2022-06-09
fix: item preview group drag offset

rwd:


src/components/item/ItemPreviewCardGroup.vue
```diff
@@ -151,13 +153,17 @@
 
     data() {
       return {
-        orderedItems: this.items
+        draggableCards: []
```
LeoniePeters, line 156: Lets rename this as the cards are not always draggable (on all other views besides sets and the related collections card). For example `orderedCards`

LeoniePeters (approved):

PR #1593, opened 2022-07-04
feat: call to action illustration and styles

LeoniePeters:


src/migrations/0067-illustration-content-type-primaryCallToAction.cjs
```diff
@@ -0,0 +1,55 @@
...
+    .disabled(false)
+    .omitted(false)
+    .linkType('Asset');
+
+  illustration.changeFieldControl('image', 'builtin', 'assetLinkEditor', {});
+
+  const primaryCallToAction = migration.editContentType('primaryCallToAction');
+
+  primaryCallToAction
+    .createField('illustration')
```
rwd, line 40: Change the ID of this new field to `image` instead, but keep the name "Illustration".

src/components/generic/CallToActionBanner.vue
```diff
@@ -1,9 +1,14 @@
...
-    <div class="cta-illustration align-self-stretch flex-lg-fill">
-      <div class="geometry" />
-    </div>
-    <div class="cta-content align-self-stretch flex-lg-fill">
+  <div
+    class="home-cta d-flex flex-md-row flex-column"
+    :class="variant"
+  >
+    <div
+      v-if="illustration"
```
rwd, line 7: Could we use `b-img-lazy` here instead?

src/components/generic/CallToActionBanner.vue
```diff
@@ -1,9 +1,14 @@
...
+      class="cta-illustration align-self-stretch"
+      :style="`background-image: url(${illustration.image.url})`"
```
rwd, line 9: This needs to go through our image optimisation helper function, in case images other than SVGs are uploaded and need to be scaled down to a reasonable maximum size.

rwd (changes requested):

PR #1608, opened 2022-07-14
refactor: include curated entities in homepage GraphQL

rwd:


rwd:
Nope. Performance measured by Lighthouse is better without this change.

PR #1620, opened 2022-07-28
feat: set-driven gallery listing page

lbiedinger:
Introduces a feature toggle:
`ENABLE_SET_GALLERIES`
which if set, will load 'galleries' using published user sets. Ensure to use a `EUROPEANA_SET_API_URL` where there are published sets present. For example at the time of PR creation these only exist on `https://api-acc.eanadev.org/set`


src/pages/galleries/index.vue
```diff
@@ -48,41 +67,86 @@
...
+        return this.$features.setGalleries;
+      },
+      page() {
+        return Number(this.$route.query.page || 1);
+      }
+    },
+    watch: {
+      '$route.query.page': '$fetch'
+    },
+    watchQuery: ['page'], // TODO: remove when using set galleries
```
rwd, line 100: Is this still actually doing anything? I think not with the switch to using the fetch hook.

src/pages/galleries/index.vue
```diff
@@ -11,15 +11,33 @@
           deck
           data-qa="gallery foyer"
         >
```
rwd, line 13: Add checks on $fetchState.pending and $fetchState.error, to render the loading spinner and alert message component respectively.

src/pages/galleries/index.vue
```diff
@@ -11,15 +11,33 @@
...
-            :url="{ name: 'galleries-all', params: { pathMatch: gallery.identifier } }"
-            :image-url="gallery.hasPartCollection.items[0] && imageUrl(gallery.hasPartCollection.items[0])"
-            :texts="[gallery.description]"
-            :show-subtitle="false"
-          />
+          <template
+            v-if="setGalleriesEnabled"
+          >
+            <ContentCard
+
```
rwd, line 18: Remove this blank line.

src/pages/galleries/index.vue
```diff
@@ -11,15 +11,33 @@
...
-            v-for="gallery in galleries"
-            :key="gallery.identifier"
-            :title="gallery.name"
-            :url="{ name: 'galleries-all', params: { pathMatch: gallery.identifier } }"
-            :image-url="gallery.hasPartCollection.items[0] && imageUrl(gallery.hasPartCollection.items[0])"
-            :texts="[gallery.description]"
-            :show-subtitle="false"
-          />
+          <template
+            v-if="setGalleriesEnabled"
```
rwd, line 15: Can the v-if and v-else be moved onto the ContentCard component element, making the template elements redundant?

src/pages/set/_.vue
```diff
@@ -174,6 +176,15 @@
       SetRecommendations: () => import('@/components/set/SetRecommendations')
     },
 
+    async beforeRouteEnter(to, from, next) {
```
rwd, line 179: This should be a middleware instead so that it happens earlier and can stop the rendering of the page. As it is here, there is a brief 404 error page visible, then it redirects away.

rwd (changes requested):

src/pages/galleries/index.vue
lbiedinger: no, left over from a previous iteration when it was needed.

src/pages/galleries/index.vue
lbiedinger: it goes a bit crazy when you do it like that since the content cards are looped through with v-for.
You get  the else output 20 times. So this is a bit of a workaround for that.

src/pages/galleries/index.vue
lbiedinger: yes, good thinking. 
The loading spinner especially makes paginating much more user friendly.

rwd (changes requested):

mirjamverloop (approved):

mirjamverloop (approved):

PR #1638, opened 2022-08-16
New design for organisations hub page

mirjamverloop:
* Display of both native name and English name
* Updated styling of table

### DIFF
--- a/src/cachers/collections/organisations.js
+++ b/src/cachers/collections/organisations.js
@@ -24,7 +24,7 @@ const data = async(config = {}) => {
 
   const recordCounts = await getRecordCounts();
 
-  organisationData.map(
+  return organisationData.map(
     organisation => {
       const organisationId = organisation.id;
       const organisationWithCount = recordCounts.find(facet => facet.label === organisationId);
@@ -32,8 +32,6 @@ const data = async(config = {}) => {
       organisation.recordCount = recordCount;
       return organisation;
     });
-
-  return organisationData;
 };
 
 export {
--- a/src/cachers/collections/organisations.js
+++ b/src/cachers/collections/organisations.js
@@ -1,15 +1,46 @@
 import baseData from './index.js';
 import organizationsMixin from '../../mixins/europeana/entities/organizations.js';
 import { langMapValueForLocale } from '../../plugins/europeana/utils.js';
+import { createEuropeanaApiClient } from '../utils.js';
 
-const PICK = ['slug', 'nativeLabel', 'nonNativeEnglishLabel'];
+const PICK = ['slug', 'recordCount', 'nativeLabel', 'nonNativeEnglishLabel'];
 const INTERNATIONALISE = (entities) => entities.map((entity) => ({
   ...entity,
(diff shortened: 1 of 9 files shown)


src/cachers/collections/organisations.js
```diff
@@ -1,18 +1,49 @@
...
+  const baseOrgData = await baseData({ type: 'organization' }, config);
+
+  axiosClient = createEuropeanaApiClient(config.europeana?.apis?.record);
+
+  const recordCounts = await getRecordCounts();
+
+  Object.keys(baseOrgData).forEach(
+    key => {
+      const organisationId = baseOrgData[key].id;
+      const organisation = recordCounts.filter(facet => facet.label === organisationId)[0];
```
rwd, line 37: Use `.find` instead of `.filter()[0]`

src/cachers/collections/organisations.js
```diff
@@ -1,18 +1,49 @@
...
+      const recordCount = organisation?.count;
```
rwd, line 38: Perhaps default this to 0 if there is no count?
```
const recordCount = organisation?.count || 0;
```

src/cachers/collections/organisations.js
```diff
@@ -1,18 +1,49 @@
...
+}
+
+const data = async(config = {}) => {
+  const baseOrgData = await baseData({ type: 'organization' }, config);
+
+  axiosClient = createEuropeanaApiClient(config.europeana?.apis?.record);
+
+  const recordCounts = await getRecordCounts();
+
+  Object.keys(baseOrgData).forEach(
```
rwd, line 34: baseOrgData should be an array, so Instead of this, just use:
```
for (const organisation of baseOrgData) {
```

src/cachers/collections/organisations.js
```diff
@@ -1,18 +1,49 @@
 import baseData from './index.js';
 import organizationsMixin from '../../mixins/europeana/entities/organizations.js';
+import { langMapValueForLocale } from '../../plugins/europeana/utils.js';
+import { createEuropeanaApiClient } from '../utils.js';
 
-const PICK = ['slug', 'prefLabel'];
+const PICK = ['slug', 'recordCount', 'nativeLabel', 'nonNativeEnglishLabel'];
 const INTERNATIONALISE = (entities) => entities.map((entity) => ({
   ...entity,
-  prefLabel: organizationsMixin.methods.organizationEntityNativeName(entity)
```
rwd, line 7: I'd be more inclined to (once again) cache the full prefLabel here, and have the front-end responsible for picking out what it wants with the mixin methods.

That may in fact make the whole `INTERNATIONALISE` method redundant here.

rwd (changes requested):

src/cachers/collections/organisations.js
LeoniePeters: You're right, it's an array. I've used `.map` instead.

src/cachers/collections/organisations.js
LeoniePeters: I was doubting where to handle this too. I have refactored it to handle the different language labels in the EntityTable component.

lbiedinger (approved):

PR #1642, opened 2022-08-17
fix: detect orgs with > 2 prefLabel properties

rwd:


lbiedinger (approved):

src/components/item/MetadataField.vue
```diff
@@ -69,6 +69,7 @@
 </template>
 
 <script>
+  import europeanaEntitiesOrganizationsMixin from '@/mixins/europeana/entities/organizations';
```
LeoniePeters, line 72: Remove the mixin as it's unused

LeoniePeters (changes requested):

PR #1646, opened 2022-08-18
feat: update homepage title text with highlight

lbiedinger:


src/lang/en.js
```diff
@@ -762,7 +762,8 @@ export default {
     "subHeadline": "Search, save and share art, books, films and music from thousands of cultural institutions",
     "themesCTA": "See all themes",
     "themesTitle": "Explore by theme",
-    "title": "Discover Europe’s cultural heritage"
+    "title": "Discover Europe’s {0} cultural heritage",
```
rwd, line 765: Let's use named slots instead of numbered, as we do with e.g. providedBy

src/assets/scss/fonts.scss
```diff
@@ -6,19 +6,30 @@
   font-display: block;
 }
 
-// Open Sans v28
+// Open Sans v34
```
rwd, line 9: Let's add a comment here explaining how to update the fonts.

rwd (changes requested):

rwd (approved):

PR #1656, opened 2022-08-23
feat: update EU multilingual logo and add to home hero

LeoniePeters:


src/components/generic/EULogo.vue
```diff
@@ -0,0 +1,61 @@
...
+</template>
+<script>
+  export default {
+    name: 'EULogo',
+
+    computed: {
+      localisedEULogo() {
+        let logo;
+        try {
+          logo = require(`@/assets/img/eu-funded/${this.$i18n.locale}-Funded by the EU_NEG.svg`);
```
rwd, line 18: We do not use `require` anywhere else in our .vue files.

Can you rewrite this to use import, or just hard-code the exception given that the component is specifically for this one image?

src/components/generic/EULogo.vue
```diff
@@ -0,0 +1,61 @@
...
+    width="304"
+    height="64"
+    :src="localisedEULogo"
+    :alt="$t('footer.imageDescription')"
+    class="eu-logo"
+  >
+</template>
+<script>
+  export default {
+    name: 'EULogo',
```
rwd, line 12: This is not at all generic, so move it to a different directory.

rwd (changes requested):

src/components/generic/EULogo.vue
LeoniePeters: I followed the Nuxt docs on using dynamic image assets: https://nuxtjs.org/docs/directory-structure/assets#images
Let's keep as is.

PR #1658, opened 2022-08-24
fix: default empty lang attribute to empty string

rwd:


rwd:
No. The original code is fine. WAVE complains about `lang=""` but that's valid.

PR #1659, opened 2022-08-24
refactor: introduce NPM workspaces

rwd:
Re EC-5994

* Introduce Lerna for version management
* Introduce NPM workspaces in the top-level directory `packages/`, beneath which:
  * `portal/` contains the Nuxt application code
  * `styleguide/` contains the vue-styleguidist setup
  * `contentful/` contains the Contentful migrations
  * [in future, additional packages may be introduced, e.g. for the custom Nuxt modules, or the Contentful apps, or the Express API]
* The top-level package.json only includes dependencies required by global scripts, e.g. eslint, stylelint, lerna
* Vue is locked to a specific version of 2.6 due to 2.7 introducing noisy deprecation warnings
* To install all dependencies, run `npm ci`. To install just one workspace's dependencies, run e.g. `npm ci -w packages/portal`
* The Dockerfile for the portal package only installs the dependencies it needs, which improves build time due to not installing vue-styleguidist dependencies. It also removes the base stage which is redundant given changes to CI workflow.
  * In future, it is expected that the styleguide package will get its own Dockerfile
* The CI workflow runs linting and unit tests as separate jobs, and without using Docker images, and also runs them on draft pull requests, all of which is made more practical, and speedier, due to isolation of dependencies
  * Example run on this PR while still in draft: https://github.com/europeana/portal.js/runs/8153120727?check_suite_focus=true  * NPM publish and jsDelivr warmup removed from CI workflow
* jsDelivr asset loading removed from Nuxt config, so that resources are always loaded directly from the app and/or a proxy such as Cloudflare


rwd:
Re EC-5994

lbiedinger (approved):

PR #1660, opened 2022-08-24
use row-clicked to make entire row "links"

lbiedinger:


lbiedinger:
Base PR has been updated to only make the link in the cell clickable as opposed to the approach here, which has potential accessibility issues due to the link being removed.

PR #1672, opened 2022-08-31
feat: error message for item not found

LeoniePeters:


src/lang/en.js
```diff
@@ -211,6 +211,13 @@ export default {
     }
   },
   "error": "Error",
+  "errorMessage": {
+    "item": {
```
rwd, line 215: Change this to `itemNotFound`

src/pages/item/_.vue
```diff
@@ -125,7 +124,7 @@
 </template>
 
 <script>
-  import isEmpty from 'lodash/isEmpty';
+  import isEmpty from 'lodash/isEmpty.js';
```
rwd, line 127: What does this achieve?

rwd (changes requested):

src/pages/item/_.vue
LeoniePeters: At some point test coverage failed locally and I was under the assumption this fixed it. It does not. I see now the svg transformer that I had added to the jest config is the problem. Above change is reverted and transformer fixed.

rwd (approved):

PR #1680, opened 2022-09-07
refactor: move search results count to context

rwd:
* Removes the item count from the top of the side filters, causing the h2 "Filter items" to become persistent there
* Renames the translation keys for the various search results context labels
* Switches the search results context labels translations from using numbered slots to using named slots
* Refactors the search results context component to rely on the i18n named slots to render or not each potential part of the context, instead of having multiple template conditions and duplication within each
* Switches the search results context component to using props instead of the store
* Introduces the number of results into all search results context labels


packages/portal/src/components/search/SideFilters.vue
```diff
@@ -15,21 +15,7 @@
...
-            class="filters-title"
-            data-qa="total results"
-          >
-            {{ $tc('items.itemCount', totalResultsLocalised, { count: totalResultsLocalised }) }}
-            <div
-              class="visually-hidden"
-              role="status"
-            >
-              {{ $t('searchHasLoaded', [totalResultsLocalised]) }}
-            </div>
```
LeoniePeters, line 29: We need this status message for accessibility.

LeoniePeters (approved):

packages/portal/src/components/search/SideFilters.vue
rwd: Indeed, my apologies, thank you for restoring it (elsewhere).

PR #1706, opened 2022-09-21
a11y: move feedback widget to footer

rwd:


packages/portal/src/components/feedback/FeedbackWidget.vue
```diff
@@ -380,7 +381,7 @@
       position: fixed;
       right: 1rem;
       bottom: 1rem;
-      z-index: 1000;
+      z-index: 10000;
```
lbiedinger, line 384: this seems excessive. For what scenario was this needed? I think we should be able to move whatever other element was in front of this to have a lower z-index instead.

lbiedinger (changes requested):

packages/portal/src/components/feedback/FeedbackWidget.vue
rwd: Indeed. It was unintentional. Reverted.

PR #1735, opened 2022-10-17
feat: discover related stuff

rwd:


packages/portal/src/pages/exhibitions/_exhibition/_chapter.vue
```diff
@@ -47,24 +47,22 @@
           </article>
         </b-col>
       </b-row>
-      <client-only>
```
LeoniePeters, line 50: The chapters list is client-side only for performance. Shouldn't we keep it?

rwd: You are right.

I have instead moved all three of the auxiliary blocks (chapters, related stories, related collections) client-only, resulting in a 3-point performance score improvement in Lighthouse on the exhibition chapter page.

On the credits page, both auxiliary blocks (chapters, related collections) have been made client-only.

On the landing page, two of the three blocks (related stories, related collections) have been made client-only, but chapters not as that is more core content for the landing page, and more likely to be visible on initial render.

PR #1771, opened 2022-11-18
refactor: move like/unlike logic to button component

rwd:


LeoniePeters (approved):

packages/portal/src/components/item/ItemLikeButton.vue
```diff
@@ -92,19 +95,26 @@
...
         try {
-          await this.$store.dispatch('set/like', this.identifier);
-          this.$matomo && this.$matomo.trackEvent('Item_like', 'Click like item button', this.identifier);
-        } catch (e) {
-          // TODO: remove when 100 item like limit is removed
-          if (e.message === '100 likes') {
+          // TODO: temporary prevention of addition of > 100 items; remove when no longer needed
+          await this.$store.dispatch('set/fetchLikes');
+
+          if (this.likedItems.length >= 100) {
```
lbiedinger, line 101: fetchLIkes can return `null` when there is no likes set.

PR #1802, opened 2022-12-19
feat: add pagination to user sets in account

rwd:
* Introduce pagination in the galleries tabs of the my account page (UserSets component). 20 items at a time unless the create set button is shown, in which case 19 items at a time.
* Add a new tab for published galleries to the my account page
* Stop using the store for retrieval of the logged-in user's sets, and do so in the respective components instead
* Only load the sets when needed for the tab displayed, not all of them in advance
* Update the PaginationNav component to also preserve the hash from the route
* Update API plugins and utils so that the axios params serializer always repeats for arrays, e.g. `qf`

### DIFF
--- a/packages/portal/src/components/set/AddItemToSetModal.vue
+++ b/packages/portal/src/components/set/AddItemToSetModal.vue
@@ -6,7 +6,7 @@
     hide-header-close
     :static="modalStatic"
     @show="fetchCollections"
-    @hide="hideModal()"
+    @hide="hideModal"
   >
     <b-button
       variant="primary"
@@ -18,6 +18,7 @@
     </b-button>
     <div class="collections">
       <AddItemToSetButton
+        v-if="collections"
         v-for="(collection, index) in collections"
         :key="index"
         :set="collection"
@@ -98,6 +99,7 @@
     methods: {
       async fetchCollections() {
         const creator = this.$auth.user?.sub;
+        // TODO: pagination and/or search within one's collections
         const searchParams = {
           query: `creator:${creator}`,
           profile: 'standard',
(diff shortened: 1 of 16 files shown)


packages/portal/src/components/set/AddItemToSetModal.vue
```diff
@@ -99,11 +99,22 @@
...
+          profile: 'standard',
+          pageSize: 100,
+          page: 0,
+          qf: [
+            'type:Collection'
+          ]
+        };
+
+        const searchResponse = await this.$apis.set.search(searchParams, { withMinimalItemPreviews: true });
+        this.collections = searchResponse.data.items;
```
LeoniePeters, line 116: I'm thinking we want to add an empty array fallback here for when there are no sets returned. At the moment I don't see it might break the code if `collections` will become undefined, but it will be more as expected as the default data prop is an empty array too.

packages/portal/src/components/set/AddItemToSetModal.vue
```diff
@@ -16,7 +16,10 @@
     >
       {{ $t('set.actions.createNew') }}
     </b-button>
-    <div class="collections">
+    <div
+      v-if="collections"
```
LeoniePeters, line 20: This could still be rendered when collections is an empty array, which is truthy.

packages/portal/src/components/set/AddItemToSetModal.vue
rwd: OK. Reverted.

packages/portal/src/components/set/AddItemToSetModal.vue
rwd: OK. Reverted.

PR #1805, opened 2022-12-22
feat: query based related galleries on search page

LeoniePeters:


rwd (changes requested):

packages/portal/src/assets/scss/cards.scss
```diff
@@ -480,10 +486,47 @@
...
+
+.card-group-list {
+  .header-card,
+  .related-galleries-card,
+  .related-collections-card {
+    margin-left: 0;
+    margin-right: 0;
+  }
+
+  .related-galleries-card {
```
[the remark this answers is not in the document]
LeoniePeters: I've moved it into the component together with some of the other related-galleries-card styles from masonry.scss

packages/portal/src/components/related/RelatedGalleries.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+  <b-card
+    v-if="relatedGalleries.length > 0"
```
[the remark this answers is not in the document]
LeoniePeters: As we do with the other related components. Done.

packages/portal/src/components/related/RelatedGalleries.vue
```diff
@@ -0,0 +1,117 @@
...
+      ContentCard
+    },
+
+    props: {
+      /**
+       * search query
+      */
+      query: {
+        type: String,
+        // Default to an empty string (rather than `null`), to ensure that only
```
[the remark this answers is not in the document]
LeoniePeters: Incorrectly copied from `RelatedEditorial` component. I've cleaned up the comment and set default to `null`

packages/portal/src/pages/search/index.vue
```diff
@@ -5,16 +5,38 @@
...
       >
-        <client-only>
-          <RelatedSection
-            :query="searchQuery"
-            :overrides="relatedCollections"
-            data-qa="related section"
-            @fetched="handleRelatedSectionFetched"
-          />
-        </client-only>
+        <aside
```
rwd, line 10: Can we instead make `aside` the root element inside the RelatedSection and RelatedGalleries components?

That should prevent empty aside elements being rendered when there is nothing related, e.g. no related galleries to "wolf".

packages/portal/src/pages/search/index.vue
```diff
@@ -5,16 +5,38 @@
...
+          v-masonry-tile
```
rwd, line 11: This is applied even when using the list view. Is that a problem?

rwd (changes requested):

packages/portal/src/pages/search/index.vue
LeoniePeters: I have moved the `aside` to the RelatedSection and RelatedGalleries components but without adding the conditional render. We either need to
- keep the aside placeholder for the related cards to reserve the space in the masonry, also when there are no related results, otherwise it will still cause a jump in the layout to fill the empty space
- Wait for everything to be ready and then show the masonry layout. We'd need to do state management to check if all components in the masonry are ready. To me this seems a bit overtly complex, harder to maintain, longer time before anything is visible and error prone.
- accept a jumpy layout, maybe tweaking the transition a bit to smudge the asynchronous loading times.

packages/portal/src/pages/search/index.vue
LeoniePeters: I changed it so the directive is active only when not in list view (also for other masonry tiles).

PR #1821, opened 2023-01-19
feat: theme page

LeoniePeters:
- Adds graphQL query for Contentful Theme page content type
- Adds `entityUri` field to Contentful Theme page content type migration
- Extends responsive (big screen/gridless) styles for buttons and cards
- Moves `dailyOffset` logic to plugins to be reusable for daily rotated curated items
- Adds new Vue themes pages
- Adds `EntityCardGroup` component
- Adds `SetCardGroup` component
- Adds `entityLinks` mixin (used by `EntityBadges` and `EntityCardGroup` components)
- Extends `RelatedEditorial` component to use card group without card wrapper


packages/portal/src/pages/themes/_.vue
```diff
@@ -0,0 +1,175 @@
...
+      EntityBadges: () => import('@/components/entity/EntityBadges'),
+      EntityCardGroup: () => import('@/components/entity/EntityCardGroup'),
+      ItemPreviewCardGroup: () => import('@/components/item/ItemPreviewCardGroup'),
+      RelatedEditorial: () => import('@/components/related/RelatedEditorial'),
+      SetCardGroup: () => import('@/components/set/SetCardGroup'),
+      SmartLink: () => import('@/components/generic/SmartLink')
+    },
+
+    mixins: [pageMetaMixin],
+
```
lbiedinger, line 88: add a data section here with some defaults:

```
    data() {
      return {
        name: '',
        nameEN: null,
        entityUri: null,
        description: null,
        primaryImageOfPage: null,
        hasPartCollection: null
      };
    },
```

packages/portal/src/modules/contentful/queries/theme-page.graphql
```diff
@@ -0,0 +1,82 @@
...
+  themePage: themePageCollection(
+    preview: $preview
+    locale: $locale
+    limit: 1
+    where: {identifier: $identifier}
+  ) {
+    items {
+      identifier
+      name
+      nameEN: name(locale: "en-GB")
```
lbiedinger, line 11: what's this used for?

packages/portal/src/pages/themes/_.vue
```diff
@@ -0,0 +1,175 @@
...
+    async asyncData({ params, query, error, app }) {
+      const variables = {
+        locale: app.i18n.isoLocale(),
+        identifier: params.pathMatch,
+        preview: query.mode === 'preview'
+      };
+
+      try {
+        const response = await app.$contentful.query('themePage', variables);
+        const theme = response.data.data.themePage.items[0];
```
lbiedinger, line 98: check that an actual theme was returned, as in `theme` isn't null and/or has an `identifier`, otherwise we should show a 404 page.

packages/portal/src/pages/themes/_.vue
LeoniePeters: Added .

packages/portal/src/modules/contentful/queries/theme-page.graphql
LeoniePeters: Nothing. That's removed.

packages/portal/src/pages/themes/_.vue
LeoniePeters: Added a check.

rwd:
See #1831

PR #1835, opened 2023-02-02
feat: preserve relatedLinks on theme migration

lbiedinger:


rwd (changes requested):
If we decide we want to keep the link, then this is not the way to do it. Instead, we would just not change `relatedLink`, i.e. remove it from the fields being transformed.

PR #1853, opened 2023-02-16
fix: prevent undefined Europeana runtime config settings

rwd:
* Move all logic for establishing which API URL to use into the `apis` plugin (reducing the logic in nuxt.config.js) and spec it
* Make API env var naming consistent and thus derivable
* Declare the APIs requiring authentication, to receive API key lookup from env var

Env var name changes:
* `EUROPEANA_RECORD_API_FULLTEXT_URL` renamed to `EUROPEANA_FULLTEXT_API_URL`
* `EUROPEANA_MEDIA_IIIF_PRESENTATION_API_URL` renamed to `EUROPEANA_IIIF_PRESENTATION_API_URL`
* `EUROPEANA_MEDIA_PROXY_URL` renamed to `EUROPEANA_MEDIA_PROXY_API_URL`


rwd:
Re EC-6241

lbiedinger (approved):

PR #1855, opened 2023-02-17
feat: notification banner for maintenance

LeoniePeters:


packages/portal/.env.example
```diff
@@ -179,3 +179,6 @@ OAUTH_CLIENT="YOUR_CLIENT"
 
 # Username of the official Europeana account for published set-driven galleries.
 # APP_GALLERIES_EUROPEANA_ACCOUNT=europeana
+
+# Name of maintenance to announce in a notification banner
+# APP_FEATURE_NOTIFICATION=setsKeycloakMaintenance
```
lbiedinger, line 184: `APP_NOTIFICATION_BANNER`

lbiedinger: actually let's add more context here:
```
# Name of simple i18n key for a global notification banner,
# ensure the key has a translation at least in /src/lang/en.js
# at notificationBanner.text.[KEY]
# APP_NOTIFICATION_BANNER=key
```

LeoniePeters: Done.

lbiedinger (approved):

PR #1867, opened 2023-03-02
fix: fallback to first prefLabel where no english one is available

lbiedinger:


packages/portal/src/components/item/ItemPinModal.vue
```diff
@@ -20,7 +20,7 @@
         class="mr-auto"
       >
         <!-- TODO: localise here, even if not on the set itself -->
```
rwd, line 22: Given that there is a TODO note here to handle l10n, let's do so now.

* Instead of more localisation logic in the template, import and use `langMapValueForLocale` from `@/plugins/europeana/utils.js`
* As well as the value to display, use the language code returned by that to set the `lang` property on HTML elements

rwd (changes requested):

rwd:
See #1873

PR #1884, opened 2023-03-17
remove pagination in LanguageSelector

lbiedinger:
To prevent no results on localised index pages where results are localised.

### DIFF
--- a/packages/portal/src/components/generic/LanguageSelector.vue
+++ b/packages/portal/src/components/generic/LanguageSelector.vue
@@ -10,7 +10,7 @@
     <b-dropdown-item
       v-for="locale in availableLocales"
       :key="locale.code"
-      :href="handledLocalePath(locale.code)"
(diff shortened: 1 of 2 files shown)


packages/portal/src/components/generic/LanguageSelector.vue
```diff
@@ -32,6 +32,20 @@
...
+        return ['galleries', 'stories'].some((routeNameBase) =>  $nuxt.$route.name.startsWith(routeNameBase));
+      }
+    },
+    methods: {
+      handledLocalePath(code) {
+        const path = this.switchLocalePath(code);
+        if (this.removePaginationAtLanguageSwitch) {
+          return path.replace(/(\?|&)page=\d+/, '');
+        }
+        return this.switchLocalePath(code);
```
lbiedinger, line 47: no need to call switchLocalePath again, just return path

PR #1910, opened 2023-04-19
feat: lookup WR edm:type from URL

rwd:
**NB: don't merge. Refactor for #1727.**

If no ebucore:hasMimeType stored.

e.g. item `/09102/_GNM_693983` where video and audio WRs do not have technical metadata, but their URLs include filename extensions.


rwd:
No, let's not do this. It should come from tech metadata.

PR #1929, opened 2023-05-11
feat: browse page styles update

LeoniePeters:


packages/portal/src/components/browse/BrowseSections.vue
```diff
@@ -50,6 +50,13 @@
         :text="section.text"
         :url="section.url"
       />
+      <CallToActionBanner
+        v-else-if="contentType(section, 'PrimaryCallToAction') && section.image"
```
lbiedinger, line 54: why is the image required here?

lbiedinger: for the primary CTA component to be used otherwsie. 

Styling for that has been updated.

PR #1930, opened 2023-05-12
refactor: add mirador workspace

rwd:
Re EC-6310

Instead of loading Mirador from a CDN, add a new mirador workspace package which imports, and re-exports, the Mirador ESM.

This new package also contains the Europeana Mirador theme, and in future could potentially also house some of the Mirador-specific logic from IIIFViewer.

The outcome of this is a reduction of ~ 300 kb uncompressed network traffic on CSRs and SSRs for IIIF item pages, and a typical Lighthouse desktop performance boost of around 10 points.

Also:
* Fix the glitchiness around full screen & CSRs to Mirador item pages
* Use redux-saga instead of subscribing to Mirador's store. Because subscribing to the store has the listener be triggered on every change to the state.
* Just store the manifest when its request is post-processed.


packages/portal/src/components/iiif/IIIFViewer.vue
```diff
@@ -720,13 +583,13 @@
       },
 
       findDownloadLinkForPage2(pageId) {
```
lbiedinger, line 585: not related to this PR specifically, but we should rethink the way these method names are constucted, especially on this and the `findDownloadLinkForPage3` method the name is potentially confusing.

`findDownloadLinkForPageV3` & `findDownloadLinkForPageV2` would be clearer.

lbiedinger (approved):

packages/style/scss/iiif.scss
```diff
@@ -79,6 +79,7 @@
   .mirador-canvas-nav {
     background-color: rgba(255 255 255 / 95%);
     align-items: center;
+    height: auto !important;
```
LeoniePeters, line 82: This is no longer needed I think (it was to fix a glitch, which we shouldn't have anymore)

packages/portal/src/components/iiif/IIIFViewer.vue
```diff
@@ -94,170 +102,10 @@
...
-                  root: {
-                    display: 'none'
-                  }
-                }
-              }
-            }
+            europeana: window.MiradorTheme
+          },
+          createGenerateClassNameOptions: {
+            seed: 'europeana'
```
LeoniePeters, line 108: This is no longer needed.

packages/portal/src/components/iiif/IIIFViewer.vue
rwd: Agreed, done.

packages/style/scss/iiif.scss
rwd: Gone.

packages/portal/src/components/iiif/IIIFViewer.vue
rwd: Gone.

PR #1947, opened 2023-05-25
feat: tooltips item page

LeoniePeters:
- Adds tooltips to 
  - RightsStatementButton
  - ItemAddButton
  - ItemLikeButton
  - ItemLanguageSelector
- Adds reusability definitions for all rights statements in rightsStatement mixin
- Styles for link inside tooltip and tooltip icon button

### DIFF
--- a/packages/portal/src/components/item/ItemLanguageSelector.vue
+++ b/packages/portal/src/components/item/ItemLanguageSelector.vue
@@ -60,6 +60,12 @@
                 <span>{{ metadataLanguageLabel }}</span>
               </i18n>
             </b-link>
+            <b-button
+              v-b-tooltip.bottom
+              :title="$t('multilingual.translateMetadata')"
+              class="icon-info tooltip-button"
+              variant="light-flat"
+            />
           </b-col>
         </b-row>
       </b-container>
--- a/packages/portal/src/components/item/ItemLanguageSelector.vue
+++ b/packages/portal/src/components/item/ItemLanguageSelector.vue
@@ -23,7 +23,6 @@
                 v-else
                 path="multilingual.translateLanguage"
                 tag="span"
-                class="pr-1"
                 data-qa="translate item suggestion"
               >
                 <b-dropdown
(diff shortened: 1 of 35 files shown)


packages/portal/src/components/item/ItemLanguageSelector.vue
```diff
@@ -46,6 +45,12 @@
                   </b-dropdown-item>
                 </b-dropdown>
               </i18n>
+              <b-button
```
rwd, line 48: Why is this a button? Clicking it does not do anything.

packages/portal/src/lang/en.js
```diff
@@ -1045,6 +1046,14 @@ export default {
   "removeFilter": "Remove {filterLabel} filter",
   "reset": "Reset",
   "resourceWikimedia": "Resource from Wikimedia Commons",
+  "rights": {
+    "tooltip": {
+      "open": "{rightsStatementName} means you can use this item freely. {readMoreLink}",
+      "permission": "{rightsStatementName} means you may be able to use this item if you seek permission. {readMoreLink}",
+      "restricted": "{rightsStatementName}  means you can use this item with conditions. {readMoreLink}",
+      "readMore": "Read more at creativecommons.org"
```
rwd, line 1054: The hostname needs to be extracted from the rights statement URI as it's not always creativecommons.org

rwd (changes requested):

packages/portal/src/components/item/ItemLanguageSelector.vue
LeoniePeters: The bootstrap-vue docs examples use buttons for the tooltip trigger elements, see https://bootstrap-vue.org/docs/components/tooltip
It might not have to be a button, but it should be an element that can receive focus, so tooltips also are shown on focus (for example keyboard only navigation).

rwd: Understood, thank you.

rwd (approved):

PR #1956, opened 2023-05-31
refactor: logout without going to keycloak

rwd:


rwd:
This does not log you out fully, hence is insecure.

PR #1979, opened 2023-06-08
feat: replace API toggle with fulltext input

rwd:


rwd:
Going with #1982 instead

PR #1981, opened 2023-06-08
feat: add full-text advanced search field

rwd:


rwd:
Going with #1982 instead

PR #1982, opened 2023-06-08
feat: full-text advanced search field, in Bootstrap Dropdown

rwd:
Changes:
* In search query builder rule component, replace selects for field & modifier with Bootstrap Dropdowns
* To the field dropdown:
  * add full-text as the first option, with a tooltip
  * next, add aggregated fields as a group
  * next, add individual fields as a group
* From the newspapers and WWI theme searches, remove the API metadata/full-text toggle, and display a dismissible alert advising to use the full-text field in adv search instead
* When searching the Record API, look for the presence of a full-text adv search rule, and if present, query the full-text Record API, combining both full-text and metadata queries
* Move all adv search rules to API param `qf`, except any for full-text which are promoted to `query` with any other query going into a `qf`
* Refactor SearchInterface to have API params and options as computed properties
* Validate search query builder rules with Bootstrap Vue feedback component


packages/portal/src/components/search/SideDateFilter.vue
```diff
@@ -145,22 +145,4 @@
...
-        height: calc(1.5 * 2.5rem);
-        margin-left: calc(1.5 * 0.25rem);
-      }
-    }
-  }
-
-  .apply-icon {
-    transform: rotate(-90deg);
-    display: inline-block;
-  }
```
LeoniePeters, line 165: Why are these styles moved outside the component?

packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -294,22 +289,53 @@
...
     computed: {
       advancedSearchQueryCount() {
-        return this.userParams?.qa ? [].concat(this.userParams?.qa).length : 0;
+        return this.qa.length;
       },
       userParams() {
         return this.$route.query;
       },
+      apiParams() {
+        const params = ['boost', 'qf', 'query', 'reusability', 'sort'].reduce((memo, field) => {
```
LeoniePeters, line 298: Why is the sort param added?

rwd: Because it's convenient for testing to be able to put one's own sort param into the URL.

packages/portal/src/components/search/SearchQueryBuilderRule.vue
```diff
@@ -149,56 +145,111 @@
...
     computed: {
-      // If any field has a value, all are required. If none have a value, the
-      // rule will be ignored and none are required.
-      areAllRequired() {
-        return [this.value.term, this.value.field, this.value.modifier].some((value) => !!value);
+      aggregatedFieldOptions() {
+        return this.fieldOptions
+          .filter((field) => this.aggregatedFieldNames.includes(field.value));
+      },
+      dropdownSections() {
```
lbiedinger, line 167: We don't strictly need to have these as computed properties, this could all just be in the data for now. It will never change at the moment. 
Computed properties help split it up a bit though and I think this will help us in future should we need to have field specific modifiers.

packages/portal/src/components/search/SearchQueryBuilderRule.vue
```diff
@@ -6,104 +6,94 @@
...
-            class="d-inline-flex align-items-center"
+            <b-form-input
+              v-if="component === 'term'"
+              v-model="term"
+              :data-qa="`advanced search query builder: ${component} control`"
+              :placeholder="$t('search.advanced.placeholder.term')"
+              :state="validations.term.state"
+              @change="(value) => handleRuleChange('term', value)"
+            />
+            <b-dropdown
```
lbiedinger, line 49: should dropdowns have their own component?
There's a bit of logic to these now.

packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -347,12 +394,22 @@
         set(value) {
           this.$store.commit('search/setView', value);
         }
+      },
+      collection() {
+        return filtersFromQf(this.apiParams.qf).collection?.[0];
+      },
+      fulltextQas() {
+        // NOTE: this intentionally excludes the "does not contain" rules for full-text
```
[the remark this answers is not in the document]
rwd: Removed.

packages/portal/src/components/search/SearchQueryBuilderRule.vue
rwd: As it's dependent on data properties from the advancedSearch mixin, I don't think it can go into data on the component.

packages/portal/src/components/search/SearchQueryBuilderRule.vue
rwd: Done.

lbiedinger (approved):

PR #2036, opened 2023-08-09
feat: log to APM search interactions

rwd:


lbiedinger (approved):

packages/vue-session-id/README.md
```diff
@@ -0,0 +1,53 @@
+# `@europeana/vue-session-id`
+
+Vue plugin to generate a session ID using [`uuid`](https://www.npmjs.com/package/uuid)
+and store it in sessionStorage.
+
+## Installation
+
+Install the package:
+```sh
+npm install --save @europeana/vue-session-d
```
lbiedinger, line 10: > npm install --save @europeana/vue-session-id

lbiedinger (changes requested):
The user buttons and the copyright and type info section in the list view also cause events to fire when they're clicked on.

Adding an `@click.stop=""` attribute to the wrapper div in portal/src/components/user/UserButtons.vue looks like it successfully prevents this from happening.

rwd:
@lbiedinger to prevent clicking the user buttons from logging the search interaction, I have used @click.stop as you suggested, but put it into ItemPreviewCard instead of within UserButtons.

PR #2038, opened 2023-08-14
feat: cache Matomo visits and organization count

rwd:
* Matomo REST API auth token needs to be set in env var `MATOMO_AUTH_TOKEN`
* New cachers:
  * `collections:organisations:count` - caches the number of Europeana-scoped organization-type entities
  * `matomo:visits` - caches the number of Matomo visits yesterday to the configured site
* NOTE: no new cacher for the total number of items, as that can be derived from the total of all the type-specific items already cached in `items:type-counts`

### DIFF
--- a/packages/portal/src/cachers/index.js
+++ b/packages/portal/src/cachers/index.js
@@ -8,6 +8,7 @@ const runtimeConfig = defu(nuxtConfig.privateRuntimeConfig, nuxtConfig.publicRun
 
 const cacherNames = [
   'collections:organisations',
+  'collections:organisations:count',
   'collections:organisations:featured',
   'collections:places',
   'collections:places:featured',
@@ -16,7 +17,8 @@ const cacherNames = [
   'collections:topics',
   'collections:topics:featured',
   'items:recent',
-  'items:type-counts'
+  'items:type-counts',
+  'matomo:visits'
 ];
 
 const cacherModule = (cacherName) => {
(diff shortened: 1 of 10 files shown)


packages/portal/src/cachers/index.js
```diff
@@ -16,7 +17,8 @@ const cacherNames = [
   'collections:topics',
   'collections:topics:featured',
   'items:recent',
-  'items:type-counts'
+  'items:type-counts',
+  'matomo:visits'
```
lbiedinger, line 21: add the new cachers to the readme

packages/portal/src/cachers/matomo/visits.js
```diff
@@ -0,0 +1,30 @@
...
+      format: 'JSON',
+      idSite: config.matomo.siteId,
+      method: 'VisitsSummary.get',
+      module: 'API',
+      period: 'day',
+      'token_auth': config.matomo.authToken
+    }
+  });
+
+  const totalVisits = Object.keys(response.data)
```
lbiedinger, line 20: When there is no response data because for example the connection failed or the auth token doesn't work this will set the totalVisists to 0.
This may look odd on the promo page, shouldn't we check for a successful response and only update the number when some data came back?

lbiedinger (changes requested):

packages/portal/src/cachers/matomo/visits.js
rwd: Yep. Handling has been added.

packages/portal/src/cachers/index.js
rwd: Done.

lbiedinger (approved):

PR #2050, opened 2023-08-28
feat: landing automated card group

LeoniePeters:
- Migration to add 'Europeana numbers' genre to AutomatedCardGroup content type
- Refactors InfoCard to be reusable as card without link
- Adds new LandingAutomatedCardGroup component

### DIFF
--- a/packages/styleguide/styleguide.root.js
+++ b/packages/styleguide/styleguide.root.js
@@ -93,6 +93,7 @@ Object.assign(Vue.prototype, {
     to: route => route,
     href: () => null
   },
+  localePath: () => '/',
   $path: () => {
     return '/';
   },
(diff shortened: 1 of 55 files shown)


packages/styleguide/styleguide.root.js
```diff
@@ -98,6 +98,7 @@ Object.assign(Vue.prototype, {
     return '/';
   },
   $route: { query: {} },
+  localePath: (args) => args,
```
lbiedinger, line 101: flagged as duplicate, remove the one at the top and have this fallback to `/` if that's needed still.

lbiedinger (changes requested):

lbiedinger (approved):

PR #2054, opened 2023-08-31
refactor: load Mirador from CDN

rwd:
* Reverts to loading Mirador from CDN, to avoid build-time dependency on all of Mirador's dependencies
* Moves some of the Mirador handling logic from the IIIFViewer to a mixin under the portal workspace, including what was formerly in the mirador workspace
  * NB: the mixin could have gone in the mirador workspace, but that results in SSR ES require errors
  * Therefore, removes the mirador workspace


rwd:
Re EC-6478

rwd:
No, we won't do this.

PR #2057, opened 2023-09-01
feat: get page views from Matomo REST API

rwd:
**POC: do not merge**

* Adds a new server middleware accessible at `/_api/matomo/page-hits` which:
  * Accepts a URL query param `url`, which should be the full URL of a page on the website
  * Converts that URL to its canonical form, i.e. without the language code
  * Looks in the Redis cache to see if the number of hits for that canonical URL are stored, and if so respond with that
  * If not cached, for each of the website's UI languages, makes a separate request to the Matomo REST API to retrieve the number of hits on that localised version of the page URL over the past 30 days
    * i.e. if there are 25 UI languages, then 25 API requests are made
  * Get the sum of the hits in the different languages, cache it to Redis for 24 hours, and respond with it
* Adds a new component, `StatsPageHits`, which calls the server middleware to get the number of hits for the supplied URL and display it as a pill-style badge
* In `ItemHero`, render the `StatsPageHits` component, but only client-side


packages/portal/src/server-middleware/api/matomo/page-hits.js
```diff
@@ -0,0 +1,119 @@
...
+};
+
+const sum = (vals) => vals.reduce((memo, val) => memo + (val || 0), 0);
+
+const cacheKey = (url) => {
+  return `${CACHE_KEY_PREFIX}:${url}`;
+};
+
+// TODO: limit to URLs at PORTAL_BASE_URL
+const pageHits = async(url, langs) => {
```
LeoniePeters, line 36: Rename this to something more specific like `getHitsForAllLocales` to distinguish from `getHitsFromMatomo`

packages/portal/src/server-middleware/api/matomo/page-hits.js
```diff
@@ -0,0 +1,119 @@
...
+
+  const dayHits = dayStats.map((stats) => stats['nb_hits'] || 0);
+
+  return sum(dayHits);
+};
+
+const sum = (vals) => vals.reduce((memo, val) => memo + (val || 0), 0);
+
+const cacheKey = (url) => {
+  return `${CACHE_KEY_PREFIX}:${url}`;
```
lbiedinger, line 32: with URLs starting with `https://...` the `:` splits the cache key into a sub directory. It should probably be escaped.

PR #2058, opened 2023-09-01
feat: add configured recordApi to video manifest URL

lbiedinger:
to aid in previewing new content which is not yet available on the production API

### DIFF
--- a/packages/portal/src/pages/media/index.vue
+++ b/packages/portal/src/pages/media/index.vue
@@ -49,8 +49,10 @@
 
     computed: {
       manifest() {
-        return `${this.$config.europeana.apis.iiifPresentation.url}${this.id}/manifest?format=3`;
+        const recordEndpointParam = this.$config.europeana.apis.record.url ? '&recordApi=' + this.$config.europeana.apis.record.url.replace('/api/v2', '') : '';
+        return `${this.$config.europeana.apis.iiifPresentation.url}${this.id}/manifest?format=3${recordEndpointParam}`;
       },
+
       dashRequired() {
         return new WebResource({ ebucoreHasMimeType: this.mediaType }).requiresDashJS;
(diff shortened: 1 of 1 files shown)


packages/portal/src/pages/media/index.vue
```diff
@@ -49,8 +49,10 @@
 
     computed: {
       manifest() {
-        return `${this.$config.europeana.apis.iiifPresentation.url}${this.id}/manifest?format=3`;
+        const recordEndpointParam = this.$config.europeana.apis.record.url ? '&recordApi=' + this.$config.europeana.apis.record.url.replace('/api/v2', '') : '';
```
rwd, line 52: This will not always result in the correct recordUrl parameter.

Our IIIF Presentation API expects recordUrl to be just the base URL with no path, i.e. the origin. If portal.js is configured with a Record API URL of https://api.acceptance.eanadev.org/record, then this replace will not remove the path, resulting in a failing IIIF manifest URL of e.g. https://iiif.acceptance.eanadev.org/presentation/170/data_euscreenXL_EUS_67F3ED0D435191E8A739326CD5F5F9F8/manifest?format=3&recordApi=https://api.acceptance.eanadev.org/record

Instead, use `new URL(this.$config.europeana.apis.record.url).origin`

packages/portal/src/pages/media/index.vue
```diff
@@ -49,8 +49,10 @@
...
+        return `${this.$config.europeana.apis.iiifPresentation.url}${this.id}/manifest?format=3${recordEndpointParam}`;
```
rwd, line 53: Use the `URL` class to construct URLs.

rwd (changes requested):

rwd (approved):

PR #2064, opened 2023-09-05
feat: cta sections for landing pages

lbiedinger:


packages/portal/src/components/content/ContentPrimaryCallToAction.vue
```diff
@@ -1,10 +1,11 @@
 <template>
   <div class="primary-cta text-center">
-    <ContentRichText
-      :text="text"
-      :rich-text-is-card="false"
+    <!-- eslint-disable vue/no-v-html -->
+    <div
+      v-html="parseMarkdownHtml(text)"
       class="primary-cta-rich-text text-left"
```
LeoniePeters, line 6: There's a lint warning. Put `class` before `v-html`.

packages/portal/src/modules/contentful/queries/browse-landing-static-page.graphql
```diff
@@ -225,6 +225,17 @@ query BrowseLandingStaticPage(
...
           }
+          ... on PrimaryCallToAction {
+            name
+            text
+            relatedLink {
+              url
+              text
+            }
+            image {
+              ...illustrationFields
```
LeoniePeters, line 236: We are not using the image, so consider not requesting it. Once (if) we start using the CallToActionBanner on the landing page we'll use the image.

LeoniePeters (changes requested):

PR #2069, opened 2023-09-08
refactor: classify Europeana API plugins

rwd:
* Refactors plugins for Europeana APIs into classes, as subclasses of the new base class `EuropeanaApi`
* Introduces two classes for handling configuration of Europeana APIs:
  * `EuropeanaApiEnvConfig`: reads config from environment variables, handling public/private (i.e. client-side/server-side) scopes, and shared API keys
    * Used to populate Nuxt runtimeConfig
  * `EuropeanaApiContextConfig`: reads config from context, handling preferring URLs from custom HTTP request headers (via the store for client-side support), then env, then base URL
* Adds unit tests for all 16 scenarios that may determine API URL selection from the 4 2-state factors:
  * Env URL - public: set/unset
  * Env URL - private: set/unset
  * Custom HTTP header URL: set/unset
  * Request type: server-side/client-side


packages/portal/src/plugins/europeana/iiif/presentation.js
```diff
@@ -1,3 +1,7 @@
+import EuropeanaApi from '../apis/utils/base.js';
 import { BASE_URL as EUROPEANA_IIIF_BASE_URL } from './index.js';
```
lbiedinger, line 2: this isn't loading a BASE_URL

rwd: No, it's importing the base class that the others extend.

lbiedinger (approved):

PR #2071, opened 2023-09-13
feat: item download tracking updates

lbiedinger:


packages/portal/src/components/download/DownloadButton.vue
```diff
@@ -58,6 +58,10 @@
           return '_blank';
         }
         return '_self';
+      },
+      canonicalPageUrl() {
```
rwd, line 62: Is there a reason we can't use the `canonicalUrl` mixin for this?

lbiedinger: only that `canonicalUrlWithoutLocale` uses $route.fullPath, which includes params & anchor tags, which would potentially cause different URLs being logged.

rwd: I have refactored the mixin to have a single method supporting with/without locale, and path/fullPath.

LeoniePeters (approved):

PR #2072, opened 2023-09-14
refactor: load Mirador from CDN & custom pkg

rwd:


rwd:
No, we won't do this.

PR #2075, opened 2023-09-19
feat: custom media presentation component

rwd:
* nests the item page Mirador (for IIIF items) and Swiper (for non-IIIF) components into a parent component ItemMediaLegacy
* introduces an alternative custom media presentation component ItemMediaPresentation intended to replace the combination of Mirador and Swiper for presenting all item page media through a single component
  * if supplied a URI for a IIIF Presentation manifest will fetch it and normalize and reduce v2 and v3 responses into the minimal data we are interested in on the item page
  * if instead supplied an array of web resources (from the Record API), will convert those into a similar data structure as though IIIF canvases
  * for any images in the content gathered from either of those approaches, will render them using OpenLayers and can render either IIIF Image API tile images or static images
    * OpenLayers controls for zooming and going full-screen are activated but not styled
  * for any AV media playable via Europeana Media player, renders an iframe that uses it to do so
  * for any PDFs, renders an iframe that loads the PDF via the media proxy, to use the browser's native PDF functionality
    * **NOTE**: intended as a further demo of how we would handle additional media types, but will very likely need to be placed behind a user interaction like clicking a thumbnail first as some of the PDFs are huge in file size
  * for any other content formats, will for now just output the data it has in a code block, to be replaced with handlers for various formats in future iterations before going live
  * has a demo sidebar for showing the IIIF manifest URL
  * has demo pagination using the PaginationNav component, which updates the URL for the selected page
  * has a crude thumbnail display component
* simplifies the logic for rending the ItemLanguageSelector component by always doing so on the item page, not passing it down to child component(s) in a slot which appears redundant since recent refactoring of the non-IIIF swiper

By default, at build-time, both legacy and new presentation components will be built, which increases bundle & image sizes, but permits specification of which to use at run-time. To enable the new presentation component at run-time, set:
```sh
ENABLE_NEW_MEDIA_PRESENTATION=1
```

To only build one of the legacy or new components, for benchmarking purposes, use the build- and run-time variable `BUILD_MEDIA_PRESENTATION_PLUGIN`:
```sh
# only build and use the legacy component:
BUILD_MEDIA_PRESENTATION_PLUGIN=legacy
# or, only build and use the new component:
BUILD_MEDIA_PRESENTATION_PLUGIN=new
```

These options work via the plugins in packages/portal/src/plugins/media, each of which makes ItemMediaPresentation globally available in a different way. These are only intended for temporary use while the new media presentation functionality is under development.


rwd:
Re EC-6584

packages/portal/src/components/iiif/IIIFMiradorViewer.vue
```diff
@@ -44,6 +44,11 @@
         default: null
       },
 
+      webResources: {
```
lbiedinger, line 47: These aren't being passed in to the viewer. They are only needed for the swiper one level up.
No need for this prop.

lbiedinger (approved):

packages/portal/src/components/iiif/IIIFMiradorViewer.vue
rwd: Indeed. It's gone.

PR #2093, opened 2023-10-12
fix: add default baseUrls for cachers

lbiedinger:


packages/portal/src/cachers/utils.js
```diff
@@ -2,6 +2,8 @@ import axios from 'axios';
 import { createClient } from 'redis';
 import _pick from 'lodash/pick.js';
 import { daily, langMapValueForLocale } from '../plugins/europeana/utils.js';
+import EuropeanaRecordApi from '../plugins/europeana/record.js';
+import EuropeanaEntityAPi from '../plugins/europeana/entity.js';
```
LeoniePeters, line 6: Only import the BASE_URL's

PR #2094, opened 2023-10-16
feat: log events to PostgreSQL db

rwd:
* Adds a PostgreSQL service dependency, where an events schema holds tables for logging user interaction events with items
  * Added to Docker Compose config, plus pgadmin
  * SQL to init schema is in file `docker/compose/postgres/docker-entrypoint-initdb.d/db.psql`
* Adds middleware to write event logs to new db, and return top trending items based on user activity in the past 24 hours
  * Behind feature toggle `ENABLE_EVENT_LOGGING=1`
* Adds a mixin `logEvent` to post an event to the middleware, used when liking items, viewing items, adding items to a gallery, or downloading an item's media
  * Behind feature toggle `ENABLE_EVENT_LOGGING=1`
* Adds a page at `/trending` to show the trending items


packages/portal/src/server-middleware/api/events/log.js
```diff
@@ -0,0 +1,92 @@
...
+    await client.query(`
+      INSERT INTO events.actions (object_id, action_type_id, session_id, occurred_at)
+      SELECT $1, at.id, $2, current_timestamp
+      FROM events.action_types at
+      WHERE at.name=$3
+      `,
+    [objectRow.id, sessionRow.id, actionType]
+    );
+  } catch (err) {
+    console.error(err);
```
lbiedinger, line 90: We'll need to log this too.

packages/portal/src/mixins/canonicalUrl.js
```diff
@@ -1,10 +1,11 @@
 export default {
   methods: {
-    canonicalUrl({ locale = true, fullPath = true } = {}) {
-      let path = fullPath ? this.$route.fullPath : this.$route.path;
+    canonicalUrl({ fullPath = true, locale = true, route = null } = {}) {
```
lbiedinger, line 3: where is the route being passed into canonicalUrl?

packages/portal/nuxt.config.js
```diff
@@ -39,6 +39,22 @@ const redisConfig = () => {
   return redisOptions;
 };
 
+const postgresConfig = () => {
+  const postgresOptions = {
+    enabled: featureIsEnabled('eventLogging'),
+    connectionString: process.env.POSTGRES_URL
```
lbiedinger, line 45: the env variable should probably be POSTGRES_CONNECTION_STRING
We could still have POSTGRES_URL/PORT/USER/PASSWORD set independently if we needed finer level control.

packages/portal/src/server-middleware/api/events/log.js
rwd: Log to where? The console? `console.error` does log to the console.

packages/portal/nuxt.config.js
rwd: It's been named `POSTGRES_URL` for consistency with our other env vars, and it is a URL, just that `pg` pkg has it named `connectionString`. There is no separate option to `pg` named `url`.

packages/portal/src/server-middleware/api/events/log.js
lbiedinger: To the APM I was thinking. Although I see we're not doing that for redis connection errors either, just those would at least respond with an error status in the response.

packages/portal/nuxt.config.js
lbiedinger: a connectionString by any other name would establish a connection as successfully.

lbiedinger (approved):

PR #2095, opened 2023-10-18
feat: don't track searches after native browser navigation

lbiedinger:
Relying on `window.onpopstate` in the default layout and setting navigation state in the store.


rwd:
This approach is not suitable, due to the Vue Router also using that event, and not therefore being able to reliably act on that event before the routing has occurred.

PR #2161, opened 2023-12-12
fix: prevent duplicate event log requests

rwd:
Before the initial request is complete, the `eventMayBeLogged` watcher will also trigger and send another identical request, i.e. before `eventLogged` has been set to `true`. Prevent this via an additional progress-monitoring boolean, `eventBeingLogged`.

SQL migration to clean up, and prevent recurrence, to be run **after** code deployment.

### DIFF
--- a/docker/compose/postgres/model/06-clean-up-duplicate-events-actions.psql
+++ b/docker/compose/postgres/model/06-clean-up-duplicate-events-actions.psql
@@ -1,0 +1,30 @@
+SET search_path TO EVENTS;
+
+-- remove the event actions, leaving one of each
+DELETE
+FROM actions
+WHERE id in
+    (SELECT id
+     FROM actions a
+     INNER JOIN
+       (SELECT object_id,
+               action_type_id,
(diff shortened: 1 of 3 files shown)


docker/compose/postgres/model/06-clean-up-duplicate-events-actions.psql
```diff
@@ -0,0 +1,30 @@
...
+     INNER JOIN
+       (SELECT object_id,
+               action_type_id,
+               session_id,
+               count(a.id) num,
+               min(a.id) first_action_id
+        FROM actions a
+        INNER JOIN objects o ON a.object_id=o.id
+        INNER JOIN action_types AT ON a.action_type_id=at.id
+        INNER JOIN sessions s ON a.session_id=s.id
```
lbiedinger, line 18: these inner joins don't really do anything.

lbiedinger (changes requested):

lbiedinger (approved):

PR #2166, opened 2023-12-18
refactor: remove Mirador

rwd:
This is not intended to be merged as-is, as it removes IIIF viewer functionality from the app, but is for the purposes of comparing build times, bundle sizes, etc w/ vs w/o Mirador.


rwd:
See #2387

PR #2208, opened 2024-02-20
feat: exact match modifier for fulltext

lbiedinger:


packages/portal/src/mixins/advancedSearch.js
```diff
@@ -12,7 +12,7 @@ const FIELD_SUGGEST_CONCEPT = 'concept';
 const FIELD_SUGGEST_PLACE = 'place';
 
 const advancedSearchFields = [
-  { name: 'fulltext', type: FIELD_TYPE_FULLTEXT },
+  { name: 'fulltext', type: FIELD_TYPE_FULLTEXT, supportsExact: true },
```
rwd, line 15: Instead of adding supportsExact to specific fields, derive it from the fact the the field type has a query property in the exact modifier.

packages/portal/src/lang/en.js
```diff
@@ -1078,6 +1078,7 @@ export default {
         "individual": "Individual fields"
       },
       "modifiers": {
+        "exact": "exact match",
```
rwd, line 1081: It would be better if this made grammatical sense when the full rule is read, which `Full-text exact match european union` does not (whereas `Full-text contains european union` does).

packages/portal/src/mixins/advancedSearch.js
```diff
@@ -95,17 +101,23 @@ export default {
   },
 
   methods: {
+    advancedSearchFieldByName(name) {
+      return this.advancedSearchFields.find((field) => field.name === name);
+    },
```
rwd, line 106: Add a blank line after each of the new methods for consistency with what pre-existed.

packages/portal/src/components/search/SearchQueryBuilderRule.vue
```diff
@@ -152,10 +152,15 @@
           ],
           modifier: [
             {
-              options: this.advancedSearchModifiers.map((mod) => ({
-                value: mod.name,
-                text: this.$t(`search.advanced.modifiers.${mod.name}`)
-              }))
+              options: this.advancedSearchModifiers.map((mod) => {
+                if (mod.name === 'exact' && (this.field !== this.fulltextFieldName)) {
```
rwd, line 156: This should be derived instead of hard-coded. See comment re `supportsExact` in the mixin.

rwd (changes requested):

rwd:
See #2256

PR #2227, opened 2024-02-29
chore: separate landing page CTF graphql query

rwd:
* create landingPage mixin with hard-coded list of known landing page slugs
* separate landing page into own CTF graphql query, and request tags on landing page image assets
* use landingPage mixin to determine if pages/index should request landing page data or browse/static page data
* remove distinct ds4ch page, relying on dynamic layout
* use landingPage mixin to determine if LandingPage component should use pro or ds4ch variant
* extend pageMeta mixin to permit overriding suffix for page title element, and set that to null for ds4ch layout


packages/portal/src/modules/contentful/queries/browse-static-page.graphql
```diff
@@ -0,0 +1,183 @@
...
+      }
+      relatedLinks {
+        ...linkGroupFields
+      }
+    }
+  }
+
+}
+fragment imageFields on Asset {
+  title
```
LeoniePeters, line 93: We no longer need the title field as that was only added for image exceptions which we will now handle using CTF tags.

LeoniePeters: Make sure to remove also from the landing-page query.

PR #2235, opened 2024-03-06
feat: derive landing page section IDs

rwd:


lbiedinger (changes requested):
we'll need to add:
```nameEN: name(locale: "en-GB")```
to the `imageCardFields` fragment at least, as without that the about-us link isn't working (since the ID is empty now)

However also the `cardGroupFields` don't have a `name` field at all, so either we add:
```nameEN: headline(locale: "en-GB")``` there, which is a bit funny because it's coercing the headline field into the name, or we extend the `sectionId` method to look in multiple places.

File: packages/portal/src/components/landing/LandingEmbed.vue [hunk 1]
```diff
@@ -64,12 +62,6 @@
         type: Object,
         default: null
       }
-    },
-
-    data() {
-      return {
-        containerId: kebabCase(this.englishTitle)
```
LeoniePeters, line 71: `englishTitle` prop is not used anywhere anymore, so can be removed

rwd: Removed.

lbiedinger (approved):

PR #2238, opened 2024-03-06
refactor: move klaro code to mixin

rwd:
* move klaro code to mixin
  * remove redundant TODOs re i18n in klaro code
* use mixin in both default and ds4ch layout
* permit whitelisting of klaro services to notify users about
  * default layout has all
  * ds4ch layout has only i18n and auth strategy


packages/portal/src/mixins/klaro.js
```diff
@@ -0,0 +1,123 @@
...
+    },
+
+    trackKlaroClickEvent(eventName) {
+      this.$matomo?.trackEvent('Klaro', 'Clicked', eventName);
+    },
+
+    klaroConfig() {
+      // TODO: uncomment when we have translations
+      // const locale = $i18n.locale;
+      // TODO: remove when we have translations
```
LeoniePeters, line 75: Are these TODO's still relevant? I see the cookie banner is translated.
I'm not sure how it picks up on the translations though as the locale is still hard coded to 'en' below.

rwd: TODOs no longer relevant. It was translating anyway because $i18n was translating per the user's UI locale and supplying that to Klaro, just with the `en` key.

packages/portal/src/layouts/ds4ch.vue
```diff
@@ -19,20 +19,37 @@
...
+      PageCookieConsent: () => import('@/components/page/PageCookieConsent')
+    },
+
+    mixins: [
+      klaroMixin
+    ],
+
+    data() {
+      return {
+        klaroServices: ['auth-strategy', 'i18n']
```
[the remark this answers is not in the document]
rwd: Done.

PR #2244, opened 2024-03-11
chore: DS footer links

LeoniePeters:
- Moves footer link group styles to LinkGroup
- Adds LinkGroup to DS4CHPageFooter
- Adds a visually hidden h2 for correct heading structure and easier nav for keyboard users

### DIFF
--- a/packages/portal/src/components/generic/LinkGroup.vue
+++ b/packages/portal/src/components/generic/LinkGroup.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div :class="`link-group-${variant}`">
     <component
       :is="titleTag"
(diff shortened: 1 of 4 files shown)


packages/portal/src/components/generic/LinkGroup.vue
```diff
@@ -77,19 +75,12 @@
       },
 
       /**
-       * Optional class for styling of links
+       * Variant for specific styles
+       * @values light, social
```
lbiedinger, line 79: We always use the "light" variant. Let's just remove that and call it "default", then the only variant needs to be "social".

lbiedinger (changes requested):

PR #2246, opened 2024-03-12
refactor: mv home page to component

rwd:
* mv the logic for home page retrieval and display from pages/home/index.vue to components/home/HomePage.vue
* make the catch-all pages/index.vue detect when there is no slug in the route and render HomePage
* DRY up the browse/static/landing page logic in pages/index.vue
* stop using a random image from the home page image set for the social media image as it should always have a specific one set
* rm redundant slug property from index page

### DIFF
--- a/packages/portal/nuxt.config.js
+++ b/packages/portal/nuxt.config.js
@@ -406,13 +406,6 @@ export default {
       'redirects'
     ],
     extendRoutes(routes) {
-      const nuxtHomeRouteIndex = routes.findIndex(route => route.name === 'home');
-      routes[nuxtHomeRouteIndex] = {
-        name: 'home',
-        path: '/',
-        component: 'src/pages/home/index.vue'
-      };
-
       const nuxtCollectionsPersonsOrPlacesRouteIndex = routes.findIndex(route => route.name === 'collections-persons-or-places');
       routes.splice(nuxtCollectionsPersonsOrPlacesRouteIndex, 1);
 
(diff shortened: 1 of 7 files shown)


packages/portal/nuxt.config.js
```diff
@@ -75,6 +75,7 @@ export default {
           origin: [process.env.PORTAL_BASE_URL].concat(process.env.APP_FEEDBACK_CORS_ORIGIN?.split(',')).filter((origin) => !!origin)
         }
       },
+      homeSlug: process.env.APP_HOME_SLUG,
```
lbiedinger, line 78: this isn't used anywhere, would the idea be to use it as an identifier in the homePage contentful query?
Does it need a default value here?

packages/portal/src/components/home/HomePage.vue
```diff
@@ -82,31 +83,16 @@
...
+          ogImage: this.$contentful.assets.optimisedSrc(this.socialMediaImage, { w: 1200, h: 630, fit: 'fill' })
         };
-      },
-
-      pageSubHeadline() {
-        return this.$t('homePage.subHeadline');
-      },
-
-      headMetaOgImage() {
-        const image = this.socialMediaImage ? this.socialMediaImage : this.backgroundImage?.image;
```
lbiedinger, line 96: if we're not using this fallback because the homepage _should_ have a distinct social media image, shouldn't we also make the social media image a required field?

packages/portal/nuxt.config.js
rwd: Removed to avoid distraction from what this PR is for.

packages/portal/src/components/home/HomePage.vue
rwd: Background image isn't required either, so, no.

lbiedinger (approved):

PR #2247, opened 2024-03-12
fix: div on v-for instead of template

rwd:
To prevent Nuxt re-fetching child components, causing janky layout shifts & messing with anchor links.

* Static import of all potential child components in LandingPage and LandingSubSection


rwd:
See #2248

PR #2248, opened 2024-03-12
fix: scroll to hash

LeoniePeters:
- Uses router link in hash paths in SmartLink
- Adds fixed heights to swiper to prevent layout shift while loading
- Replaces template with a div on v-for LandingPage to prevent re-fetch
- Removes the scrollToRouteHash mixin to depend on browser-native (and/or Nuxt router under the hood?) functionality
- Refactors visible-on-scroll directive to not hide top header when header is mounted on a page with hash route or when hash is changed


packages/portal/src/pages/index.vue
```diff
@@ -111,6 +111,8 @@
       } else {
         this.$error(404, { scope: 'page' });
       }
+
+      this.page = Object.freeze(this.page);
```
lbiedinger, line 115: Does this do anything still?

rwd: Just a general optimisation on that large data object to prevent Vue from making the whole thing reactive.

lbiedinger (approved):

PR #2256, opened 2024-03-19
refactor: adv search rule v-model

rwd:
* extend the use of v-model throughout the (advanced) search query builder components, and search query options
* move validation logic for field rules into the advancedSearch mixin
* have only the main query builder component trigger the validations when needed, then pass validation states on to child components
* add exact match modifier, only for full-text field
* in ContentCard, limit number of words displayed from hit text prefix and suffix to 15 each, to increase likelihood of highlighted exact match part showing in the card body
* switch to using v-click-outside package to detect when to apply advanced and standard search updates without form submission, e.g. when clicking elsewhere on the page having made changes to the form input
  * also when focus changes, so that tabbing out of forms applies updates


packages/portal/src/components/search/SearchQueryBuilder.vue
```diff
@@ -113,64 +111,56 @@
...
+          const allRulesValid = this.validations.every(this.advancedSearchRuleValidationsPass);
+          if (allRulesValid) {
             this.trackAdvancedSearch();
             this.$store.commit('search/setLoggableInteraction', true);
-            this.$router.push(this.advancedSearchRouteQueryFromRules(this.validQueryRules));
+            this.$router.push(this.advancedSearchRouteQueryFromRules(this.nonEmptyQueryRules));
           }
         });
       },
       handleValidRule(index) {
```
LeoniePeters, line 136: This doesn't seem to be used anymore.

packages/portal/src/components/search/SearchQueryBuilder.vue
```diff
@@ -113,64 +111,56 @@
...
-      handleChangeRule(index, field, value) {
-        this.queryRules[index][field] = value;
-        const validRule = this.checkIfValidRule(this.queryRules[index]);
-        if (validRule || field === 'term') {
+      handleRuleChange(newVal) {
+        if (this.advancedSearchRuleIsValid(newVal)) {
           this.handleSubmitForm();
         }
       },
       handleInvalidRule(index) {
```
LeoniePeters, line 119: This doesn't seem to be used anymore

lbiedinger (approved):

PR #2257, opened 2024-03-20
feat: institutions page

LeoniePeters:
- Updates the organisations cacher to retrieve the country from entity API and manipulate the value to a prefLabel: 
  - Production entity API returns a country code: uses `i18n-iso-countries` to get prefLabels. (no Maltese (mt) support)
  - Acceptance entity API returns entity URI. Does a prefLabel look up for each country
  - Localises the country prefLabel
- Adds search/filter feature to the EntityTable
- Adds pagination to the EntityTable
- Adds the country column for organisations type table
- Adds a row details slot for organisations type table to toggle display of (more) data on smaller screens
- Saves and applies the filter, sort and page queries in the URL
- Updates the search form styles (reused on Stories page)
- Prevents URL query parameters from Stories and collections type pages to persist to Search pages
- Adds a description (when available) and share button to the collections type page

### DIFF
--- a/packages/portal/src/cachers/collections/organisations.js
+++ b/packages/portal/src/cachers/collections/organisations.js
@@ -2,7 +2,7 @@ import baseData from './index.js';
 import { createEuropeanaApiClient } from '../utils.js';
 import { isEntityUri } from '../../plugins/europeana/entity.js';
 import uniq from 'lodash/uniq.js';
-// TODO: remove and uninstall when deprecated afte API released with place references for countries
+// TODO: remove and uninstall when deprecated after API released with place references for countries
 import countryCodes from 'i18n-iso-countries';
 import localeCodes from '../../plugins/i18n/codes.js';
 
--- a/packages/portal/src/cachers/collections/organisations.js
+++ b/packages/portal/src/cachers/collections/organisations.js
@@ -47,7 +47,7 @@ const data = async(config = {}) => {
       const entityId = country.split('/').pop();
       organisationCountriesPrefLabels[country] = await getCountryPrefLabel(`/place/${entityId}.json`);
     } else if (country) {
-      // Production Entity API returns country code. This as in between solution. No support for Maltese (mt)
+      // Production Entity API returns country code. This as in between solution.
       // TODO: remove when deprecated after API released with place references for countries
       const countryPrefLabelForLocale = {};
       for (const locale of localeCodes) {
--- a/packages/portal/src/cachers/collections/organisations.js
+++ b/packages/portal/src/cachers/collections/organisations.js
@@ -6,7 +6,8 @@ import uniq from 'lodash/uniq.js';
 import countryCodes from 'i18n-iso-countries';
 import localeCodes from '../../plugins/i18n/codes.js';
 
-const PICK = ['slug', 'recordCount', 'prefLabel', 'country', 'countryPrefLabel'];
+const PICK = ['slug', 'recordCount', 'prefLabel', 'countryPrefLabel'];
+const LOCALISE = 'countryPrefLabel';
 
 let axiosClient;
 let axiosClientEntity;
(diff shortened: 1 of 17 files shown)


packages/portal/src/components/entity/EntityTable.vue
```diff
@@ -107,8 +172,10 @@
         let collections = response.data[this.cacheKey];
         if (this.type === 'organisations') {
           collections = collections.map(this.organisationData);
+          this.collections = collections; // Do not freeze as _showDetails prop needs to be reactive for toggling the details display on small screens
```
rwd, line 175: Shouldn't this be for all entity types, not just organisations?

packages/portal/src/components/entity/EntityTable.vue
```diff
@@ -120,18 +187,34 @@
...
+    watch: {
+      '$route.query.filter'() {
+        this.filter = this.$route.query.filter;
+        this.updateRouteQuery({ page: 1 });
+      },
+      '$route.query.page'() {
+        this.currentPage = Number(this.$route?.query?.page) || 1;
+      },
+      '$route.query.sort'() {
+        this.sortBy = this.$route.query.sort.split(' ')[0];
```
rwd, line 203: As the logic for these is duplicated here and in data, could we use a computed getter/setter instead?

packages/portal/src/components/stories/StoriesTagsDropdown.vue
```diff
@@ -13,6 +13,7 @@
     >
       <b-form
         class="search-form"
+        :class="{ 'show': showDropdown}"
```
rwd, line 16: Add a space before the closing brace.

packages/portal/src/cachers/collections/organisations.js
```diff
@@ -17,24 +24,59 @@ async function getRecordCounts() {
...
+  const organisationCountries = uniq(organisationData.map(organisation => organisation.country));
+
+  const organisationCountriesPrefLabels = {};
+  for (const country of organisationCountries) {
+    // Acceptance Entity API returns entity URI for country
+    if (isEntityUri(country)) {
+      const entityId = country.split('/').pop();
+      organisationCountriesPrefLabels[country] = await getCountryPrefLabel(`/place/${entityId}.json`);
+    } else if (country) {
+      // Production Entity API returns country code. This as in between solution. No support for Maltese (mt)
```
rwd, line 50: No support for Maltese? It's in the list on the pkg readme: https://www.npmjs.com/package/i18n-iso-countries

packages/portal/src/lang/en.js
```diff
@@ -984,10 +984,13 @@ export default {
   "pages": {
     "collections": {
       "organisations": {
+        "description": "Explore the institutions which share digital cultural heritage items on Europeana.eu.",
         "table": {
-          "name": "Name"
+          "country": "Country",
+          "name": "Name",
+          "showMoreData": "Show more data for {organisation}"
```
rwd, line 991: This may be reused in other entity types in future, and does not contain any org-specific text, so move to the shared `table` group, and replace `{organisation}` w/ `{entity}`.

rwd (changes requested):

packages/portal/src/components/entity/EntityTable.vue
LeoniePeters: Previously we freezed all the collections objects. Since we started using the row details for organisations on this PR,  we need the collection objects to be reactive. Under the hood the BoostrapVue row details functionality sets and toggles the `_showDetails` prop on each collection. At the moment, only the organisations type uses this functionality. So I we can keep freezing the collections for the other types. I'm assuming that's better for performance.

packages/portal/src/components/entity/EntityTable.vue
LeoniePeters: Yes and that's done.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
LeoniePeters: Done.

packages/portal/src/lang/en.js
LeoniePeters: Moved.

packages/portal/src/cachers/collections/organisations.js
rwd: Excellent!

PR #2277, opened 2024-04-18
chore: update fulltext Record API URL

rwd:
to official URL: https://api.europeana.eu/fulltext


lbiedinger (changes requested):
There's still a reference in tests/e2e/docker/nginx/conf.d to `newspapers.eanadev.org` let's update that as well.

lbiedinger (approved):

PR #2284, opened 2024-04-22
chore: mv authScheme to auth directory

rwd:
Because it is not a Nuxt plugin


packages/portal/tests/unit/auth/schemes/authScheme.spec.js
```diff
@@ -1,5 +1,5 @@
 import sinon from 'sinon';
-import * as plugin from '@/plugins/authScheme';
+import * as plugin from '@/auth/schemes/authScheme';
```
LeoniePeters, line 2: Consider renaming `plugin` to `scheme`.

rwd: Done.

LeoniePeters (approved):

PR #2285, opened 2024-04-22
refactor: mv visible-on-scroll directive to package

rwd:
And register on each of the components that need it, not globally.


lbiedinger (commented):
the `v-` prefix is applied to the directive name, which is being imported as `visibleOnScrollDirective` shouldn't the new package just be named `visibleOnScroll` or `visibleOnScrollDirective` then?

rwd:
@lbiedinger fair point. It's been renamed to `@europeana/vue-visible-on-scroll`

packages/vue-visible-on-scroll/tests/index.spec.js
```diff
@@ -0,0 +1,187 @@
...
+        describe('when scrolling down', () => {
+          describe('when the element is visible already (should always be the case)', () => {
+            it('updates the scrollPosition', () => {
+              const wrapper = factory();
+              const visibleOnScrollElement = wrapper.findComponent('p');
+              visibleOnScrollElement.element.scrollPosition = 20;
+              visibleOnScrollElement.element.scrolledVisible = true;
+
+              window.scrollY = 140;
+              window.dispatchEvent(new CustomEvent('scroll', {}));
```
rwd, line 62: Why are the CustomEvent instantiations called with an empty object as second argument throughout these tests?

lbiedinger: It was taken from an example that had some arguments in the custom event and then just copy pasted for all the specs, but it's not needed.
Has been removed.

lbiedinger (approved):

PR #2304, opened 2024-05-03
fix: use double quotes in lang files

lbiedinger:
Because Lokalise needs hand-holding.


lbiedinger:
Just update the english file instead: https://github.com/europeana/portal.js/pull/2306

The rest can be automatically generated with the lokalise action.

PR #2310, opened 2024-05-07
fix(workaround): retrieve each entity individually when finding

rwd:
So that Entity API performs redirects when old organisation IDs are used.


rwd:
#2311 was a non-workaround fix.

PR #2323, opened 2024-05-15
feat: remove from gallery button and toast on item add, remove, like or unlike

LeoniePeters:
- Adds new ItemRemoveButton component to user buttons
  - only displayed on user's own gallery (reusing in part the logic for the item move button)
- Makes toast on item like, unlike, add and remove
- Adds missing icons to the style guide
- Adds TransitionGroup to ItemPreviewCardGroup


packages/portal/src/components/DS4CH/DS4CHLandingHero.vue
```diff
@@ -5,6 +5,7 @@
       :style="imageCSSVars"
       role="img"
       :aria-label="heroImage?.image?.description"
+      :alt="heroImage?.image?.description"
```
lbiedinger, line 8: Why was included here?

packages/portal/src/components/item/ItemRemoveButton.vue
```diff
@@ -0,0 +1,73 @@
...
+       * Identifier of the item
+       */
+      identifier: {
+        type: String,
+        required: true
+      },
+      /**
+       * Button variant to use for styling the buttons
+       */
+      buttonVariant: {
```
lbiedinger, line 41: I guess this is okay to leave here for consistency with the other user buttons.
Although, since we don't have the list view available on the set pages this won't ever be used.

packages/portal/src/i18n/lang/en.js
```diff
@@ -1178,6 +1180,10 @@ export default {
     },
     "notifications": {
       "deleted": "Your gallery has been deleted.",
+      "itemAdded": "Item has been added to gallery {gallery}.",
```
lbiedinger, line 1183: Let's make this:
`"itemAdded": "Item has been added to the gallery \"{gallery}\".",`

lbiedinger (changes requested):

packages/portal/src/components/DS4CH/DS4CHLandingHero.vue
LeoniePeters: The lack of an alt attribute on this element was flagged as an a11y violation in the e2e tests. Not sure why that hasn't been flagged before.

packages/portal/src/components/item/ItemRemoveButton.vue
LeoniePeters: It is used to pass down the correct button variant and keep it aligned with the other buttons. We could hard code it, but like this is a bit more robust for future changes.

packages/portal/src/i18n/lang/en.js
LeoniePeters: Done.

lbiedinger (approved):

packages/portal/src/components/DS4CH/DS4CHLandingHero.vue
lbiedinger: I think it's because of the test data, probably before we cloned production into test, that hero didn't have an image at all so didn't expect an alt text for accessibility reasons.

PR #2336, opened 2024-05-28
feat: stories view counts

lbiedinger:


docker/compose/postgres/model/07-clean-up-blog-objects.psql
```diff
@@ -0,0 +1,37 @@
...
+INSERT INTO objects (uri)
+  (SELECT DISTINCT REPLACE(uri, '/blog/', '/stories/')
+   FROM objects) ON CONFLICT DO NOTHING;
+
+-- on actions/history linked to /blog/ objects, update the link to be to the /stories/ object
+
+UPDATE actions
+SET object_id=good_id
+FROM
+  (SELECT bad.id bad_id,
```
rwd, line 14: These are not the result of a bug, so let's call them old and new instead of bad and good.

rwd (changes requested):

rwd (approved):

PR #2349, opened 2024-06-06
fix: stop importing icon scss in components

rwd:
It greatly bloats inline scss as is duplicated for all scoped style sections.


rwd:
Will open a new one based on v1.151.0.

PR #2362, opened 2024-06-12
refactor: move CTF assets plugin to utils

rwd:
* move Contentful assets helper functions from contentful module/plugin to a utils library
* rename exported functions to include "contentful"
* remove contentful store module
  * was used to detect and store the client's accepted image formats to only return WebP when they support it, but the browser versions we target all do support WebP, so it's extraneous
  * i.e. always return WebP unless format is specified or original is an SVG


rwd:
Superseded by #2770 and #2749

PR #2390, opened 2024-06-28
chore: add trending items to home

LeoniePeters:


packages/portal/src/i18n/lang/en.js
```diff
@@ -1239,6 +1238,7 @@ export default {
     "themes": "Themes",
     "description": "Browse by theme"
   },
+  "trendingItemsHeadline": "Explore today’s popular items",
```
rwd, line 1241: Let's try and avoid cluttering the top-level keys of the language files. Make this something like `items.trending.headline` instead.

rwd (changes requested):

rwd (approved):

PR #2408, opened 2024-07-26
fix: stories tags dropdown keyboard accessibility

lbiedinger:


packages/portal/src/components/stories/StoriesInterface.vue
```diff
@@ -243,6 +246,12 @@
         }
       },
 
+      setFocusToTypeFilter() {
```
LeoniePeters, line 249: Let's instead call `blur `on `esc` in StoriesTagsDropdown.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
```diff
@@ -9,7 +9,10 @@
     />
     <div
       ref="tagsdropdown"
+      v-click-outside="clickOutsideConfig"
       class="position-relative mb-4"
+      @focusin="handleFocusin"
```
LeoniePeters, line 14: Why is the focus event handler moved up to this div? Shouldn't it stay on the input as that's the element receiving focus?

packages/portal/src/components/stories/StoriesTagsDropdown.vue
```diff
@@ -41,9 +44,11 @@
       >
         <RelatedCategoryTags
           v-if="displayTags.length > 0"
+          ref="relatedCategoryTags"
           :tags="displayTags"
           :selected="selectedTags"
           :heading="false"
+          tabindex="-1"
```
LeoniePeters, line 51: I don't think this would have any effect as this is a div which receives no focus by default.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
```diff
@@ -30,7 +33,7 @@
           :aria-controls="showDropdown ? 'tags-options' : null"
           :aria-expanded="showDropdown"
           :aria-label="$t('categories.label')"
-          @focus="showDropdown = true"
+          @keydown.down="handleInputDown"
```
LeoniePeters, line 36: This only works for moving from the input to the first tag, and sometimes scrolls a bit (as that is the native behaviour for the up/down arrows in this case). I'm doubting if that could be confusing. In the main search form the options are ordered top to bottom and don't have a scrollbar. In the search facet dropdowns the up and down arrows navigate through the options instead of control the scroll bar. In the stories tag dropdown I don't think we want either as there are many options to key through before scrolling would happen. So I'd say remove this event listener.

LeoniePeters (changes requested):

packages/portal/src/components/stories/StoriesInterface.vue
LeoniePeters: I've pushed a commit to make that so. I think that's more correct and cleaner.

lbiedinger: This doesn't allow a keyboard only user to get past the tag search box without tabbing through all the options.
I think moving the focus to the following element is better UX overall.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
lbiedinger: yes, agreed, it's not the best. Will remove.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
lbiedinger: It fixes it in dev mode, where for some reason the div does get focus.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
lbiedinger: removed.

packages/portal/src/components/stories/StoriesTagsDropdown.vue
lbiedinger: It was to try and make sure the click outside config was always active when either the input or the dropdown are in focus. 
It's not possible to open the dropdown without focusing on the input first now though, so yeah I've moved the handler to the input.
This way there are less events being fired when moving focus around inside the dropdown. 👍
It still re-activates the config when tabbing back to the input, but at least it's not activating on just navigating within the dropdown.

packages/portal/src/components/stories/StoriesInterface.vue
lbiedinger: The issue with moving focus only exists in firefox. It is possible to get around by hitting esc twice, which moves the focus outside of the input and let's the user tab to the story type toggle.

LeoniePeters: After some testing and discussions this mainly seems a bit of an issue in FF. Other latest browsers seem to keyboard navigate as expected.

LeoniePeters (approved):

PR #2417, opened 2024-08-09
chore: create feature ideas page and feature content types

LeoniePeters:


packages/contentful/migrations/0143-create-features-page-content-type.cjs
```diff
@@ -0,0 +1,140 @@
...
+    .type('Symbol')
+    .localized(true)
+    .required(true)
+    .validations([{ size: { max: 100 },
+      message: 'Text must be max. 100 characters.' }])
+    .disabled(false)
+    .omitted(false);
+
+  voteableFeature
+    .createField('identifier')
```
lbiedinger, line 22: I'm pretty sure we can just use the contentful sys id instead of having to come up with our own ones.

packages/contentful/migrations/0143-create-features-page-content-type.cjs
```diff
@@ -0,0 +1,140 @@
...
+    .name('Description')
+    .type('Symbol')
+    .localized(true)
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false);
+
+  featureIdeasPage.changeFieldControl('description', 'builtin', 'singleLine', {
+    helpText: 'For SEO, please make sure there is a short description.'
```
lbiedinger, line 104: This is not just for SEO though, it's also what's displayed at the top of the page.

The description from the design already doesn't fit into the limit. If we want to ensure there is a short(er) SEO description perhaps we should split out another field for the displayed description which could then also use the rich-text editor for more flexibility down the line.

lbiedinger (changes requested):

packages/contentful/migrations/0143-create-features-page-content-type.cjs
```diff
@@ -0,0 +1,140 @@
...
+    .localized(false)
+    .required(false)
+    .validations([])
+    .disabled(false)
+    .omitted(false)
+    .items({
+      type: 'Link',
+      validations: [
+        {
+          linkContentType: ['richText', 'voteableFeature']
```
lbiedinger, line 135: ...and this won't need to handle richtText sections if we make the description be a once off field at the page level.

packages/contentful/migrations/0143-create-features-page-content-type.cjs
LeoniePeters: This description is meant to be only for SEO. The intro text on the design is indeed too long to be a description, and is also displayed below the 'content header' (below the share button), styled as a static page text section. So I modeled it as a rich text section. Under sections you can add a rich text section which should be used for the introduction text.

LeoniePeters: We could split it into a separate field on the page level, instead of as a section.

lbiedinger: yes, let's do that. Otherwise editors could add multiple richtexts, with only one showing as the description.

lbiedinger (approved):

PR #2427, opened 2024-09-03
chore: test coverage for feature voting

lbiedinger:


packages/portal/tests/unit/server-middleware/api/keycloak.spec.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+describe('server-middleware/api/keycloak', () => {
+  beforeAll(() => nock.disableNetConnect());
+  afterAll(() => nock.enableNetConnect());
+
+  describe('userId', () => {
+    it('extracts the ID from the user info', async() => {
+      keycloak.config = config;
+      nock(config.origin)
```
rwd, line 18: Include the authorization header here with [nock.reqheaders](https://github.com/nock/nock/tree/v13.3.3?tab=readme-ov-file#specifying-headers)

packages/portal/tests/unit/components/generic/FeatureIdeas.spec.js
```diff
@@ -52,24 +59,118 @@ describe('components/generic/FeatureIdeas', () => {
         expect(errorMessage.exists()).toBe(true);
       });
     });
-
-    // describe('when there are features', () => {
-    //   it('fetches the features votes from the API', async() => {
```
rwd, line 57: Let's write a spec for this.

rwd (changes requested):

packages/portal/tests/unit/server-middleware/api/keycloak.spec.js
lbiedinger: Yes! 
I had tried this, but kept getting errors with the requests not matching, so removed the headers to simplify it. It must have been something else and I never added it back in. Anyway, the headers are back in.

rwd (approved):

PR #2435, opened 2024-09-18
feat: add customisation options to PaginationNavInput component

rwd:
for:
* displaying next/prev buttons as text/icons/both
* optionally not showing the page input
* optionally showing a progress indicator
* optionally using a different icon
* adds a tooltip when there is not button text

Includes a refactor to
- create a `hideTooltips` mixin with a method to hide all tooltips (to not keep tooltips open on mouseclick + mouseleave)
- set the item media sidebar links tab tooltip on the button to be focusable (keyboard a11y)


rwd:
Re EC-6934

packages/portal/src/components/generic/PaginationNavInput.vue
```diff
@@ -81,6 +120,22 @@
       maxResults: {
         type: Number,
         default: null
+      },
+      buttonIcons: {
```
lbiedinger, line 124: When would we not want icons?
Looks like the intention was to use this prop to allow passing a class in so that we can have the circled or plain arrows.
At the moment this isn't used.

lbiedinger (changes requested):

packages/portal/src/components/generic/PaginationNavInput.vue
LeoniePeters: Well spotted. I've refactored it to `buttonsIconClass` to allow passing in the desired icon class.

lbiedinger (approved):

PR #2438, opened 2024-09-20
feat: thumbnail bar for new media presentation

lbiedinger:
- add a toggle button to show/hide the item thumbnails
- a11y
  - reorder elements to the visual order
  - add aria-attributes, tooltips and custom focus handling when opening/closing sidebars (both the thumbnail pages and left sidebar)
- add a new ItemMediaThumbnail component, to work on separately from the item swiper thumbnails
- refactor ItemMediaThumbnails to add a transition and more correct HTML list markup
- use scrollTo composable to scroll to the active thumbnail on change, page load (with page param set) or window resize.


packages/portal/src/components/item/ItemMediaThumbnails.vue
```diff
@@ -1,32 +1,94 @@
...
+    watch: {
+      selectedIndex: 'updateThubmnailScroll'
+      // TODO: consider watching resize/orientation changes to re-center thumbnails
+    },
+
+    methods: {
+      clickThumbnail(index) {
+        this.$router.push({ ...this.$route, query: { ...this.$route.query, page: index + 1 } });
+      },
+      updateThubmnailScroll() {
```
LeoniePeters, line 57: Fix typo.

packages/portal/src/components/item/ItemMediaThumbnails.vue
```diff
@@ -1,32 +1,94 @@
...
   >
-    <h2>Thumbnails</h2>
     <div
-      class="d-flex overflow-auto"
+      ref="mediaThumbnails"
+      class="media-thumbnails d-flex flex-row flex-lg-column"
     >
-      <b-img-lazy
-        v-for="(image, index) in thumbnails"
+      <ItemMediaThumbnail
```
LeoniePeters, line 9: We should eager load the first 5 thumbnails as they are always in view on page load.

packages/portal/src/components/item/ItemMediaThumbnail.vue
```diff
@@ -0,0 +1,170 @@
...
+  >
+    <MediaCardImage
+      :media="resource"
+      :lazy="lazy"
+      :offset="offset"
+      :edm-type="edmType"
+      thumbnail-size="small"
+      :linkable="false"
+    />
+    <span class="thumbnail-page">{{ `p. ${offset + 1}` }}</span>
```
LeoniePeters, line 14: Perhaps this should be a translation string, as I'm not sure the `p.` for page makes sense in all languages.

packages/portal/src/components/item/ItemMediaThumbnail.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+  <div
+    class="item-media-thumbnail text-lowercase"
+    data-qa="item media thumbnail"
+  >
```
LeoniePeters, line 5: This div should be a button element as clicking it causes an interaction. Buttons are also keyboard accessible. 
Currently the click is handled by the parent. Consider refactoring the div into a button (`b-button`) and perhaps moving the click event down to this button.

packages/portal/tests/unit/components/item/ItemMediaThumbnail.spec.js
```diff
@@ -0,0 +1,37 @@
+import { createLocalVue } from '@vue/test-utils';
+import { shallowMountNuxt } from '../../utils';
```
LeoniePeters, line 2: This can be just `shallowMount`

packages/portal/tests/unit/components/item/ItemMediaThumbnail.spec.js
```diff
@@ -0,0 +1,37 @@
...
+import ItemMediaThumbnail from '@/components/item/ItemMediaThumbnail';
+// import sinon from 'sinon';
+
+const localVue = createLocalVue();
+
+const factory = ({ propsData = {}, mocks = {} } = {}) => shallowMountNuxt(ItemMediaThumbnail, {
+  localVue,
+  attachTo: document.body,
```
LeoniePeters, line 10: Is this needed?

packages/portal/tests/unit/components/item/ItemMediaThumbnail.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  propsData,
+  mocks: {
+    $route: { query: {} },
+    $t: (key) => key,
+    ...mocks
+  }
```
LeoniePeters, line 16: `$route` is not used in the component ATTOW. And there are never any mocks passed in this file's tests.

LeoniePeters (changes requested):

packages/portal/src/components/item/ItemMediaThumbnails.vue
```diff
@@ -1,32 +1,94 @@
...
+      },
+      selectedIndex: {
+        type: Number,
+        required: true
+      }
+    },
+
+    watch: {
+      selectedIndex: 'updateThubmnailScroll'
+      // TODO: consider watching resize/orientation changes to re-center thumbnails
```
LeoniePeters, line 50: This is a good idea

packages/portal/src/components/item/ItemMediaThumbnail.vue
LeoniePeters: Better even than my previous comment, refactor the thumbnail into a link to the specific item media page.

packages/portal/src/components/item/ItemMediaThumbnails.vue
```diff
@@ -1,32 +1,94 @@
...
   <div
-    class="p-2"
+    class="d-flex flex-column position-relative"
   >
-    <h2>Thumbnails</h2>
     <div
-      class="d-flex overflow-auto"
+      ref="mediaThumbnails"
+      class="media-thumbnails d-flex flex-row flex-lg-column"
     >
```
LeoniePeters, line 8: Let's refactor this into an unordered list (`ul`) with list item children (`li`)

lbiedinger: I've made it an ordered list, since these pages are inherently in sequence. Is there a reason to use ul instead?

packages/portal/tests/unit/components/item/ItemMediaThumbnail.spec.js
lbiedinger: Since the refactor for the thumbnail to be a link, i've added some more specs, some of those now use the route.

packages/portal/src/components/item/ItemMediaThumbnails.vue
lbiedinger: I'm not convinced about this. Since the thumbnails are always fixed sizes, there's no layout shift from having them lazy load.
While usually 5 thumbnails appear, this is somewhat dependent on the screen size.
Further there's an edge case where the user opens a link with a page already selected, in which case it becomes trickier to identify which thumbnails need to be loaded already.

LeoniePeters: I see there are the grey squares reserving the space for the thumbnail images. And that is both when lazy or eager loaded, so doesn't seem to really make a difference. Let's leave it to lazy load for now.

LeoniePeters (approved):

PR #2439, opened 2024-09-24
feat: fetch and highlight IIIF annotations

rwd:
* upgrade ol (openlayers) package to latest version
* move EuropeanaMediaPresentation class to packages/portal/src/utils/europeana/media/presentation.js
* introduce base class EuropeanaMediaBase for working with media manifests etc, and transforming their data to that required for display
* introduce new sub-classes for working with annotations, with support for embedding external annotation text
* introduce a new component MediaAnnotationList for displaying annotations in the media presentation UI
* update MediaImageViewer to display a rectangle around the active annotation
* adds a plugin to use axios-cache-interceptor on media resource requests to cache http responses client-side, in particular to prevent many duplicate requests for the same (fulltext) annotation resource when embedding bodies


packages/portal/nuxt.config.js
```diff
@@ -476,21 +477,38 @@ export default {
...
+      'ol/reproj/DataTile.js',
+      'ol/reproj/Tile.js',
       'ol/source/IIIF.js',
       'ol/source/ImageStatic.js',
+      'ol/source/Source.js',
+      'ol/source/static.js',
+      'ol/source/Vector.js',
       'ol/structs/LRUCache.js',
+      'ol/style/Style.js',
+      'ol/style/RegularShape.js',
```
LeoniePeters, line 511: Are these all needed? I don't see them used directly, or are they dependencies for other OpenLayers modules we are using?

lbiedinger (approved):

PR #2440, opened 2024-10-03
feat: add scrollTo composable

rwd:
* adds a new Vue composable for scrolling to an element, or scrolling so that an element is centred in its container
* the composable will keep a queue of scroll requests and do one at a time, with a pause of 300 ms between, and only permitting each element to be on the queue once at a time
* replaces the use of vue-scrollto plugin with this composable

### DIFF
--- a/packages/portal/src/components/stories/StoriesInterface.vue
+++ b/packages/portal/src/components/stories/StoriesInterface.vue
@@ -1,6 +1,5 @@
 <template>
   <div
-    ref="container"
     data-qa="stories interface"
   >
     <client-only>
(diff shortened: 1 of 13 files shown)


packages/portal/src/components/stories/StoriesInterface.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div
+    ref="container"
```
LeoniePeters, line 3: Remove this.

rwd: Done.

PR #2477, opened 2024-11-12
feat:  annotation highlighting on hover

lbiedinger:


lbiedinger:
use https://github.com/europeana/portal.js/pull/2490 instead.

PR #2511, opened 2024-12-10
fix: ol image load check for createImageBitmap

rwd:


rwd:
This was just to verify the fix works. [Issue](https://github.com/openlayers/openlayers/issues/16457) reported upstream with a [fix](https://github.com/openlayers/openlayers/pull/16458).

PR #2520, opened 2024-12-16
perf: render item media presentation server-side

rwd:
- to prevent large layout shifts client-side
- but don't _fetch_ server-side due to slow/large manifests


rwd:
No. Results in a significant increase in Node memory usage. Trying #2522 instead.

rwd:
Resurrected in #2523

PR #2522, opened 2024-12-17
refactor: mv media viewer wrapper to ItemHero

rwd:
- so that client-side rendering of media viewer does not result in such big layout shifts
- but keeping the media viewer rendered only client-side due to significant increase in node memory usage otherwise


rwd:
Preferred approach in #2523

PR #2526, opened 2024-12-18
refactor: media viewer wrapper in ItemHero

rwd:
Because #2523 causes breaking hydration errors on some items, e.g. /09102/_GNM_693983


rwd:
Fixed by #2527 instead.

PR #2527, opened 2024-12-18
fix: mv ItemMediaPresentation client-only higher

rwd:
to prevent hydration errors on e.g. non-IIIF items w/ images


packages/portal/src/components/item/ItemMediaPresentation.vue
```diff
@@ -422,6 +424,7 @@
 
   .media-viewer-wrapper {
     position: relative;
+    background-color: $black;
```
LeoniePeters, line 427: Revert this to be scoped to `.media-viewer-inner-wrapper`. Otherwise the black bg will show briefly when closing the thumbnails menu on mobile where that looks a bit off.

PR #2533, opened 2025-01-09
refactor: recursive cookie services

lbiedinger:
Allow nesting of cookie sections.

A section may be an individual service or grouping. Groupings are based on klaro purposes.


rwd:
Re EC-7025

PR #2537, opened 2025-01-16
refactor: use klaro mixin in cookie components

rwd:
instead of in layouts, to prevent passing klaro manager/config down to components, and isolate use of the mixin methods to fewer locations

abstract waitForMatomo function to be a utility function for waiting for arbitrary callbacks to return truthy values, and also use it to have klaro-dependent components wait for it to become available before attempting to work with it

### DIFF
--- a/packages/portal/src/components/page/PageCookiesWidget.vue
+++ b/packages/portal/src/components/page/PageCookiesWidget.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    v-if="cookieConsentRequired"
+    v-if="!onlyShowIfConsentRequired || cookieConsentRequired"
   >
     <b-toast
       v-if="renderToast"
@@ -243,6 +243,10 @@
       klaroServices: {
         type: Array,
         default: null
+      },
+      onlyShowIfConsentRequired: {
+        type: Boolean,
+        default: true
       }
     },
 
--- a/packages/portal/src/components/page/PageCookiesWidget.vue
+++ b/packages/portal/src/components/page/PageCookiesWidget.vue
@@ -201,6 +201,7 @@
 
 <script>
   import PageCookiesCheckbox from './PageCookiesCheckbox';
+  import klaroMixin from '@/mixins/klaro.js';
 
   export default {
     // TODO: rename as this is more generally about services than solely cookies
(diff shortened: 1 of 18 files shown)


packages/portal/src/mixins/klaro.js
```diff
@@ -18,14 +23,13 @@ export default {
...
+    script.setAttribute('src', `https://cdn.jsdelivr.net/npm/klaro@${version}/dist/klaro-no-css.js`);
+    script.setAttribute('defer', true);
 
-    // If Matomo plugin is installed, wait for Matomo to load, but still render
-    // Klaro if it fails to.
-    const renderKlaroAfter = this.$waitForMatomo ? this.$waitForMatomo() : Promise.resolve();
-    renderKlaroAfter.catch(() => {}).finally(this.renderKlaro);
+    script.addEventListener('load', this.onKlaroScriptLoad);
+
+    document.head.appendChild(script);
```
LeoniePeters, line 32: It's possible to use the `head` method in components too. So we could set the script in there instead.

packages/portal/src/mixins/klaro.js
```diff
@@ -66,12 +70,30 @@ export default {
...
+      // If Matomo plugin is installed, wait for Matomo to load, but still render
+      // Klaro if it fails to.
+      const renderKlaroAfter = this.$waitForMatomo ? this.$waitForMatomo() : Promise.resolve();
+
+      renderKlaroAfter.catch(() => {}).finally(this.renderKlaro);
+    },
+
+    onKlaroRendered() {
+      // may be overriden by components including the mixin
+    },
```
LeoniePeters, line 87: This does nothing. Is it just for reference?

packages/portal/src/components/page/PageCookiesWidget.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <div>
+  <div
+    v-if="cookieConsentRequired"
+  >
```
LeoniePeters, line 4: Probably remove this as the widget is also used in the EmbedGateway when cookieConsentRequired can be false

packages/portal/src/mixins/klaro.js
rwd: No, but refactored so that it's redundant.

packages/portal/src/components/page/PageCookiesWidget.vue
rwd: Refactored to handle both scenarios.

LeoniePeters (approved):

PR #2538, opened 2025-01-16
feat: embed gateway for embed html

LeoniePeters:
- Wraps the EmbedGateway around the EmbedHTML component in ContentSection
- Wraps the EmbedGateway around the EmbedHTML component in LandingEmbed
- Adds logic to recognise iframes and scripts in an embed code string, if found, sets the providerUrl from its source and iframe dimensions if available
  - uses DOMParser, broswer-native on client-side, from jsdom on server-side
- Adds the multimedia icon as overlay background
- Adds Codepen to landing page cookie widget and refactors PageCookiesWidget to add the purposes conditionally
- Refactor to call `checkConsentAndOpenEmbed` on `watchKlaroManagerUpdate` instead of emitting from PageCookiesWidget. This is to trigger it on all EmbedGateway instances
- Adds URL schemes to services
- Cleans up unused services


packages/portal/src/components/embed/EmbedGateway.vue
```diff
@@ -127,7 +142,30 @@
...
 
     mounted() {
-      this.provider = serviceForUrl(this.url);
+      this.providerUrl = this.url;
+
+      if (this.embedCode) {
+        const template = document.createElement('div');
+        template.innerHTML = this.embedCode;
+        const iframe = template.querySelector('iframe');
+        const script = template.querySelector('script');
```
LeoniePeters, line 151: I'm doubting this is the best way to get the `src` and iframe dimensions from the embed code. This requires it to always run on the client (for document to be available). Also I'm not 100% sure if this is save as it does create iframe and/or script elements (although not appended to the body). Perhaps we could extract the `src`, `height` and `width` using a regex. Then we could call all this in `fetch`.

PR #2546, opened 2025-01-23
chore: embedded media notification style fixes

LeoniePeters:


packages/portal/src/components/embed/EmbedGateway.vue
```diff
@@ -253,11 +259,25 @@
...
           this.checkConsentAndOpenEmbed();
         }
+      },
+
+      async openMainCookieModalAndScrollToThirdPartyContent() {
+        this.$bvModal.show('cookie-modal');
+        await this.$nextTick();
+
+        // Wait for b-modal to be ready
+        setTimeout(() => {
```
rwd, line 274: Could/should we use the waitFor utility to ensure the availability of the element, and also not wait longer than necessary for it?

lbiedinger (approved):

PR #2553, opened 2025-01-29
a11y: refactor download attribution and item embed code snippets

LeoniePeters:
- Refactors the item attribution (in the DownloadSuccesModal) and ItemEmbedCode into a new component ItemSnippetCopyButton:
  - from forms with `textarea`'s to `cite` and `code` elements
  - the text snippet is accompanied by a hidden button for AT (assistive technologies) users to be able to reach, understand and 'click' (copy) the text snippet
  - move the embed code fetch from deprecated ItemEmbedCode to ItemHero (once fetched the `embedCode` is stored in the component's state preventing re-fetching on modal hide/show)
  - align styles for both snippets to former attribution snippet
- Sets correct title and header tags for download success and share modals


packages/portal/src/components/share/ShareSocialModal.vue
```diff
@@ -6,9 +6,11 @@
     hide-footer
     data-qa="share modal"
     title-tag="h2"
+    header-tag="div"
+    @show="$emit('show')"
   >
     <p>{{ $t('shareWhere') }}</p>
-    <div class="icon-wrapper pt-1">
+    <div class="icon-wrapper pt-1 mb-4">
       <ShareSocialButton
```
lbiedinger, line 14: unrelated, but this component should be called "shareSocialButtons" (plural) or the v-for statement moved out to here.

packages/portal/src/components/item/ItemSnippetCopyButton.vue
```diff
@@ -0,0 +1,181 @@
...
+    >
+      <span class="icon-info-outline d-inline-block mr-1" />
+      {{ helpText }}
+    </p>
+  </div>
+</template>
+
+<script>
+  export default {
+    name: 'ItemSnippetCopyButton',
```
lbiedinger, line 41: I'm not convinced that we should refer to the whole component a "button" when it does not appear styled as one to the user.
Something more like: `ItemShareableSnippet` might be clearer.

packages/portal/src/components/share/ShareSocialModal.vue
LeoniePeters: I've renamed the component.

packages/portal/src/components/item/ItemSnippetCopyButton.vue
LeoniePeters: I've renamed the component to ShareSnippet. As in the future we might reuse it for non-item features.

PR #2557, opened 2025-01-30
fix: do not override query on collection pages when fulltext search is applied

LeoniePeters:
Handles the `overrideParams.query` when there is a full text search. Because the full-text search query uses the `query` param, pushing the main query to `qf`. Overrides should be handled the same.

### DIFF
--- a/packages/portal/src/components/search/SearchInterface.vue
+++ b/packages/portal/src/components/search/SearchInterface.vue
@@ -412,6 +412,12 @@
 
         if (this.advancedSearchQueryCount > 0) {
           if (this.hasFulltextQa) {
+            // override query overrides already here to prevent full-text search breaking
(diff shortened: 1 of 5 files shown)


packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -412,6 +412,12 @@
 
         if (this.advancedSearchQueryCount > 0) {
           if (this.hasFulltextQa) {
+            // override query overrides already here to prevent full-text search breaking
+            if (this.overrideParams.query) {
+              params.query = this.overrideParams.query;
+              delete this.overrideParams.query;
```
rwd, line 418: This doesn't seem quite right as it's modifying a prop.

rwd:
Alternative approach favoured: #2564

PR #2564, opened 2025-02-05
refactor: prefer local params over overrides

rwd:
Simplify the handling of overrides vs local params from the user/URL-constructed query by always favouring the latter over the former.

### DIFF
--- a/packages/portal/src/components/search/SearchInterface.vue
+++ b/packages/portal/src/components/search/SearchInterface.vue
@@ -215,7 +215,7 @@
         type: Boolean,
         default: false
       },
-      overrideParams: {
+      defaultParams: {
         type: Object,
         default: () => ({})
       }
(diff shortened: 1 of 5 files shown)


packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -399,16 +401,18 @@
...
+          qf: this.qf,
+          query: this.query,
+          reusability: this.reusability,
+          sort: this.sort,
+          page: this.page,
+          profile: 'minimal',
+          rows: this.perPage
+        }, isUndefined);
+
+        const params = merge(this.overrideParams, localParams);
```
LeoniePeters, line 415: Rename `overrideParams` as it's not overriding, but rather setting defaults.

rwd: Note to self: or `initialParams`

lbiedinger (approved):

PR #2577, opened 2025-02-11
perf: store (don't compute) web resources

rwd:
media is being stored, but only ever used to compute web resources, resulting in increased memory usage. just store web resources in the first place.


rwd:
Can't do that as custom classes don't survive hydration.

PR #2598, opened 2025-02-14
feat: update itemDescriptions profile to items.meta

lbiedinger:
Sets new default params for set retrieval where previously the itemDescription profile was used.

* For the new API, ensure to retrieve both items & set metadata, then merge this to keep compatibility with the existing components.
* For the old API, request params are modified/removed in order to continue getting the same response data.

_Note that not deleting the page & perPage on get requests on the "old" API will also remove the metadata._


packages/portal/src/plugins/europeana/set.js
```diff
@@ -32,6 +32,10 @@ export default class EuropeanaSetApi extends EuropeanaApi {
       if (params.profile === 'items.meta' && options.withMinimalItemPreviews === true) {
         params.profile = 'standard';
       }
+      // check for items.description profile
+      if (params.profile === 'items.meta') {
+        params.profile = 'itemDescriptions';
+      }
```
LeoniePeters, line 38: This is going to override the previous condition (added as part of https://europeana.atlassian.net/browse/EC-7058). Perhaps move this condition above that?

LeoniePeters (changes requested):

packages/portal/src/plugins/europeana/set.js
lbiedinger: If the "withMinimalItemPreviews" condition is true it will switch to the standard profile and this won't trigger. Which is fine.

rwd: Indeed. I have rewritten the conditions to make this clearer and DRYer.

PR #2604, opened 2025-02-21
chore: get with items for EntityBestItemsSet

rwd:


packages/portal/src/mixins/europeana/entities/entityBestItemsSet.js
```diff
@@ -30,7 +30,7 @@ export default {
     },
 
     async fetchEntityBestItemsSetPinnedItems(entityBestItemsSetId) {
-      const entityBestItemsSet = await this.$apis.set.get(entityBestItemsSetId, {
+      const entityBestItemsSet = await this.$apis.set.getWithItems(entityBestItemsSetId, {
```
lbiedinger, line 33: `getWithItems` only takes the id as a param.
It should be extended to also allow the params that we pass here.

The profile in this case should be updated to be one of `items` or `items.meta`. Here we only need `items`. It needs to propagate to the `getItems` call.
`pageSize` for now we're always getting 100 and that's already hard coded in `getItems`, so that's not essential.

lbiedinger (changes requested):

packages/portal/src/mixins/europeana/entities/entityBestItemsSet.js
rwd: I think instead we could and should remove the second argument from the call to `getWithItems` and have it handle the required options to pass on to `get`, as it already does.

rwd: That method has now been retired.

lbiedinger (approved):

PR #2626, opened 2025-02-28
feat: item card select state

LeoniePeters:
- Adds new ItemSelectCheckbox component that toggles the item select state and has custom icon, card overlay and hover styles
- Refactors ContentCard to only use SmartLink when there is a `url`
- Refactors ItemPreviewCard to handle item select state: render a ItemSelectCheckbox and hide user buttons
- Adds `selectItem` and `deselectItem` store mutations to handle item select state


packages/portal/src/components/content/ContentCard.vue
```diff
@@ -350,7 +351,7 @@
       },
 
       displayLabelMatchString() {
-        return this.url.match(new RegExp(`/(${this.displayLabelTypes})[/.]`));
+        return this.url && this.url.match(new RegExp(`/(${this.displayLabelTypes})[/.]`));
```
rwd, line 354: Use `this.url?.match` here

packages/portal/src/components/item/ItemSelectCheckbox.vue
```diff
@@ -0,0 +1,143 @@
...
+    v-model="selected"
+    class="item-select-checkbox position-absolute"
+    :class="{ active: selected }"
+    @change="toggleItemSelection"
+  >
+    <span
+      class="m-3 position-relative d-inline-block"
+      :class="selected ? 'icon-select-circle' : 'icon-select-circle-outlined'"
+    />
+    <span class="visually-hidden">{{ selectCheckboxLabel }}</span>
```
rwd, line 12: Is repeating the item title the best choice of label for the checkbox in terms of a11y?

packages/portal/src/store/set.js
```diff
@@ -36,6 +37,12 @@ export default {
     },
     addItemToActive(state, item) {
       state.active.items.push(item);
+    },
+    selectItem(state, itemId) {
+      state.selectedItems.push(itemId);
+    },
+    deselectItem(state, itemId) {
+      state.selectedItems = state.selectedItems.filter(id => id !== itemId);
```
rwd, line 45: Wrap the function parameter in parentheses `.filter((id) =>`

rwd (changes requested):

packages/portal/src/components/item/ItemSelectCheckbox.vue
LeoniePeters: I think the item title is a clear label for the user to determine they want to select/deselect it.
In the select state the only interactive element on the item card is the checkbox. When tabbing and using a screen reader, and tabbing over the cards (i.e. checkboxes) it will read out this label and the checked state ('ticked' or 'unticked').
I do see a possible refactor: to use the card title as checkbox label (similar to it being a link in the regular state). Then we don't need duplicate card title and checkbox label elements. I opened a separate PR to do so: https://github.com/europeana/portal.js/pull/2631

PR #2656, opened 2025-03-14
feat: confirm removal of items from set

rwd:


packages/portal/src/components/item/ItemRemoveButton.vue
```diff
@@ -8,11 +8,20 @@
       data-qa="item remove button"
       :aria-label="$t('actions.remove')"
       :title="tooltipTitle"
-      @click="removeItem"
+      @click="handleClickButton"
     >
       <span class="icon-remove-circle-outlined" />
       {{ buttonText ? $t('actions.remove') : '' }}
     </b-button>
+    <ConfirmDangerModal
```
lbiedinger, line 16: On gallery pages, this results in many identical modals. They all have the same id attribute.
A side effect of that being that they are all shown.

We could declare the modal once pages that we know will potentially need it and then pass the ID as a ref to the buttons.

lbiedinger (changes requested):

packages/portal/src/components/item/ItemRemoveButton.vue
rwd: I have refactored so that it's only rendered when it's needed.

packages/portal/src/components/generic/ConfirmDangerModal.vue
```diff
@@ -61,6 +62,23 @@
...
+    },
+
+    data() {
+      return {
+        show: this.value
+      };
+    },
+
+    watch: {
+      value() {
```
LeoniePeters, line 80: Does this ever get called, now the the ConfirmDangerModal instancces always have the same value for v-if and v-model?

LeoniePeters (approved):

lbiedinger (approved):

PR #2658, opened 2025-03-17
feat: log add/like multiple items

rwd:
in a single request to the API

### DIFF
--- a/packages/portal/src/mixins/logEvent.js
+++ b/packages/portal/src/mixins/logEvent.js
@@ -15,7 +15,7 @@ export default {
       return !!(
         this.$features?.eventLogging &&
         this.eventToLog &&
-        !this.eventLogged &&
+        !this.eventLogged && // QUESTION: is this giving false positives w/ multi-select...?
         !this.eventBeingLogged &&
         !this.$fetchState?.error &&
         process.client &&
--- a/packages/portal/src/server-middleware/api/events/log.js
+++ b/packages/portal/src/server-middleware/api/events/log.js
@@ -14,64 +14,67 @@ export default async(req, res, next) => {
     }
 
     const { actionType, objectUri, sessionId } = req.body;
-    const url = new URL(objectUri);
-    // Ignore any search query or hash
-    const uri = `${url.origin}${url.pathname}`;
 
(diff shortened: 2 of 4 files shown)


packages/portal/src/server-middleware/api/events/log.js
```diff
@@ -14,64 +14,67 @@ export default async(req, res, next) => {
...
+          ON a.action_type_id=at.id
+        WHERE a.object_id=$1
+          AND at.name=$2
+          AND a.session_id=$3
+        `,
+      [objectRow.id, actionType, sessionRow.id]
+      );
+      if (selectActionResult.rowCount > 0) {
+        // this session has already logged this action type for this object; don't log it again
+        return;
```
lbiedinger, line 66: This needs to `continue;`  instead of return so that subsequent object interactions can be logged.

packages/portal/src/mixins/logEvent.js
```diff
@@ -15,7 +15,7 @@ export default {
       return !!(
         this.$features?.eventLogging &&
         this.eventToLog &&
-        !this.eventLogged &&
+        !this.eventLogged && // QUESTION: is this giving false positives w/ multi-select...?
```
lbiedinger, line 18: Partially yes, but there's another issue that impacts multi-select logging especiallyi around lolgging events where some items in the selection have already had an event logged for the same action. Specifically that's noted in server-middleware/api/events.

Trying to keep `eventLogged` as a data property doesn't make too much sense if the mixin is used in the buttons part of the selection toolbar where it  may be responsible for logging various unique events. Unique because each event has a different combination of objects.

lbiedinger (changes requested):

packages/portal/src/server-middleware/api/events/log.js
rwd: Refactored a different way to handle this, by moving the db queries to a new function.

packages/portal/src/mixins/logEvent.js
rwd: Addressed on a separate PR: https://github.com/europeana/portal.js/pull/2668

PR #2659, opened 2025-03-17
fix: hide item buttons when zero selected

LeoniePeters:


packages/portal/src/components/item/ItemSelectToolbar.vue
```diff
@@ -23,7 +23,10 @@
         </b-button>
       </template>
     </span>
-    <div class="my-1 d-flex">
+    <div
+      v-if="selectionCount >= 1"
```
rwd, line 27: This should probably use `v-show` as it is likely that a selection will be made and the buttons need to be displayed, so preferable to pre-render them.

LeoniePeters: That's done and also made so for the deselect selected button.

rwd (approved):

PR #2664, opened 2025-03-19
feat: multi-select recommended items

rwd:
relying on provide/inject to instruct when ItemPreviewCard needs to support multi-select

### DIFF
--- a/packages/portal/src/pages/account/index.vue
+++ b/packages/portal/src/pages/account/index.vue
@@ -102,6 +102,7 @@
                           {{ $tc('items.itemCount', likedItems.length) }}
                         </h2>
                         <ItemSelectButton
+                          v-if="$features.itemMultiSelect"
                           class="ml-auto"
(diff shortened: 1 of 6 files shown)


packages/portal/src/pages/account/index.vue
```diff
@@ -102,6 +102,7 @@
                           {{ $tc('items.itemCount', likedItems.length) }}
                         </h2>
                         <ItemSelectButton
+                          v-if="$features.itemMultiSelect"
                           class="ml-auto"
                           @select="(newState) => selectState = newState"
```
LeoniePeters, line 107: Rename `selectState` to `itemMultiSelect` and provide from this page as well.

LeoniePeters (changes requested):

packages/portal/src/pages/account/index.vue
rwd: Indeed. Done.

PR #2672, opened 2025-03-21
refactor: create ItemPreviewInterface component

rwd:
- incorporating heading, view toggles, multi-select, item preview card group, pagination, and methods from item preview card group view mixin
- resulting in availability of item multi-select on: main search, entity pages, liked items
- stop treating searches with 0 results as errors, which spam the APM logging
  - instead just display them _like_ an illustrated error, via the ItemPreviewInterface's no-items slot


packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -25,115 +25,89 @@
...
+                #no-items
+              >
+                <ErrorMessage
+                  :error="{ code: 'searchResultsNotFound', i18n: $t('errorMessage.searchResultsNotFound') }"
+                  :full-height="false"
+                  :show-message="false"
+                  title-tag="h2"
+                />
+              </template>
+              <template #card-group-default>
```
LeoniePeters, line 81: This slot is used by collection pages to insert the EntityHeader. So now we are naming the slot we might want to call it something more relevant like `collection-header-card`.

packages/portal/src/pages/galleries/_.vue
```diff
@@ -125,105 +125,47 @@
           />
         </div>
       </b-container>
-      <LoadingSpinner
```
LeoniePeters, line 128: Don't we want to keep the loading spinner here?

LeoniePeters (changes requested):

packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -25,115 +25,89 @@
...
+          <ErrorMessage
+            v-if="$fetchState.error"
+            :error="$fetchState.error"
+            :full-height="false"
+            :show-message="showErrorMessage"
+            title-tag="h2"
+          />
+          <b-row v-else>
+            <!-- TODO: check this is working as intended after the ItemPreviewInterface refactor -->
+            <p
```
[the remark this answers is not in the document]
rwd: Addressed by moving it into ItemPreviewInterface, as a slot with a default value. SearchInterface supplies its own text for this, but other scenarios also now get "No more items" if one paginates beyond the end of the line.

packages/portal/src/components/search/SearchInterface.vue
rwd: I have renamed slots to make explicit that they are for the card group, and this particular one becomes card-group-header all the way down then finally header.

packages/portal/src/pages/galleries/_.vue
rwd: It's included earlier on in the component.

packages/portal/src/components/search/SearchInterface.vue
LeoniePeters: Great. Only missed the entity header slot for the list view in ItemPreviewCardGroup. I've pushed a commit to make that so.

LeoniePeters (approved):

PR #2695, opened 2025-04-02
chore: upgrade node.js to v20

rwd:
Node.js v18 loses active support on 30 Apr 2025.

This updates to v20 in advance of that, and brings a reduction in Docker image size from ~ 377 MB to ~ 347 MB.


rwd:
Let's automate this. See https://github.com/europeana/portal.js/pull/2700

PR #2702, opened 2025-04-04
refactor: multilingual advanced search

LeoniePeters:


packages/portal/src/components/search/SearchInterface.vue
```diff
@@ -368,20 +368,26 @@
...
-          params.qf = (params.qf || [])
-            .concat(this.qa.filter((qa) => !this.fulltextQas.includes(qa))).concat(this.qaes);
+          // All other advanced search rules are added to the query concatenated by AND.
+          params.query = [params.query].filter(Boolean)
+            .concat(this.qa.filter((qa) => !this.fulltextQas.includes(qa))).join(' AND ');
+
+          // concatenate any advanced search rules with added entity value
+          if (this.qaes.length) {
+            params.query = params.query
+              .concat(' OR ').concat(this.qaes.join(' OR '));
```
rwd, line 389: Shouldn't the first OR here be an AND? i.e.
```js
            params.query = params.query
              .concat(' AND ').concat(this.qaes.join(' OR '));
```

LeoniePeters: No, what this does is add the entity URI value queries for which we have custom entity look up (e.g. 'creator', 'subject', defined in [advancedSearch](https://github.com/europeana/portal.js/blob/31996d063248a1f690cf4ec97237306e3d186225/packages/portal/src/mixins/advancedSearch.js#L49) mixin). The same fields with the keyword value are added before. If we AND that first OR it will AND only the first entity value query and returns only results filtered to the advanced search field value that contain both the keyword and entity value, which in many cases seem to be much less.
Looking at this again though and trying multiple advanced search queries, I think the keyword and entity values should be added to just one field instead with an OR (for example `proxy_dc_subject:(Agricultura+OR+"http://data.europeana.eu/concept/134")`, but we'll have to see what effect that has on multilingual search.

PR #2705, opened 2025-04-09
feat: track multilingual toggle in matomo

lbiedinger:


packages/portal/src/components/search/SearchMultilingualButton.vue
```diff
@@ -35,6 +35,7 @@
 
     methods: {
       toggle() {
+        this.$matomo.trackEvent('Translate search', `${this.selected ? 'Disabled' : 'Enabled'} translated search`, this.url);
```
LeoniePeters, line 38: Rename the event to 'multilingual search' as that's what we call the feature in general.
Also, instead of the url, pass in the locale. So like this:

```
this.$matomo.trackEvent('Multilingual search', `${this.selected ? 'Disabled' : 'Enabled'} multilingual search`, this.$i18n.locale);
```

LeoniePeters (changes requested):

LeoniePeters (approved):

PR #2726, opened 2025-04-22
style: api keys page and table

LeoniePeters:


packages/portal/docs/style/Colours.md
```diff
@@ -30,6 +30,7 @@ Colours are also defined on the [Colours page on Figma](https://www.figma.com/fi
           { name: 'white', hex: '#fff' },
           { name: 'smoke', hex: '#f5f5f5' },
           { name: 'offwhite', hex: '#f1f1ee' },
+          { name: 'lightgrey', hex: '#fbfbfb' },
```
rwd, line 33: There are now two lightgrey colours here. Is that as it should be?

packages/portal/src/pages/account/api-keys.vue
```diff
@@ -69,9 +78,6 @@
                       </template>
                     </template>
                   </b-table>
-                  <p v-else>
-                    {{ $t('apiKeys.noKeys') }}
```
rwd, line 73: If no longer needed, remove the key from the i18n file.

rwd (changes requested):

packages/portal/docs/style/Colours.md
LeoniePeters: Nope, the first should be `lightergrey`. That's updated.

packages/portal/src/pages/account/api-keys.vue
LeoniePeters: That's unused and removed.

rwd (approved):

PR #2735, opened 2025-04-25
feat: debug API requests user API keys select

rwd:


rwd:
Only 1 will be active, so don't need a dropdown. See #2738 instead.

PR #2741, opened 2025-04-29
feat: cookies plugin

rwd:
Creates a custom cookies plugin for getting and setting cookies on either SSR or CSR.

So that we may remove the dependency on cookie-universal-nuxt which has not been updated since August 2022.


rwd:
Re EC-7194

PR #2755, opened 2025-04-30
fix: skip link without hash

LeoniePeters:


packages/portal/src/components/generic/SkipLink.vue
```diff
@@ -0,0 +1,38 @@
...
+<script>
+  export default {
+    name: 'SkipLink',
+
+    props: {
+      selector: {
+        type: String,
+        default: '#main'
+      },
+      classes: {
```
lbiedinger, line 20: When would this be changed?

PR #2771, opened 2025-05-13
refactor: mv contentful module to composable

rwd:


packages/portal/src/graphql/queries/landingPage.graphql
```diff
@@ -106,20 +114,7 @@ query LandingPage(
...
   }
 }
-fragment imageFields on Asset {
-  title
-  url
-  contentType
-  description
-  width
-  height
-  contentfulMetadata {
```
lbiedinger, line 116: This was the only place we are getting contentfulMetadata and the asset title back.
I don't think they're being used, but since they're included in the new fragment it's going to add unnecessary object data to any page/component that's using contentful images.

lbiedinger (approved):

PR #2781, opened 2025-06-04
fix: amazon s3 thumbnails

rwd:


rwd:
Only intended as a temporary workaround for an IBM Cloud IAM outage now resolved.

PR #2794, opened 2025-07-14
chore: add PollUnit to services

LeoniePeters:


rwd:
Let's get the refactor to move service definitions to JSON files in first, then run the GHA workflow that comes with to add PollUnit.

PR #2806, opened 2025-08-01
chore: delete blog post content entries

rwd:
- rename contentful package utility script `purgeContentType` to `deleteContentEntries`
- alter that script to only delete the content entries, not the model itself
- modify historical migration scripts that used `purgeContentType` to be two separate migrations: the first to delete the content entries; the second to delete the content type
- add a migration to delete the blog post content entries (but not the content type as we will keep using it)


packages/contentful/migrations/0083.1-delete-heroHeader-content-type-entries.cjs
```diff
@@ -0,0 +1,5 @@
+const deleteContentEntries = require('../src/utils/deleteContentEntries.cjs');
+
+module.exports = async function (migration, context) {
+  await deleteContentEntries('heroHeader', migration, context);
+};
```
LeoniePeters, line 5: Could these migrations contain both the entries and content type deletion? Like so:
```
const deleteContentEntries = require('../src/utils/deleteContentEntries.cjs');

module.exports = async function (migration, context) {
  await deleteContentEntries('heroHeader', migration, context);

  migration.deleteContentType('heroHeader');
};
```

rwd: Yes, thanks, that makes more sense, and has been actioned.

LeoniePeters (approved):

PR #2816, opened 2025-09-09
feat: create CTF contentHubPage content type

rwd:
To support new DS site "data-space" page w/ blog posts and projects, later also to replace europeana.eu site's stories page.


packages/contentful/migrations/0153-create-contentHubPage-type.cjs
```diff
@@ -0,0 +1,243 @@
...
+      {
+        unique: true,
+      },
+      {
+        size: {
+          max: 100,
+        },
+
+        message:
+          'This is an H1 field, it needs to be unique for SEO and not more than 100 characters.',
```
lbiedinger, line 35: also update here re: uniqueness

PR #2982, opened 2026-04-08
refactor: set oEmbed responsiveness in service definitions

rwd:
- stop hard-coding in EmbedOEmbed component which oEmbed service providers have responsive displays and record instead in the service definitions
- update the GHA workflow and accompanying script for adding media services to support specifying the responsiveness property, for oEmbed services only


packages/portal/src/utils/services/definitions/2D.json
```diff
@@ -23,6 +23,7 @@
   {
     "name": "institutNationalDeLAudiovisuel",
     "oembed": "https://oembed.europeana.eu/",
+    "responsive": false,
```
LeoniePeters: Set to `true`.

packages/portal/bin/media-embed-service-add.js
```diff
@@ -4,28 +4,49 @@ const PROPERTIES = [
...
     if (prop.split) {
       service[prop.id] = service[prop.id].split(',');
     }
   }
 
+  for (const prop of PROPERTIES) {
+    if (prop.dependsOn && !service[prop.dependsOn]) {
+      delete service[prop.id];
+    }
+  }
```
LeoniePeters, line 48: Can't this be in the above `for...of` loop?

packages/portal/src/components/embed/EmbedOEmbed.spec.js
```diff
@@ -23,7 +23,7 @@ const response = {
 const factory = ({ propsData = {}, data = {} } = {}) => shallowMountNuxt(EmbedOEmbed, {
   propsData: {
     url,
-    endpoint,
+    // endpoint,
```
LeoniePeters: Remove.

LeoniePeters (changes requested):

packages/portal/src/utils/services/definitions/2D.json
rwd: Done.

packages/portal/bin/media-embed-service-add.js
rwd: If it were, then the order may sometimes interfere, i.e. if the dependent is handled before the dependency.

packages/portal/src/components/embed/EmbedOEmbed.spec.js
rwd: Done.

PR #3000, opened 2026-04-13
POC: three.js 3D media viewer

LeoniePeters:


rwd:
We will go with #3015 instead

PR #3013, opened 2026-04-14
feat: add isFormatOf to displayableWebResources

LeoniePeters:


rwd:
Alternative incorporated into #3015 instead

PR #3019, opened 2026-04-16
feat: prefer gltf format to oEmbed

rwd:
For oEmbed web resources, look to see if they have a dcterms:isFormatOf property linking to a gltf 3D web resource, and if so, prefer that to the oEmbed so that it may be displayed without the 3rd-party embed.


rwd:
incorporated into #3015

PR #3023, opened 2026-04-17
chore: add A/V error handling for euscreen embeds

lbiedinger:


packages/portal/src/components/media/MediaAudioVisualPlayer.vue
```diff
@@ -131,11 +131,12 @@
 
     async fetch() {
       if (this.euScreenEmbedUrl) {
-        // TODO: Error handling on the embed response
-        const response = await axios.get(this.euScreenEmbedUrl);
-
-        this.mediaUrl = response.data.location;
-        this.mediaFormat = response.data.format;
+        const response = await axios.get(this.euScreenEmbedUrl)
```
rwd: If using await, also use try/catch. Setting mediaUrl and mediaFormat can then just go in the try block.

rwd (approved):

PR #3028, opened 2026-04-21
feat: download menu for isFormatOf WRs

rwd:
- where a web resource has other formats available via isFormatOf, offer them for download in a dropdown menu
- menu item text displays the file extension derived from ebucoreHasMimeType and if available file size from ebucoreFileByteSize
- in the item page, gather a list of proxyable web resources and provide a function to descendent components to check whether a given WR should be proxied
  - to be proxied are any WRs linked from the provider aggregation, except for DASH videos which are manifest-based and not yet fully supported by the media proxy
  - use this in DownloadButton, MediaAudioVideoPlayer and Media3DViewer


packages/portal/src/components/download/DownloadButton.spec.js
```diff
@@ -315,13 +340,14 @@ describe('components/download/DownloadButton', () => {
             const wrapper = factory({ propsData });
 
             await wrapper.vm.handleClickDownloadButton(event);
-
+            console.log('downloadUrl', wrapper.vm.downloadUrl, url);
+            console.log(wrapper.vm.$apm.captureError.getCalls()[0].args);
```
LeoniePeters: Clean up `console.log`

packages/portal/src/components/download/DownloadButton.vue
```diff
@@ -30,7 +33,12 @@
     components: {
       LoadingSpinner
     },
-    inject: ['canonicalUrl'],
+    inject: {
+      canonicalUrl: {},
+      isProxyable: {
+        default: null
```
LeoniePeters, line 39: Does this need a default? And if so, shouldn't it be a function, as that's the expected type?

LeoniePeters (changes requested):

packages/portal/src/components/download/DownloadButton.vue
rwd: It simplifies testing, and with the call being `isProxyable?.(url)` can be null.

PR #3032, opened 2026-04-23
chore: remove feature toggle for webResourceMetadata display

lbiedinger:


packages/portal/src/components/item/ItemMediaPresentation.spec.js
```diff
@@ -22,7 +22,7 @@ const factory = ({ data = {}, propsData = {}, mocks = {} } = {}) => shallowMount
     $error(error) {
       throw error;
     },
-    $features: { webResourceMetadata: true },
+    $features: {},
```
lbiedinger, line 25: this is still needed as an empty object for the video.js specs

PR #3033, opened 2026-04-24
style: format ebucore duration as time(hh:mm:ss)

lbiedinger:
To make the technical metadata more understandable for humans.
Use custom logic to format the durations.
The format will be:
hh:mm:ss
With any '00' elements removed.


packages/portal/src/components/metadata/MetadataField.spec.js
```diff
@@ -354,4 +354,64 @@ describe('components/metadata/MetadataField', () => {
...
+
+        it('removes leading zeros', () => {
+          const wrapper = factory({ props });
+          const fieldValue = wrapper.find('[data-qa="metadata field"] ul [data-qa="literal value"]');
+
+          expect(fieldValue.text()).toEqual('2:30');
+        });
+      });
+
+      describe('when there is a decimal minute value', () => {
```
rwd: "decimal minute value"? what does that mean?

packages/portal/src/components/metadata/MetadataField.vue
```diff
@@ -244,12 +249,28 @@
...
+      isDurationValue() {
+        return this.name === 'ebucoreDuration';
       }
     },
 
     methods: {
       isNumberValue(value) {
         return typeof value === 'number';
+      },
+      formatDuration(value) {
```
rwd, line 262: This is not yet handling properly where there is a minute value < 10, which should have a leading "0".

Example item: /428/_kmo_av_sid_16458

Data value is 3950160, which is displayed as "1:5:50" but should be "1:05:50"

rwd (changes requested):

packages/portal/src/components/metadata/MetadataField.spec.js
lbiedinger: This has been reworded as "double digit" and the wording updated to single/double digit in the other specs for consistency.

packages/portal/src/components/metadata/MetadataField.vue
lbiedinger: that's addressed and covered by the specs too with an update to the values there.

rwd (approved):

PR #3056, opened 2026-05-07
refactor: custom auth plugin

rwd:
- reproduces the functionality of the @nuxtjs/auth module together with existing custom extensions for keycloak, into a new local auth plugin
- moves the login/logout pages from /account to /auth


packages/portal/src/components/page/PageNavigation.vue
```diff
@@ -38,13 +38,32 @@
...
-        return this.isAuthenticated ? [
-          { url: '/account', text: this.$t('account.title'), dataQa: 'account link' },
-          this.sidebarNav && { url: '/account/logout', to: '/account/logout', text: this.$t('account.linkLogout'), dataQa: 'log out link' }
-        ] : [
-          { url: '/account/login', to: { name: 'account-login', query: { redirect: this.$route.fullPath } }, text: this.$t('account.linkLoginJoin'), dataQa: 'log in link' }
-        ];
+        if (this.isAuthenticated) {
+          return [
+            { url: '/account', text: this.$t('account.title'), dataQa: 'account link' },
+            this.sidebarNav && { url: '/auth/logout', to: { name: 'auth-logout', query: { redirect: this.logoutRedirect } }, text: this.$t('account.linkLogout'), dataQa: 'log out link' }
```
lbiedinger: the case for `'/account/logout'` in `renderIcon` needs to be updated to handle `'/auth/logout'` instead.

packages/portal/src/components/page/PageNavigation.vue
```diff
@@ -38,13 +38,32 @@
...
+          ];
+        } else {
+          return [
+            { url: '/auth/login', to: { name: 'auth-login', query: { redirect: this.$route.fullPath } }, text: this.$t('account.linkLoginJoin'), dataQa: 'log in link' }
```
lbiedinger: the case for `'/account/login''` in `renderIcon` needs to be updated to handle `'/auth/login'` instead.

packages/portal/src/pages/item/_.vue
```diff
@@ -184,17 +184,10 @@
...
     middleware: [
       // When entering a translated item page, but not logged-in,
-      // redirect to Keycloak to login, unless user just logged out in which case,
-      // redirect to page without translation.
+      // redirect to Keycloak to login.
       ({ $auth, redirect, route }) => {
         if (route.query.lang && !$auth.loggedIn) {
-          if ($auth.$storage.getUniversal('portalLoggingOut')) {
-            // just logged out: redirect to page w/o lang param
-            return redirectToAltRoute({ query: { lang: undefined } }, { redirect, route, status: 303 });
```
lbiedinger, line 190: this case still needs to be handled.

packages/portal/src/components/page/PageNavigation.vue
rwd: Done

packages/portal/src/components/page/PageNavigation.vue
rwd: Done

packages/portal/src/pages/item/_.vue
rwd: it is supposed to be handled instead by the `logoutRedirect` computed property in PageNavigation, but that's not working quite right at the moment

PR #3074, opened 2026-05-12
feat: add aggregator tables to tab

LeoniePeters:


packages/portal/src/components/entity/EntityTable.vue
```diff
@@ -263,6 +331,25 @@
       },
       updateRouteQuery(newQuery) {
         this.$router.push({ ...this.$route, query: { ...this.$route.query, ...newQuery } });
+      },
+      async fetchAggregatesFrom(aggregatorId, limit) {
+        const collectionIndex = this.collections.findIndex(c => c.id === aggregatorId);
+
+        const aggregatorFullData = await this.$apis.entity.find([aggregatorId]);
```
lbiedinger: This will re-fetch the aggregator entity when expanding the 'view all' list.

LeoniePeters: I've refactored and moved fetch and display of the badges into a separate component. The aggregator is no longer re-fetched when clicking 'view more'.

rwd (approved):

PR #3084, opened 2026-05-19
refactor: fetch from API for entity table

rwd:
- show record count on all entity types
- add aggregator links for institutions


rwd:
We can't go with this approach due to Entity API limitations and issues. See https://europeana.atlassian.net/browse/EC-7767?focusedCommentId=173730

PR #3085, opened 2026-05-19
refactor: add BFF for entity table

rwd:
adds a backend-for-frontend (BFF) for supplying the data for the EntityTable, responsible for its filtering, sorting and pagination, so that the full dataset does not need to be loaded into the front-end and its manipulation handled client-side

- BFF is at `/_api/collections/:type`, accepting query params `lang`, `query`, `page`, `pageSize` & `sort`
- EntityTable updated to call this endpoint on CSRs, or call the middleware's data retrieval function directly on SSR (to avoid server-to-self sub-requests)
- EntityTable now rendered both client- and server- side
- organisations cacher updated to use `isAggregatedBy.recordCount` from Entity API for the cached record count instead of getting it from Record API
- organisations & aggregators cachers will determine each organisation's native and non-native English name and cache them, so that EntityTable does not have to determine them
- aggregators cacher updated to use entity retrieve endpoint instead of making multiple fetch requests
- from organisations cacher, removes redundant compatibility handling for different types of country data on Entity API responses as it always returns objects w/ ID and prefLabel now


packages/portal/src/components/entity/EntityTable.vue
```diff
@@ -201,48 +201,32 @@
...
         }
-      } catch (e) {
-        // TODO: set fetch state error from message
-        console.error({ statusCode: 500, message: e.toString() });
       }
-    },
 
-    fetchOnServer: false,
+      if (this.orgOrAggType) {
+        collections = collections.map(this.organisationData);
```
LeoniePeters, line 223: This seems to be skipped over for the aggregator types.

LeoniePeters (changes requested):

packages/portal/src/components/entity/EntityTable.vue
rwd: The aggregators cacher needed to be updated, which it has now.

lbiedinger (approved):

PR #3097, opened 2026-05-28
feat: aggregator badges for providing institutions

rwd:
Requires updating the cache: `npm run -w packages/portal cache set collections:organisations`


packages/portal/src/components/entity/EntityBadges.vue
```diff
@@ -64,7 +64,7 @@
        * Title which appears above the badges
        */
       title: {
-        type: String,
+        type: [Boolean, String],
```
LeoniePeters: Is this still needed, when we depend on the `showTitle` prop for display?

PR #3104, opened 2026-06-02
feat: expand entity table row details from route hash

rwd:
- updates useActiveTab composable:
  - to accept an array of tab IDs, not hashes
    - updates all consuming components accordingly
  - to check for an option `query` which if present will cause it to inspect/update route query param named therein instead of the hash
  - replaces its use of vue2-helpers for useRoute and useRouter with custom implementations of those as on SSRs, route.query would always be empty
- updates EntityTable component to expand aggregated institutions based on URL:
  - puts the tab selection into route query like `?tab=aggregators`
  - puts the last toggled collection's numeric ID into the route query like `?show=1234`
  - does not scroll when the row is toggled
  - does scroll when the page is loaded with the hash (with native `scrollIntoView`)
- moves (minimal) entity organisation tabs code into EntityOrganisationsPageContent component 
- removes client-only wrapper around EntityOrganisationsPageContent content as it is no longer reliant on the URL hash


packages/portal/src/components/entity/organisations/EntityOrganisationsPageContent.vue
```diff
@@ -136,12 +156,19 @@
           }
         ];
       }
+    },
+
+    methods: {
+      tabRoute(tab) {
+        return { ...this.$route, query: { tab: tab.id }, hash: undefined };
```
LeoniePeters: Is it still needed to override the route hash here?

packages/portal/src/components/entity/EntityTable.vue
```diff
@@ -299,20 +307,61 @@
...
       }
     },
 
+    mounted() {
+      this.scrollToFocusedCollection();
+    },
+
     methods: {
+      expandFocusedCollection() {
+        const focusedCollection = this.collections?.find((entity) => entity.numericId === this.$route.query.show);
```
LeoniePeters: Use the `focusedCollection` computed prop instead.

packages/portal/src/components/entity/organisations/EntityOrganisationsPageContent.vue
rwd: No. Removed.

packages/portal/src/components/entity/EntityTable.vue
rwd: Refactored otherwise.

PR #3106, opened 2026-06-02
refactor: use BFF for aggregatesFrom

rwd:
to reduce the amount of data retrieved vs calling Entity API directly


rwd:
Nope. #3112

PR #3107, opened 2026-06-03
fix: "leaving" flag for pinned sets, to prevent erroneous route modifications

lbiedinger:
Set a flag, to avoid the route being updated via the redirectToPrefPath logic when leaving a pinned set.


rwd:
#3113

PR #3113, opened 2026-06-04
refactor: create usePinnedItems composable

rwd:
to replace mixin & store state, combined w/ provide/inject of currentSet and currentEntity


rwd:
Re EC-7783

lbiedinger (approved):

PR #3130, opened 2026-06-17
chore: cache map data as GeoJSON

lbiedinger:
- rename the collections/organisations/map cacher to collections/organisations/geo
- have it cache the data in GeoJSON format
- add a new backend endpoint which returns just this GeoJSON data, at /_api/collections/organisations/geo


packages/portal/src/cachers/collections/organisations/map.js
```diff
@@ -1,25 +1,39 @@
...
 
-const PICK = ['id', 'geo'];
+const toGeoJsonFeature = (id, lat, long) => {
+  return {
+    type: 'Feature',
+    id,
+    geometry: {
+      type: 'Point',
+      coordinates: [
+        lat,
```
rwd: longitude needs to go first in GeoJSON coordinates

packages/portal/src/cachers/collections/organisations/map.js
```diff
@@ -1,25 +1,39 @@
...
-        ...entity,
-        geo
-      };
-    });
+  return {
+    type: 'FeatureCollection',
+    crs: {
+      type: 'name',
+      properties: {
+        name: 'urn:ogc:def:crs:OGC:1.3:CRS84'
```
rwd: what is this?

rwd (changes requested):

packages/portal/src/cachers/collections/organisations/map.js
lbiedinger: It's supposed to specify the Coordinate Reference System: https://datatracker.ietf.org/doc/html/rfc7946#section-4
It was taken from the maplibre-gl-js example. It does not appear needed actually, as leaving it out uses the default system which seems fine.

rwd (approved):

PR #3160, opened 2026-07-20
feat: use @europeana/map on item page

rwd:
- create a new component EmbedEuropeanaMap which is responsible for loading @europeana/map and rendering the map
- update EntityOrganisationsMap to use EmbedEuropeanaMap, supplying it the URL to the organisations GeoJSON
- create a new component ItemLocationMap which uses EmbedEuropeanaMap, suppyling it GeoJSON created from the latitude and longitude of the item location
- update MetadataBox to use ItemLocationMap if `ENABLE_ORGANISATIONS_MAP=1`, else continue using `EmbedMap` (to be deprecated)

### DIFF
--- a/packages/portal/src/components/embed/EmbedEuropeanaMap.vue
+++ b/packages/portal/src/components/embed/EmbedEuropeanaMap.vue
@@ -33,8 +33,8 @@
   const VUE_3_CDN_BASE_URL = 'https://cdn.jsdelivr.net/npm/vue@3.5.39/dist';
   const VUE_3_SCRIPT_URL = `${VUE_3_CDN_BASE_URL}/vue.global.prod.js`;
 
-  // const EUROPEANA_MAP_CDN_BASE_URL = 'https://cdn.jsdelivr.net/npm/@europeana/map@0.1.8/dist';
-  const EUROPEANA_MAP_CDN_BASE_URL = 'http://localhost:4173';
+  const EUROPEANA_MAP_CDN_BASE_URL = 'https://cdn.jsdelivr.net/npm/@europeana/map@0.1.8/dist';
+  // const EUROPEANA_MAP_CDN_BASE_URL = 'http://localhost:4173';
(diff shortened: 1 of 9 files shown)


packages/portal/src/components/embed/EmbedEuropeanaMap.vue
```diff
@@ -0,0 +1,178 @@
+<template>
+  <div
+    class="embed-map mb-5"
+  >
+    <label
+      v-if="$slots.label"
+      for="europeana-map"
+      class="p-3"
+    >
```
LeoniePeters, line 8: As flagged by SonarCloud, do not use `label` element. A `div` make sense so the slot remains flexible and could be any element. Or simply pass in the slot directly without this wrapper element.

LeoniePeters (changes requested):

packages/portal/src/components/embed/EmbedEuropeanaMap.vue
```diff
@@ -0,0 +1,178 @@
...
+    methods: {
+      handleLoadVue3() {
+        this.vue3Loaded = true;
+      },
+      handleLoadEuropeanaMap() {
+        this.europeanaMap = new window.EuropeanaMap.EuropeanaMapWrapper('#europeana-map', {
+          controls: this.controls,
+          hash: this.hash,
+          json: this.json,
+          pinPopover: 'europeana-map-popover',
```
LeoniePeters: Only pass in when the popover slot is used

LeoniePeters (changes requested):

packages/portal/src/components/embed/EmbedEuropeanaMap.vue
rwd: It's now a `div`

packages/portal/src/components/embed/EmbedEuropeanaMap.vue
rwd: Done

LeoniePeters (approved):

PR #3195, opened 2026-07-30
feat(portal): enable hash for map section

rwd:
with dynamically generated element IDs and script/style hids based on section name


lbiedinger (approved):

lbiedinger (changes requested):
tests need to be updated.

lbiedinger (approved):
test are passing again

PR #3233, opened 2026-08-21
feat(postgresql): add web_element_event_types table

rwd:
referenced by sessions table

after running the three migrations against a portal.js PG db, then run pg_repack on the events.sessions table:
```shell
pg_repack --dbname=${DB} --table=events.sessions
```


rwd:
Approach #3234 preferred.

PR #3235, opened 2026-08-26
update audiovisual theme identifer

lbiedinger:


lbiedinger:
Use hotfix PR instead: https://github.com/europeana/portal.js/pull/3236

